PR #35, opened 2016-12-07
Add WalletLoaderService functionality to prepare wallet for actions

alexlyp:
We are matching the functionality of Paymetheus and the use of the grpc WalletLoaderService:
https://github.com/decred/dcrwallet/blob/master/rpc/documentation/api.md#walletloaderservice

This will allow us to launch dcrwallet with --noinitialload 

Currently the wallet prep actions are linked to together.  For example:

On successful loader resolution the walletExists action is fired.  On successful walletExist, then openWallet action is fired etc.

If at any point one of these steps should fail, the hope is that one for the alternate Home.js views should take over and let the user attempt to enter non-default information into a form and complete that step and proceed.


jcvernaleo:
Looks good.  Make sure the final squashed commit message has the useful info from your comment on the PR (too many individual commits for me to tell if it is there already).

PR #48, opened 2016-12-08
Set version to something more reasonable.

### LINKED ISSUE #27: version number
Currently when one builds the package with
```
npm run package
```
the result is decrediton-1.0.0.dmg (or other platform specific extension).  Probably should not be 1.0 just yet :)

jcvernaleo:
Not even released yet, so clearly not 1.0

Also use same version in both places.

Closes #27

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,8 +1,8 @@
 {
   "name": "decrediton",
   "productName": "decrediton",
-  "version": "1.0.0",
-  "description": "Electron application boilerplate based on React, React Router, Webpack, React Hot Loader for rapid application development",
+  "version": "0.0.1",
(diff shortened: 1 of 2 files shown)


app/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "decrediton",
   "productName": "decrediton",
-  "version": "1.0.0",
+  "version": "0.0.1",
   "description": "Electron application boilerplate based on React, React Router, Webpack, React Hot Loader for rapid application development",
```
alexlyp: Can you update that description as well?

jcvernaleo: Done.  No idea how I missed that one when it was right under something i did edit.

PR #52, opened 2016-12-08
Finish boilerplate for redux/grpc calls

alexlyp:
The PR accomplishes a few things: 

- Split out middleware/grpc services into individual files (client/loader/seeder).
- Create boilerplate structure for api requests: (*Request, *RequestAttempt, 
  *Error, *Response), with the expectation to refactor/clean into a more 
  reusable model.  
- Confirmed that rough loader/exist/open/create/close procedure works.


jcvernaleo:
I think there might be a bit too much in this PR.

The seed part is what concerns me mostly (since it is trying to make a seed even though I have one already) and the additions to the gui as well.

Would it be possible to put those in a new PR and just have all the new boiler plate stuff and reorganization alone in this PR?

jcvernaleo:
Could you update the README.md to give the username/password needed for the rpc server?  I think that is only for dcrd (wallet still doesn't need rpc, right?).

PR #66, opened 2016-12-13
Add material-ui React component implementation remove react-bootstrap

alexlyp:


jcvernaleo:
Other than the two install related things (react-bootstrap and eslint-formatter-pretty) this is looking good.

Either remove the deps for them or add them to the package.json and this will be good to go.

PR #79, opened 2016-12-14
Use the same license all over.

### LINKED ISSUE #78: License
Need to have the same license in readme, LICENSE, and both package.json files.

jcvernaleo:
Closes #78


jcvernaleo:
Obviously, if preferred, I can make it MIT all over, but since we normally use ISC and they are basically the same.

PR #95, opened 2016-12-16
Adds button on Home page to allow for users to start rescan

### LINKED ISSUE #94: Add rescan button to dashboard and trigger rescan on Create

alexlyp:
Also rigged in triggering a rescan upon wallet creation.
As @jrick rightly pointed out: this should only happen when creating a wallet from an existing seed.  Newly generated seeds have no need for rescan.
Closes #94

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -25,9 +25,12 @@ function getWalletServiceSuccess(walletService) {
 
     // Check here to see if wallet was just created, if so
     // start rescan from 0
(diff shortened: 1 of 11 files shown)


jcvernaleo:
I tested it with making a new wallet and that did a rescan.

Would it be possible to use less decimals in the percent done though?

jcvernaleo:
Also needs an npm run lint-clean

jcvernaleo:
OK again (on the new changes/

Issue #96, opened 2016-12-16
Rescan existing wallets from newly fetched blocks


alexlyp:

alexlyp:
Closed in #98

PR #98, opened 2016-12-17
Add SeedService to allow for new seed generation and existing seed processing

### LINKED ISSUE #93: Implement SeedService to offer new seeds (or decode existing ones)

alexlyp:
With the new dcrwallet gRPC SeedService, it allows for consistent seed processing from the locally connected wallet.  

Now a user can create a wallet with an existing seed as well as use an existing seed to recreate a wallet in Decrediton

Closes #93 #96 and attends to the issue @jrick mentioned in #94


jcvernaleo:
Tested with new and existing seed on testnet.  All worked.

PR #115, opened 2016-12-22
Allow packaged app to find api.proto.

### LINKED ISSUE #114: Packaged files can't read api.proto
Long term we probably should use the static codegen option, but we might want a hack for now.

jcvernaleo:
To allow the packaged app to find the api.proto write it to the app dir on startup.

This is a temporary measure only.

Closes #114


jcvernaleo:
And I tested with the packaged version.

PR #118, opened 2016-12-22
grpc: Switch to pregenerated javascript bindings instead of dynamic loading of api.proto

### LINKED ISSUE #113: Use static codegen for grpc
To avoid having to dynamically load the grpc proto descriptor, add the static codegen as described here:
https://github.com/grpc/grpc/tree/master/examples/node/static_codegen

alexlyp:
Fixes #113


jcvernaleo:
All good on linux and mac.  Just squash and it's ready.

Issue #121, opened 2017-01-02
start daemons from package on OSX


jcvernaleo:
on osx, dcrd and dcrwallet do not start automatically when running the production/packaged version.  This might be a signing issue.  Regardless, needs to be addressed.

jcvernaleo:
Looks like if the binaries are in the .app bundle this works even if not signed.  I also tested where I downloaded the .dmg from the internet to make sured osx pops up a warning about it and it still worked with #127

PR #127, opened 2017-01-04
Include dcr* binaries when packaging.

### LINKED ISSUE #123: Bundled cmd line tools with decrediton
decrediton needs to bundle dcrd, dcrwallet, etc so a user only needs to install a single package rather than using dcrinstall as a separate step.

jcvernaleo:
Closes #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,19 +70,32 @@ Start decrediton
 npm run dev
 ```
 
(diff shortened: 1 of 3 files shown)


jcvernaleo:
Also
Closes #121

PR #132, opened 2017-01-05
don't assume the certificate exists and print error if it doesn't

### LINKED ISSUE #131: unchecked wallet certificate error
I get the following error in the console if the wallet is not running/the certificate doesn't exist:

Uncaught Error: ENOENT: no such file or directory, open '/home/jolan/.decrediton/rpc.cert'

Partially related to #108.

jolan:
Closes #131.


jcvernaleo:
I like the idea but when I run it on linux in dev mode, the check fails even when wallet is running.

jolan:
Oops, yeah I didn't test the success case. exists() is deprecated anyway and shouldn't be used.

PR #145, opened 2017-01-09
add missing radium dep

jolan:


alexlyp:
Oops good catch, must have not done npm install --save-dev

PR #157, opened 2017-01-11
Give user some feedback for daemon syncing

### LINKED ISSUE #109: GUI should show something about IBD
When starting decrediton for the first time, nothing happens after clicking Get Started button for a long time while the initial block download happens.  The terminal shows the dcrd log, but a GUI only user would have no idea what is going on.  We really should either show some progress indicator or at least give a warning.

alexlyp:
Due to various constraints on paths of communication, we currently can't easily determine what the progress on a daemon/wallet block syncness is.

To give users some sort of feedback, we've decided to calculate a rough block height that a daemon "should be at" given the current date.  So, for example, on testnet:  It's been roughly 350 days since the chain started.  With block time set in the dcrd params we can roughly expect 720 blocks mined each day on testnet.  Then we multiple by 95% to accomodate for a bit of variance.  So the current "neededBlocks" value is 239864 which is just under our current height of ~242000.

Obviously this is not ideal, but should be a good replacement in the meantime until we have another option.
Fixes #109

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -32,8 +32,8 @@ function getWalletServiceSuccess(walletService) {
     setTimeout( () => {dispatch(transactionNftnsStart());}, 1000);
     const { fetchHeadersResponse, walletCreateExisting } = getState().walletLoader;
     if ( walletCreateExisting && fetchHeadersResponse !== null ) {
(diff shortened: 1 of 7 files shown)


jcvernaleo:
Worked in my first test (fully synced daemon) now.  Letting daemon miss a few blocks then I'll try again but looking good so far.

PR #163, opened 2017-01-13
display RPC errors

### LINKED ISSUE #162: need to display startRPC errors
Currently if you hit invalid auth or something you get no feedback on what is wrong.

jolan:
Closes #162.

### DIFF
--- a/app/components/GetStarted.js
+++ b/app/components/GetStarted.js
@@ -122,7 +122,7 @@ class Home extends Component{
         </div>
       );
     } else {
-     startRpc = (
+      startRpc = (
       <div>
         <ShowError error={startRpcError} />
       </div>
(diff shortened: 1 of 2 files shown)


app/components/GetStarted.js
```diff
@@ -254,6 +255,7 @@ class Home extends Component{
           <Step>
             <StepLabel>Start RPC Concensus</StepLabel>
             <StepContent>
+              <ShowError error={startRpcError} />
               {}
```
alexlyp: Can you define a new const for 
startRpc = (
<div>
  <ShowError>
</div>
);
then in put the startRpc in the curly brackets there.

Basically the same way it is done for the other steps

jolan: Done.  Also added the progress indicator.

PR #164, opened 2017-01-13
Improve the documentation.

### LINKED ISSUE #136: linux: document libXScrnSaver dependency
there is no documentation about the libXScrnSaver package that was required on fedora, you only learn you need it when the binary fails to start. instructions for adding this package on common linux variants should be in doco somewhere.

jcvernaleo:
Add notes for extra package needed on fedora.

Move install to the top since most people will need that rather than
building instructions.

Mention glide in packaging section.

Some minor formatting fixes.

Closes #136

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,47 @@
 decrediton is a cross-platform GUI for decred written in node.js using
 Electron.
 
+## Installation
+
+Currently decrediton has only been tested and built on Linux and
+macOS.  Additional systems will be added in the future.
+
+Decrediton will NOT use or in any way disrupt the wallet file you may
+already be using at this time.
+
+Download the decrediton release for your operating system.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,47 @@
...
+Currently decrediton has only been tested and built on Linux and
+macOS.  Additional systems will be added in the future.
+
+Decrediton will NOT use or in any way disrupt the wallet file you may
+already be using at this time.
+
+Download the decrediton release for your operating system.
+
+On macOS, Ubuntu (14.04 and later), and recent Debians, there should be
+no additional dependancies needed.
```
jolan: We use 'dependencies' across all the repos even though both are acceptable.

jcvernaleo: corrected

PR #166, opened 2017-01-13
prevent infinite loop

### LINKED ISSUE #165: need a more reliable method of finding the block height
Code as it stands now can get stuck for awhile fetching 0 headers until a block comes in.

15:14:22 2017-01-13 [INF] WLLT: Fetching headers
15:14:22 2017-01-13 [INF] WLLT: Fetched 0 new header(s)
15:14:23 2017-01-13 [INF] WLLT: Fetching headers
15:14:23 2017-01-13 [INF] WLLT: Fetched 0 new header(s)
15:14:24 2017-01-13 [INF] WLLT: Fetching headers
15:14:24 2017-01-13 [INF] WLLT: Fetched 0 new header(s)

jolan:
This is a workaround/debug for #165.  Not sure if this is correct since I didn't dig deep enough yet to understand what is going on.

The debug output I get is:

curBlocks 0
getFetchedHeadersCount 0
neededBlocks 241203.04985583338
newCurBlock1 0
newCurBlock2 0
getFirstNewBlockHeight 0

Without this it will just keep fetching 0 headers seemingly forever.

Closes #165.

### DIFF
--- a/app/actions/WalletLoaderActions.js
+++ b/app/actions/WalletLoaderActions.js
@@ -351,10 +351,17 @@ function fetchHeadersProgress(response) {
   return (dispatch, getState) => {
     const { curBlocks, neededBlocks } = getState().walletLoader;
     var newCurBlock = curBlocks + response.getFetchedHeadersCount();
+    console.log("curBlocks " + curBlocks);
+
     if (curBlocks == 0) {
       newCurBlock += response.getFirstNewBlockHeight();
     }
-    if ( newCurBlock > neededBlocks ||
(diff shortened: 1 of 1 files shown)


jcvernaleo:
Doesn't pass npm run lint

jolan:
Ah well I was going to remove those debugs anyway if we agree this is proper but fixed anyway.

jolan:
After discussion, this is just kind of side effect of not having all the proper infrastructure in place and will be fixed via another method.

PR #171, opened 2017-01-16
Simplify grpc build instructions.

### LINKED ISSUE #170: Simplify grpc build instructions.
Electron support is in grpc master (but not in a release) so we can use a simpler build step for it.

Local build will still be needed (some #1 is still relevant) until there is a release with this.

jcvernaleo:
Since electron support went into grpc we no longer need
to use a special branch or build instructions.

Until there is a new GRPC release local building is still necessary.

Closes #170

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,12 +78,8 @@ cd code
 git clone https://github.com/decred/decrediton.git
 git clone https://github.com/grpc/grpc
 cd grpc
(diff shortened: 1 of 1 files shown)


jolan:
Clean checkout and dev/release builds worked fine for me on Linux.

PR #172, opened 2017-01-16
Sidebar/Header revamp, plus error page on wallet ping error

alexlyp:
First pass to get sidebar and header looking more like what the design spec wants.

Also add an error path when grpc Ping attempts return an error.

Currently it just lands the user on a page to tell them to reload, but will send them back through the
get started loop and have them reconnect all grpc services instead.

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -178,7 +178,10 @@ export const GETPING_FAILED = 'GETPING_FAILED';
 export const GETPING_SUCCESS = 'GETPING_SUCCESS';
 
 function getPingError(error) {
-  return { error, type: GETPING_FAILED };
+  return (dispatch) => {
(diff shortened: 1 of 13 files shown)


jcvernaleo:
It seems to work (shows the error page) if wallet dies after the initial 8 step pages.  But, if wallet dies or is killed during fetch headers for example there is still no feedback to the user.

jcvernaleo:
I really like the balance on the sidebar but the block count is a little confusing looking since it is just a number with no context.

jcvernaleo:
npm run lint seems to show a bunch of errors, many of which you need to manually take care of (not all are just spaces).

jcvernaleo:
Also on the error page, maybe it should say 'restart' instead of 'reload'.

PR #173, opened 2017-01-17
Fixed for Windows.

### LINKED ISSUE #122: Produce and test windows builds
Kind of self explanatory.  May require a windows box but not sure (see #32 for details on that).

jcvernaleo:
Need to construct the path for Windows rather than using env variables
the way were we doing it before.

Move all functions that need path and config info to a single file,
config.js.

Detached should be false, esp for windows.

Don't use fd 4 that doesn't seem to work on win with wallet.

Closes #122

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -1,3 +1,7 @@
+import fs from 'fs';
+import path from 'path';
+import os from 'os';
+
 export function getCfg() {
   const Config = require('electron-config');
   const config = new Config();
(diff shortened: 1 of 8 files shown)


jcvernaleo:
Waiting for it to actually fully sync before I want a review on this.

app/config.js
```diff
@@ -49,3 +53,66 @@ export function getCfgPath() {
...
+
+export function getCert() {
+  var cert = '';
+  var cfg = getCfg();
+  if (cfg.cert_path != '') {
+    return(cfg.cert_path);
+  }
+  var certPath = '';
+  if (os.platform() == 'win32') {
+    certPath = path.join(os.homedir(), 'AppData', 'Local', 'Decrediton', 'rpc.cert');
```
jrick: probably want `process.env.LOCALAPPDATA`.

jcvernaleo: That is what I originally had and it (and process.env.APPDATA) were both undefined when I ran it in the electron app (but had the value I expected with standalone node).

jrick: oh interesting, so that env var is missing when in production mode? :/

jcvernaleo: It does seem that way (which is super annoying).

jrick: could use a comment here if that's the case so others know why it was done this way

jcvernaleo: comment added

jcvernaleo:
Okay, it synced up all the way for me.  So @alexlyp or @jrick feel free to have at it.

PR #181, opened 2017-01-18
Handle dcrd or dcrwallet startup errors.

### LINKED ISSUE #108: Need indication in GUI when dcrd or dcrwallet can't start
If dcrd or dcrwallet cannot start (builds only, dev doesn't auto start) decrediton gives no indication of what is going on in the GUI.  The terminal shows `error starting drcd: Error: spawn dcrd ENOENT` which is good, but a user who doesn't start from a terminal will be totally lost.

jcvernaleo:
If either process fails to start (in production mode only of course)
let the user know then stop (since there is no way to continue from
there).

Only shutdown processes that are currently running.

Closes #108


jcvernaleo:
Note, this uses a new dependency so you need an npm install first.

PR #193, opened 2017-01-23
Add qr code to receive page.

### LINKED ISSUE #188: Recieve page needs real qr code, not black square
I'd like this to use qr-image or something similar.  It cannot use a module that includes native code.  Must be pure js.  I've had enough trouble with grpc that I do not want to add more native code dependencies unless absolutely necessary.

jcvernaleo:
Uses a pure js qr generating package (qr-image).

Closes #188

### DIFF
--- a/app/components/Receive.js
+++ b/app/components/Receive.js
@@ -3,6 +3,7 @@ import React, { Component, PropTypes } from 'react';
 import ErrorScreen from './ErrorScreen';
 import Button from './ButtonTanel';
 import SideBar from './SideBar';
+import qr from 'qr-image';
 
 const styles = {
   body: {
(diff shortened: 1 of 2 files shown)


app/components/Receive.js
```diff
@@ -3,6 +3,7 @@ import React, { Component, PropTypes } from 'react';
 import ErrorScreen from './ErrorScreen';
 import Button from './ButtonTanel';
 import SideBar from './SideBar';
+var qr = require('qr-image');
```
alexlyp: change to:
import qr from 'qr-image'; 

for consistency

alexlyp:
works great.  nice job

PR #205, opened 2017-01-31
Add designer requested account balance hover on sidebar

alexlyp:


jcvernaleo:
Code wise this change is fine so I will OK it.

Design and usage wise, I absolutely hate it.  Needing to hover over something to get info is not something I'm okay with.  The info should either be in an existing (or new) tab or a button should reveal/hide it.

I know we aren't targeting mobile (and it isn't clear how we even would) but this would completely not work with a mobile or any kind of touch interface.  My main objection however is the one above for normal, non-mobile use.

I hope that once we add actually account support we can get rid of this.

PR #214, opened 2017-02-02
use formsy material ui for form validation

### LINKED ISSUE #70: Form validation
Implement a form validation solution that will allow us to at least have some minimal amount of input sanitization for the various forms we have (and any forms we may have).

### LINKED ISSUE #107: Random seed generation
A new seed should only be generated right when the dcrwallet needs it for setting up a new wallet.

jolan:
This uses formsy-material-ui to validate various form inputs and also moves random seed generation to happen on demand.

Closes #70.
Closes #107.


package.json
```diff
@@ -166,6 +166,8 @@
     "electron-debug": "^1.0.1",
     "eslint-formatter-pretty": "^1.1.0",
     "eslint-plugin-react": "^6.9.0",
+    "formsy-material-ui": "github:jolan/formsy-material-ui",
```
alexlyp: No external repos should be used

jolan: Yeah I was hoping a release would happen but it hasn't yet:

https://github.com/mbrookes/formsy-material-ui/issues/190

I can downgrade to the last released version and rip out some of the new features.

BTW, we have this too:

"react-tap-event-plugin": "git+https://github.com/zilverline/react-tap-event-plugin.git#master"

PR #216, opened 2017-02-02
Open help links in Default Browser.

### LINKED ISSUE #215: New electron browser windows spawned to show external webpages are broken
They don't appear to be downloading their needed scripts

jcvernaleo:
Closes #215

### DIFF
--- a/app/components/HelpLink.js
+++ b/app/components/HelpLink.js
@@ -1,5 +1,6 @@
 import Radium from 'radium';
 import React from 'react';
+import { shell } from 'electron';
 import leftLinkColor from './icons/menu-link-left-color.png';
 
 var styles = {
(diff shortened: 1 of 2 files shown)


alexlyp:
Hmmm help links don't appear to respond to clicks.  

But this could be a Qubes centric window manager issue, so if it is confirmed to work on other OS I'm fine with it.

jcvernaleo:
It depends on you having a default browser set.  If Qubes doesn't do that (which kind of makes sense as the whole idea of qubes is to not run browser and everything in one vm) than this mehtod won't work.  But this is the 'correct' electron way to do it as far as I can tell.  And it works on the systems I've tried.

alexlyp:
OK, since it appears that this is related to my OS

PR #222, opened 2017-02-04
Major refinement to match designers' views

alexlyp:


jcvernaleo:
So 4. and 5. are fixed.  2 we decided to leave as is.  1 is almost fixed (but still one lint issue.
3 still looks like a problem and I sent you a screenshot.

jcvernaleo:
Actually, when on the correct commit, all good except for the lint issues which still need fixing.

Issue #245, opened 2017-02-08
pipe issue on win10


jcvernaleo:
dcrd fails to start up with an error on the --piperx option on fd3

jcvernaleo:
This was supposed to be closed with #246 but github decided not to.

PR #246, opened 2017-02-08
Do not use pipe for dcrd.

jcvernaleo:
On win7 we had to avoid the fd 4 pipe for wallet.
win10 appears to need the same thing for dcrd (fd 3).

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -90,8 +90,11 @@ const launchDCRD = () => {
     dcrdExe = dcrdExe + '.exe';
   }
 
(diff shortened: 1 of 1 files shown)


jcvernaleo:
Closes #245

jcvernaleo:
@jrick if this works for you on Windows 10 I think we should go for it until we can look at the startup stuff (and logging) a bit better.
This does not impact other OSes.

jcvernaleo:
I'm testing something out now to see if I can get the clean shutdowns on windows as well.

PR #254, opened 2017-02-10
Improve cmd line parsing.

### LINKED ISSUE #253: error on unknown flags

jcvernaleo:
An error is now show if an undefined argument is used.

This adds a dependency on minimist.

Closes #253

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -2,6 +2,7 @@ import { app, BrowserWindow, Menu, shell } from 'electron';
 import { getCfg, appDataDirectory } from './config.js';
 import path from 'path';
 import os from 'os';
+import parseArgs from 'minimist';
 
 let menu;
 let template;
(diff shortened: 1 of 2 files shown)


app/main.development.js
```diff
@@ -10,14 +11,26 @@ let debug = false;
 let dcrdPID;
 let dcrwPID;
 
+// Not oing to make incorrect options fatal since running in dev mode has
```
jrick: misspelled going

app/main.development.js
```diff
@@ -10,14 +11,26 @@ let debug = false;
...
+// all sorts of things on the cmd line that we don't care about.  If we want
```
jrick: what extra things? why don't we care about them?

app/main.development.js
jcvernaleo: Fixed.

app/main.development.js
jcvernaleo: In dev mode, this is argv:
```
[ '/home/jcv/code/decred/decrediton/node_modules/electron/dist/electron',
  '-r',
  'babel-register',
  '-r',
  'babel-polyfill',
  './app/main.development' ]
```
so electron things that aren't a problem in production mode but things we don't ever directly need to worry about.

jcvernaleo: And to be complete, the actual command line run by ```npn run dev```
```
cross-env HOT=1 NODE_ENV=development electron -r babel-register -r babel-polyfill ./app/main.development
```

jcvernaleo:
@jrick could you take another look at this and see if it is ready to go in?

Issue #249, opened 2017-02-09
Logging


jcvernaleo:
decrediton needs to have some more debug output and it needs to be better thought out.  It should only show when using the -debug flag.

jcvernaleo:
Having lived with the -debug flag for a few months now, I'm pretty happy with what gets logged.  I think more would be overkill.  Going to close this (and we can always add individual things to the logging as they come up).

PR #259, opened 2017-02-16
Fix up some of the logging

jcvernaleo:


jcvernaleo:
This is a start on #249 but we can do more so won't close it.

PR #263, opened 2017-02-21
Add --extrawalletargs option.

### LINKED ISSUE #248: Pass extra args to wallet
For debug purposes it would be very helpful to pass extra args to wallet from the command line.  Suggested by @jrick based on Paymetheus.

jcvernaleo:
While there, add -d shortcut for --debug and document both options.

Closes #248

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -22,7 +22,9 @@ function unknownFn(arg) {
 // Allowed cmd line options are defined here.
 var opts = {
   boolean: ['debug'],
+  string: ['extrawalletargs'],
   default: { debug: false },
+  alias: { d: 'debug' },
   unknown: unknownFn
 };
 var argv = parseArgs(process.argv.slice(1), opts);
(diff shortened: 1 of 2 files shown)


app/main.development.js
```diff
@@ -179,6 +181,14 @@ const launchDCRWallet = () => {
     args.push('--testnet');
   }
 
+  // Add any extra args if defined.
+  if (argv.extrawalletargs != undefined) {
+    var extraArgs = argv.extrawalletargs.split(' ');
```
jrick: does this break passing multiple extra flags since they are separated by spaces?

jrick: actually I think that would be fine.. but any option (are there any?) that might take a string with spaces in it could be problematic

jcvernaleo: I actually needed to do that to get multiple args to work wothwise it was passing wallet malformed things when it push it to the args list for the spawn command.

jcvernaleo: That probably would break commands that take args with spaces in them but I can live with that.

jrick: e.g. `--extrawalletargs="--appdata=/tmp/path\ with\ spaces"` or some other variation of that with escaping that works :)

jcvernaleo: I think people who use paths with spaces in them should be either shunned or punished for that type of behavior but I'll see if I can find a way to accommodate that perversion.

jcvernaleo: So this would be very easy in a language with a real regex library (PCRE).  Unfortunately, js regex doesn't suppoer negative lookbehing which is the easiest way to catch that case.

How would you feel about this going in as is and them I open a new issue to allow extraargs with spaces (that way I can work on the things that we wanted this for and then worry about the less likely case later)/

jrick: go for it

PR #272, opened 2017-03-02
Fix startup of procedures

### LINKED ISSUE #271: Percent complete during block/header downloading
There are reports that the % given during a startup where the cmd line tools aren't fully synced either doesn't move at all or isn't accurate.  This should be investigated and fixed if confirmed.

alexlyp:
Remove fetch headers progress and move to post-startup.

Still need to add algos to properly create text for sidebar bottom.  timeSince and timeBack

Fixes #271

### DIFF
--- a/app/actions/NotificationActions.js
+++ b/app/actions/NotificationActions.js
@@ -15,7 +15,7 @@ function transactionNtfnsData(response) {
     const { neededBlocks } = getState().walletLoader;
     var currentHeight = 0;
     if (response.getAttachedBlocksList().length > 0) {
-      currentHeight = response.getAttachedBlocksList()[0].height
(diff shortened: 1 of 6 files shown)


jcvernaleo:
This looks good during IBD.  When started up with a synced dcrd, it never shows the 'Sync to:' text.  Discussed this in irc, just wanted to point it out here.  Once that is resolved this will be good to go.

jcvernaleo:
Pretty sure that issue was a combination of a config loading bug and my setup having weird and conflicting things from testing other things so this PR looks all good.

jcvernaleo:
Just to be complete #273 has the bug that I was seeing.

PR #299, opened 2017-03-24
Add functionality to clear account error/success messages

### LINKED ISSUE #297: Change functionality of success/error messages in header
Currently they are simply populated by whether redux state is !== null.  This is not ideal since they will be forever shown.  Need a mechanism to show the success/error message temporarily until it's no longer needed to be viewed.

alexlyp:
Clears when you click the X or when you click away from the current view.
Fixes #297


jcvernaleo:
You might want to mention in the commit message that it also clears them when you navigate away.

PR #309, opened 2017-03-30
Add --testnet and --mainnet flags

### LINKED ISSUE #308: Need a command line option to run on testnet.
We had one early on but lost it when we changed the way we handle options.  Need to resurrect it.

jcvernaleo:
Closes #308


jcvernaleo:
This one cannot go in until after #307 goes in (since it depends on the bug fix in that).

PR #333, opened 2017-04-14
Add voting gui to purchase tickets page

alexlyp:
As provided from designers' (linutee and sander) html template.

Makes use of new services, VotingService and AgendaService to populate available agendas to vote on and allow users to set what they would like their wallets to use when voting. 

Still will need to integrate into stakepool api, but the infrastructure is there.


app/components/AgendaOverview.js
```diff
@@ -0,0 +1,73 @@
...
+                  <span style={StakePoolStyles.agendaOverviewAgendaId}>
+                    Agenda ID: <span style={StakePoolStyles.agendaOverviewAgendaIdId}>#{this.props.agenda.getId()}</span>
+                  </span>
+                  <br/>
+                  <br/>
+                  <span>
+                    Once the majority of the PoW miners have upgraded (75% of the 100 most recent blocks are at the latest version) and the majority of the PoS
+                    miners have upgraded (75% of the votes in a 2016 block interval) have upgraded, the voting process begins.
+                  </span><br/>
+                  <a target="_blank" href="http://decred.org" style={StakePoolStyles.agendaOverViewReadMore}>Read more »</a>
```
jcvernaleo: is the main decred page the best thing to link to here?

alexlyp: I've removed this link for now, untli we can get a better idea where this should go,
docs.decred.org?  DCP on github?

jcvernaleo: I think the dcp makes the most sense (but if we end up listing them on docs that might be clearer).

PR #335, opened 2017-04-17
Tranasction type selection on History

alexlyp:
Now that transaction type is accessible from the GetTransactions request we can easily sort through the various transaction types.


jcvernaleo:
All good now.

PR #349, opened 2017-04-22
Dockerized build environment

### LINKED ISSUE #33: travis builds
We should be testing builds and PRs with travis like our other projects.

peterzen:
Automated build environment suitable for CI. 

```
 $ ./build-docker.sh [branch|tag] [linux|darwin|win32] [amd64|arm]
```
All options are optional. 

`branch` defaults to `master`.  
`os` defaults to `linux`.  `win32` is untested.
`arch` defaults to `amd64`.

Related to #33

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ release/
 jsconfig.json
 bin/
 .DS_Store
+.vagrant
+dist/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM ubuntu:trusty
+
+LABEL description="Decrediton builder image"
+LABEL version="1.0"
+LABEL maintainer "peter@prioritylane.com"
+
+ENV NODE_VERSION v6.9.5
+ENV TERM linux
+ENV USER build
+
+# create user
+RUN adduser --disabled-password --gecos ''  build
+
+# update base distro & install build tooling
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update && \
+    apt-get install -qy build-essential python git curl
+
+# create directory for build artifacts, adjust user permissions
+RUN mkdir /release && \
+    chown $USER /release
+
+# switch user
+USER $USER
+WORKDIR /home/$USER
+ENV HOME /home/$USER
+ENV NVM_DIR $HOME/.nvm
+
+# install & configure NodeJS
+RUN curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.1/install.sh | bash && \
+    . $NVM_DIR/nvm.sh && \
+    nvm install $NODE_VERSION && \
+    nvm use $NODE_VERSION && \
+    npm set progress=false; npm set cache-min 999999999
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+/dist/
+node_modules
+*/
--- a/build-docker.sh
+++ b/build-docker.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+set -e
+
+BUILD_TAG=${1:-master}
+BUILD_OS=${2:-linux}
+BUILD_ARCH=${3:-amd64}
+
+DCRD_RELEASE=v1.0.1
+
+DOCKER_IMAGE_TAG=decrediton-builder
+
+DIST_DIR=$(pwd)/dist
+
+DCRD_RELEASE_FILE=decred-$BUILD_OS-$BUILD_ARCH-$DCRD_RELEASE.tar.gz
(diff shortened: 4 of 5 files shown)


jcvernaleo:
Very cool, thanks!

I don't think I'm going to get to this until after our release but a few comments I can give right now anyway.

First, is there a reason you are using the latest node?  We are currently only testing and building on v6.9.5 so I would very much prefer if this used that version.  The README.md says what version we use.

Second, it looks like you are only building dcrd and dcrwallet.  decrediton also comes with dcrctl.

As for building dcrd, etc. vs using the release binaries, what are you building?  The latest master?  A tagged version?  For actual releases I prefer to use the actual release binaries but I'm not sure how we could do that in this way since they aren't posted anywhere at the time I do the build.

Finally, I have seen issues running decrediton on older Linuxes is the build machine is too new.  My solution for that has been to use Ubuntu14.04 as the main build system since that seems to produce a binary with the widest compatibility.  We will have to see how things built this way run on a variety of systems.

jcvernaleo:
I actually had a lot more to say about this than I expected once I got going.  Wasn't planning on doing that until after release :)

peterzen:
Thanks for your comments, much appreciated.  

This came out of my efforts to build a testnet enabled Decrediton as per the README – I kept running into `grpc` related dependency issues on `node` 6.5.9 and `electron` 1.4.6 and the only way I could get the build to work was using v7 vs. 1.6.5.  (The README doesn't specify the commit/tag on `grpc` so perhaps it's `master` that doesn't play nice with electron?)  

Anyways I'll have another go fixing the `dcrd` build in the script (it would indeed more sensible to use the release once there's a testnet2 binary) and see how it works.

Thanks again.

jcvernaleo:
For the record, I'm on this grpc commit (on the mac):
```
jcv@boo grpc $ git show
commit 42a95a21e31e3c604e195a9d3ac50af4941436bb
Merge: 7748faf fd00263
Author:     Muxi Yan <muxi@users.noreply.github.com>
AuthorDate: Fri Apr 21 11:06:34 2017 -0700
Commit:     GitHub <noreply@github.com>
CommitDate: Fri Apr 21 11:06:34 2017 -0700

    Merge pull request #10505 from muxi/move-parsing-queue
    
    Move response message processing to a user-specified queue

```
and on linux:
```
jcv@triforce grpc $ git show
commit cc2e048e84eaa418cab393553594a3fefb891037
Merge: 88a934e 44b2d08
Author:     Jan Tattermusch <jtattermusch@users.noreply.github.com>
AuthorDate: Sun Jan 15 13:24:41 2017 +0100
Commit:     GitHub <noreply@github.com>
CommitDate: Sun Jan 15 13:24:41 2017 +0100

    Merge pull request #9342 from emmenlau/emmenlau_use_import_libraries_for_mingw
    
    Makefile.template: use import libraries for mingw

```

peterzen:
Thank you.  The linux build does indeed work on these grpc commits (I'll try the OSX build also) – there must be a subsequent change that introduces the dependency issue I've encountered.

jcvernaleo:
Looking better.  I'm going to actually test and look at this closer right after the release.

jcvernaleo:
I probably won't get to this until monday, but could you rebase your branch against master.  There seem to be some conflicts.

peterzen:
Done, please let me know if you see any issues.

.gitignore
```diff
@@ -9,3 +9,4 @@ release/
 jsconfig.json
 bin/
 .DS_Store
+.vagrant
```
jcvernaleo: Could you add `dist/` to here as well?

Dockerfile
```diff
@@ -0,0 +1,35 @@
+
```
jcvernaleo: Is there a reason for these blank lines here?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+
+FROM ubuntu:yakkety
```
jcvernaleo: Could you switch to 14.04/trusty?  I know it is old but that is what I use for production builds since it seems to provide bins with the broadest compatability.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+LABEL maintainer "peter@prioritylane.com"
+
```
jcvernaleo: Should there be a label for the decrediton version here (or higher up)?

.gitignore
peterzen: Done

Dockerfile
peterzen: I've added a label to keep track of the image version – the Decrediton branch/tag you want to build is to be specified as a CLI argument though.

jcvernaleo:
Looks great, but it still can't be merged without a rebase.

You should just be able to do
```
git fetch --all
git checkout master
git pull
git co <YOUR BRANCH>
git rebase i master
```
I tried locally and that has no conflicts when you do it.  But due to the way github is I don't think I can do it to your PR.

PR #350, opened 2017-04-24
Added grpc commits to dev instructions

peterzen:


README.md
```diff
@@ -93,6 +93,7 @@ cd code
 git clone https://github.com/decred/decrediton.git
 git clone https://github.com/grpc/grpc
 cd grpc
+git checkout cc2e048e84eaa418cab393553594a3fefb891037  # linux, on OSX use commit 42a95a21e31e3c604e195a9d3ac50af4941436bb instead
```
jcvernaleo: This is my fault for being unclear but I don't think different ones are needed on each os.  Those just happen to be the commits I had on the two boxes I checked.  I'd like us to standardize on just one of them.

peterzen: `cc2e048e84eaa418cab393553594a3fefb891037` worked for me, I ran into the `cares` dependency issues on `42a95a21e31e3c604e195a9d3ac50af4941436bb`

jcvernaleo: Then let's set `cc2e048` as the one for now and not even mention `42a95`.

PR #369, opened 2017-05-01
Add a retry start rpc connection button

alexlyp:
Also add a localhost attempt after 127.0.0.1 just in case the user's setup prefers that way.


jcvernaleo:
After the lint fix this is good to go.

PR #411, opened 2017-06-01
Update API to check new ImportScript response fields

### LINKED ISSUE #410: Update wallet api for new ImportScript response fields
update to 4.12.0

alexlyp:
Now before any manual purchase ticket attempt can occur, the wallet must verify that the associated script is redeemable by the wallet and the p2sh matches the stakepool settings.

Closes #410


jrick:
This needs to also verify that the p2sh voting address matches that of the one returned by ImportScript, otherwise the multisig script redeemable check is worthless (could give a valid multisig script but assign voting rights to a different address).

jcvernaleo:
Doesn't the requiredVersion: '4.11.0' in app/index.js need to be updated to go with the api.proto update?

jrick:
nvm i misread the code, the checks appear correct.

However the error message for the second case it not correct in all cases (the p2sh address might match but the script not be redeemable).

Issue #465, opened 2017-06-23
Option to delete accounts ?


pdlckr:
I seem to have made one too many accounts, it would be great to see an option to delete an account with zero balance. 

I have also thought it would good to be able to transfer from one account to the other without having to grab the address for the other account and a section to see our voting rewards.

Thanks :)

alexlyp:
deleting accounts is a simple for the front-end, but some care needs to be taken in the actual dcrwallet infrastructure.  Definitely interested to see what we could come up with to fix this issue.

And I like the idea of a simple transfer button as well.

jrick:
I wouldn't implement deleting, but being able to hide accounts would be a worthwhile feature.

Accounts are deterministically created, so you can't really ever delete them, and you wouldn't necessarily want to create the same old account by accident when creating a new account.

pdlckr:
Yeah can understand the difficulty as its tied to the actual wallet. Hide feature sounds cool though.
Thanks

jcvernaleo:
What do people think hide should do in the case where the account has a balance?  Refuse to hide it (or not offer the option).  Hide it but leave balance in total?  Hide it but remove balance from total?
Also would we need a way to unhide a balance?  Or is it sufficient to say if you need to unhide, the restore from seed.

Finally, what happens if new funds are sent to a hidden account.

I do like the idea, but I want it thought through before we go coding something.  Opinions from any of you on those questions would be great.

jrick:
I'd only hide it if it has a zero balance.  If anything more is received, it should be unhidden.

pdlckr:
Agree with jrick, an optional tick box in settings could be implemented to automatically hide if there is zero balance. Show if balance arrives.

linnutee:
Yes, the hiding feature is actually present in the current design, but not yet implemented. 
The interaction model is as following: a user can hide every account under the account details. When doing so, the account (and it's balance) is no longer displayed or accessible anywhere else (drop-down fields, quick balance section, etc) besides the Accounts page where it can be revealed.

Hide / reveal account:
https://giphy.com/gifs/l4FGFu3EKIgttskXm

linnutee:
Regarding funds being sent to a hidden account, I think this needs to be worked out in Decreditons notification system (how new funds are displayed is another similar matter). 

I wouldn't recommend to force revealing an account but rather have the choice to either display notifications about hidden accounts or not.

jcvernaleo:
[...]

PR #468, opened 2017-06-27
Add show/hide account functionality

alexlyp:
Implements #465 
Allows users to show/hide accounts (not default or imported), so they won't be shown in various dropdowns and hovers.  Currently the total available balances shown to the user on Overview or History include all accounts to avoid any distress.  This can easily be changed if others believe that it's better to include in the show/hide functionality.


jcvernaleo:
Seems to work but I don't love that you can hide an account that has a non-zero balance.  I'd prefer what @jrick suggested and only allow hiding for accounts with zero balance.

Also, I suspect that the user who initially reported this wanted hidden accounts to be hidden from the accounts page, not just from dropdowns.  But this is a good first step.

PR #481, opened 2017-07-05
Add react-select package to allow for easier seed confirm

### LINKED ISSUE #414: Add selectable/autocomplete seed feature
After suggestions for community and @davecgh, we're going to easy the existing seed entry page by using an autocomplete feature.  Essentially this should help users avoid misspelling and speed up the entry process.

https://github.com/JedWatson/react-select appears to be exactly what we need

alexlyp:
Closes #414 
This is the first pass for this feature and will most likely get more testing and tweaks prior to release.


app/helpers/english.js
```diff
@@ -0,0 +1,4 @@
...
...got', 'revolver', 'spindle', 'sandalwood', 'spyglass', 'sardonic', 'stagehand', 'Saturday', 'stagnate', 'savagery', 'stairway', 'scavenger', 'standard', 'sensation', 'stapler', 'sociable', 'steamship', 'souvenir', 'sterling', 'specialist', 'stockman', 'speculate', 'stopwatch', 'stethoscope', 'stormy', 'stupendous', 'sugar', 'supportive', 'surmount', 'surrender', 'suspense', 'suspicious', 'sweatband', 'sympathy', 'swelter', 'tambourine', 'tactics', 'telephone', 'talon', 'therapist', 'tapeworm', 'tobacco', 'tempest', 'tolerance', 'tiger', 'tomorrow', 'tissue', 'torpedo', 'tonic', 'tradition', 'topmost', 'travesty', 'tracker', 'trombonist', 'transit', 'truncated', 'trauma', 'typewriter', 'treadmill', 'ultimate', 'Trojan', 'undaunted', 'trouble', 'underfoot', 'tumor', 'unicorn', 'tunnel', 'unify', 'tycoon', 'universe', 'uncut', 'unravel', 'unearth', 'upcoming', 'unwind', 'vacancy', 'uproot', 'vagabond', 'upset', 'vertigo', 'upshot', 'Virginia', 'vapor', 'visitor', 'village', 'vocalist', 'virus', 'voyager', 'Vulcan', 'warranty', 'waffle', 'Waterloo', 'wallet', 'whimsical', 'watchword', 'Wichita', 'wayside', 'Wilmington', 'willow', 'Wyoming', 'woodlark', 'yesteryear', 'Zulu', 'Yucatan'];
```
jcvernaleo: Is the name `english` (both the variable and the filename) a requirement of the package you are using?  If not it might be better to use something a little more descriptive.  Especially since the seed words would never be translated.

If it is a requirement then that is fine.

PR #482, opened 2017-07-07
Add transfer tx detection 

alexlyp:
Also update some transaction details styling


jcvernaleo:
And I see you have the lint fix there already.

PR #493, opened 2017-07-13
Update stakepool apikey entry

alexlyp:
- Added instructions for better clarity about how to enter/set up your stakepool with decrediton. 
- Added import script check when apikey is entered to avoid situations where the stakepool account is not associated with the current wallet.  Now to link a stakepool via apikey it MUST be for that wallet.  This is also the reason for the private pass request when adding the stakepool.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -332,6 +332,7 @@ function importScriptAction(votingAddress, cb) {
             }
           }
         } else {
+          setTimeout(() => { dispatch(getStakeInfoAttempt()); }, 1000);
           dispatch(importScriptSuccess(importScriptResponse, votingAddress, cb));
         }
       });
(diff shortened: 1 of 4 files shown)


app/actions/ControlActions.js
```diff
@@ -322,18 +313,29 @@ function hexToBytes(hex) {
...
+        if (err) {
+          if (!votingAddress && !cb) {
             dispatch(importScriptError(err + ' Please try again'));
           } else {
-            dispatch(importScriptSuccess(importScriptResponse, purchaseTickets));
+            var error = err + '';
+            if (error.indexOf('master private key') !== -1) {
+              dispatch(() => cb(err + ' Please try again.'));
+            } else {
+              err = err + '. This probably means you are trying to use a stakepool account that is already associated with another wallet.  If you have previously used a voting account, please create a new account (with any name) and try again.  Otherwise, please set up a new stakepool account for this wallet.';
```
jcvernaleo: I'm not sure I understand what `(with any name)` means.

jcvernaleo: I guess I understand it when thinking about it (you mean it does not have to be named voting)

jcvernaleo:
Tested all the cases we talked about.  Looks good.

PR #504, opened 2017-07-20
Update styling on incoming transaction notifications

alexlyp:
Will be sticking with material-ui Snackbar implementation for now until a better/more flexible solution can be developed.  Ideally we can add an onClick that would go to the transaction details page and also the ability to 'stack' notifications.  Currently it just shows one then disappears.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -332,7 +332,9 @@ function importScriptAction(votingAddress, cb) {
             }
           }
         } else {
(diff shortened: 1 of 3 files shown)


jcvernaleo:
I really like it.  Only issue (which we discussed in chat) is the double notification on ticket purchase.

jcvernaleo:
I think that is a lot better.  It looks less scary/confusing now.

OK

jcvernaleo:
Great, definitely ok now.

PR #507, opened 2017-07-21
Use gprc version instead of building seperately

alexlyp:


jcvernaleo:
Works great on linux.  Doesn't seem to work for me on osx but I'll look into it.

jcvernaleo:
Once I cleared everything out and tried again on OSX this actually works.

PR #510, opened 2017-07-24
Add daemon_rpc_host_testnet and use for start rpc

alexlyp:


app/actions/WalletLoaderActions.js
```diff
@@ -181,6 +181,12 @@ export function startRpcRequestFunc(localHost) {
   } else {
     rpcport = cfg.get('daemon_port');
   }
+  var daemon_host = '';
+  if (cfg.get('network') == 'testnet') {
```
jcvernaleo: I would probably put this as a function in config like https://github.com/decred/decrediton/blob/master/app/config.js#L208 called DaemonRPCHost or something like that so you just call it and all the checking for network happens in the function.

PR #511, opened 2017-07-25
Add logging package and do first pass in adding to main.dev.js 

alexlyp:


app/main.development.js
```diff
@@ -56,44 +57,50 @@ if (process.env.NODE_ENV === 'development') {
 app.setPath('userData', appDataDirectory());
 var cfg = getCfg();
 
+var logger;
 if (debug) {
-  console.log('Using config/data from:', app.getPath('userData'));
+  logger = new (winston.Logger)({
```
jcvernaleo: Is this the normal winston way to do it?  If so that is fine, but it seems like (if it works) to define transports with the file based one initially and then if (debug) append the console one (and the do the logger = new line a single time.

But again, if the current way is the idiomatic way then stick with it.

PR #520, opened 2017-07-27
Fixes #506

### LINKED ISSUE #506: Send page looks to overflow once try a value above a certain value
Enter 111111111 into the amount field, then slowly try to add more ones.  At some point it throws an Assertion err.  Looks like this is due to an overflow possibly.

sndurkin:


app/components/views/Send.js
```diff
@@ -111,6 +111,10 @@ class Send extends Component{
         updatedOutputErrors[i].amountError = '*Please enter a valid amount (> 0)';
         checkErrors = true;
       }
+      if (updatedOutputErrors[i].amount > 9223372036854776000) {    // 2^63
```
alexlyp: Why not use Number.MAX_SAFE_INTEGER instead here?

Usually I don't like to use magic numbers like that

sndurkin: Agreed, fixed in the latest commit.

PR #555, opened 2017-08-04
Add less to webpack

### LINKED ISSUE #546: Integrate LESS css preprocessor
As part of the overall refactor as suggested by @go1dfish in #528, we'll be moving back to using css instead of inline.  And with that we'll be using LESS to help keep the files better organized and easy to update

sndurkin:
Closes #546


alexlyp:
Runs well and code looks good. Only issue I see off hand is the hide/show account and rename account buttons are floated to the left instead of the right

Issue #528, opened 2017-07-29
Designer Maintainability Refactor Plan Proposal


go1dfish:
Below is an outline of a new directory/component structure for decrediton and a general refactor
plan to pave the way towards a more maintainable, faster, designer friendly decrediton.

The index components can be thought of as containers.  
This is where any component logic will go including connecting to redux.

Any component that is not an index component should (eventually) be simple enough to convert
to straight up functions in this form:

    const MyPresentationalComponent = ({ className, ...props }) => (
      <div {...{ className }}>
        ...
      </div>
    );

The components marked (styleonly) seem to only exist as a vehicle for Radium's inline styling.

From talking on slack we all want to eventually abandon inline styling.

That means these components can eventually go away and be replaced by solid className use in
the components that currently reference them.

# Component Structure

 * components
   * views
     * Accounts
       * index (redux)
       * Page
       * Account (Formerly AccountRow)
         * index
         * Row
     * GetStarted
       * index (redux)
       * Page
     * Help
     * History
       * index (redux)
       * Page
     * Home
       * index (redux)
       * Page
     * Receive
       * index (redux)
       * QRCode
       * Page
     * Send
       * index (redux)
       * Page
     * Settings
       * index (redux)
       * Page
     * StakePool (This probably needs a deeper breaking up internally after this initial pass)
       * index (redux)
       * Page
     * TxDetails
   * AgendaCard
   * AgendaClose (styleonly)
   * AgendaOverview
     * index
     * Overview
   * AutoBuyerSwitch
   * Balance (redux)
   * BalanceOverviewInfoButton
   * BalanceOverviewInfoModal
   * ChangePassphraseModal
     * index
     * Modal
   * CopyToClipboardButton (Edit: originally planned to split this but I don't think it's necessary)
   * CreateWalletForm
     * index
     * CreateWallet
     * Continue
   * ErrorScreen
   * Header
     * index (redux)
     * GetStarted
     * Header
   * HelpLink
   * HideShowButton
   * ImportScriptModal
     * index
     * Modal
   * KeyBlueButton
   * ManagePoolsButton
   * MenuLink (styleonly)
   * NewExistingSeedToggle
     * index
     * Toggle
   * NumTicketsInput
   * PassphraseModal
     * index
     * Modal
   * PurchaseTicketInfo
   * PurchaseTicketsInfoButton (styleonly)
   * SeedEntry
     * index
     * Entry
   * ShowError
   * Sidebar
     * index
     * Bar
[...]

PR #557, opened 2017-08-05
Split ChangePassphraseModal into logic and presentation components

go1dfish:


app/components/ChangePassphraseModal/index.js
```diff
@@ -0,0 +1,125 @@
...
+          submitPassphrase,
+          cancelPassphrase
+        }}
+      />
+    );
+  }
+
+  resetState() {
+    this.state = this.getInitialState(); // clear password data from memory
+    document.getElementById("passphrase").value = "";
```
go1dfish: I copied this logic that was there explicitly clearing these values, but I'm not sure it's necessary as resetting the state should also clear the inputs.

This makes the presentation components brittle if they change id attribute and I think that's bad.

If there is no specific justification for these value clears I would suggest we remove them.

alexlyp: I'll test with and without these clears, if functionally it is the same then I'm cool with them going.

alexlyp: OK after testing these getElementByIds can all go.

PR #563, opened 2017-08-08
Convert several components from inline styles to CSS

sndurkin:


app/components/AgendaCard.js
```diff
@@ -1,207 +1,45 @@
...
+    }
+    else {
+      return <div className="agenda-card-indicator-pending">In Progress</div>;
     }
   }
-}
 
-export default Radium(AgendaCard);
\ No newline at end of file
+}
```
alexlyp: Need an extra carriage return here

app/components/AgendaClose.js
```diff
@@ -1,15 +1,12 @@
...
-      <a style={StakePoolStyles.agendaOverviewTitleClose} onClick={this.props.onClick}></a>
+      <a className="agenda-overview-title-close" onClick={this.props.onClick}></a>
     );
 
   }
-}
-
-export default Radium(AgendaClose);
\ No newline at end of file
+}
```
alexlyp: and here

PR #569, opened 2017-08-09
Add box shadow to modals

alexlyp:
Also define in the main.less for later use when the rest of the components get used with less styling

### DIFF
--- a/app/style/ChangePassphraseModal.less
+++ b/app/style/ChangePassphraseModal.less
@@ -14,6 +14,7 @@
 .change-passphrase-modal-section {
     padding: 20px;
     background-color: #fff;
+    box-shadow: @modal-dropshadow;
 }
 
 .change-passphrase-modal-section-heading-sub-section {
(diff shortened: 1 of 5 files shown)


app/style/ChangePassphraseModal.less
```diff
@@ -14,6 +14,7 @@
 .change-passphrase-modal-section {
     padding: 20px;
     background-color: #fff;
+    box-shadow: @modal-dropshadow;
```
sndurkin: [Mixins](http://lesscss.org/features/#features-overview-feature-mixins) might be a better way to handle this specific case, because you could do something like this:

    .mixin-modal: {
        box-shadow: 0px 13px 21px rgba(0, 0, 0, 0.21);
    }
    
    .change-passphrase-modal-section {
        .mixin-modal;
        /* other rules */
    }

And if in the future we want to customize all modal dialogs in some additional way, we would just add to the `mixin-modal` class.

alexlyp: awesome i'll do that, thx

PR #586, opened 2017-08-14
Refactor TxHistory/TxRow into logic and presentation components

go1dfish:


alexlyp:
Looks like on Regular transactions transaction detail's header area it isn't populating it correctly.  It's showing a green ticket then no words.  On master it shows the grey wallet icon and the amount sent/received/transferred

PR #587, opened 2017-08-14
Refactor/Split AccountsPage into logic/presentation components

go1dfish:


alexlyp:
Looks fine, but the ordering of the accounts on the page is now backwards.

Default should always be at the top (0) and imported(max account num) should be at the bottom

Issue #588, opened 2017-08-14
Proposal: Spending Accounts, Staking Account and Governance


go1dfish:
(This proposal is still somewhat of a work in progress and feedback is appreciated)

We should split the "Tickets" page of decrediton into "Staking Account" and "Governance" pages.

The existing "Accounts" page would become "Spending Accounts".

This should not require any protocol, consensus or voting changes,
simply abstracting away the underlying mechanisms in a simpler way to the common user.

Implementation wise the auto ticket buyer will always active for this logical "staking account"
and withdrawals will be achieved by internally adjusting the minimum available balance.

The user experience becomes a deposit into this logical account with a recommended/required minimum
(the current ticket price).

A blockchain transaction for this "deposit" is not necessary, but may be beneficial for consistent
user experience and display in transaction history.

The staking account page should describe staking in purely financial terms.

Governance should be moved to a separate page, and this more advanced page should describe
the relationship between staking (account) and voting/tickets, as well as linking the user to
detailed documentation/videos.

Requirements for Staking Account page:

- [ ] Calculate Staking account balance as locked tickets + immature rewards + available funds devoted
- [ ] Clearly warn the user that staking account funds will be irrevocably locked for up to 4 months
- [ ] Suggest a minimum effective deposit (ticket price, or ticket price - free balance?)
- [ ] Estimate Monthly/Yearly % returns based on ticket price
- [ ] Estimate variance in returns based on Staking Account Balance
- [ ] Provide a means of deposit, walk through api pool setup on first deposit
- [ ] Provide a means of withdraw (reduced ticket buying) to spending account
- [ ] Display a history of staking returns geared towards finance management and tax compliance

Requirements for Governance page:

- [x] Provide prominent display of the constitution within decrediton
- [x] Display current proposals and their status (before any staking or pool api key entry)
- [x] Allow setting proposal preferences (before any staking or pool api key entry)
- [x] Display ticket voting/purchase history potentially with more detail than Staking Account
- [ ] Describe how Decred's proof of stake governance works and the relation to the staking account

alexlyp:
I really like the direction and the though behind this proposal.  Ideally we should be trying to seperate the 2 major pieces of functionality (normal transaction processing and staking)

@linnutee any care you give some feedback to this proposal as well?

PR #626, opened 2017-08-27
Split SendPage into logic and presentation components

go1dfish:
Continuation of https://github.com/decred/decrediton/issues/588

This also improves the send page experience a bit by eliminating the full page loader that would show briefly when constructing transactions.

I still keep the send button disabled while constructing but the very brief loader felt kinda buggy IMO so I have it only show for signing/sending.


alexlyp:
Hit a few issues during testing:

1) Amount field isn't properly checking the input.  I was able to put in letters and special characters with no error being shown.  ("1.s" was accepted and a transaction was able to be constructed, it looks like it just uses whatever was in there previously.)
2) No success message on successful tx publish.

PR #631, opened 2017-08-29
Refactor GetStarted screens so that startup looks much smoother

### LINKED ISSUE #625: Startup screen UIs are inconsistent
The startup screens have inconsistent UIs, making them look bad when loaded. Here are a couple things I noticed:

* Error messages are shown for a split second when there's not really any error
* Loading icons are shown for a split second, then replaced by a new loading icon in another screen
* Inputs are displayed on some screens for a split second before it disappears
* One screen (3 and 4 in the list below) is rendered twice.

When starting Decrediton with an existing wallet in dev mode, here are the following different screens that are displayed on startup:

1. image

2. image

3. image

4. image

5. image

6. image

7. image

8. image

sndurkin:
Fixes #625

I tested this on Windows with the following flows:

1. Create wallet with generated seed
2. Create wallet with existing seed
3. Create wallet with public passphrase, then restart dcrwallet and decrediton to get the public passphrase prompt
4. Open with existing wallet
5. Open without dcrwallet running


app/actions/WalletLoaderActions.js
```diff
@@ -45,38 +45,40 @@ export function walletExistRequest() {
...
-            }
-          }
-        });
+    loader.walletExists(request, function(error, response) {
+      if (error) {
+        dispatch({ error, type: WALLETEXIST_FAILED });
+      } else {
+        dispatch({response: response, type: WALLETEXIST_SUCCESS });
+        if (response.getExists()) {
+          setTimeout(() => dispatch(openWalletAttempt("public", false)), 1000);
```
sndurkin: @alexlyp I fixed a bug here; this wasn't wrapped in a function so it's actually calling `dispatch()` immediately. Can the `setTimeout()` call be removed?

PR #642, opened 2017-09-05
Change start up procedures

alexlyp:
This PR attemps to streamline the startup procedure to let the user have more control of how the daemon and wallet start in the background.

Currently, these processes (dcrd and dcrwallet) are started in the main process of decrediton.  Now, they are signalled to start from the renderer processes (React action).  This allows us to begin making more intuitive design choices and adds flexibility to the startup procedure.  With this change, the user should be fully made aware of the syncing happening in daemon and hopefully will avoid confusion in terms of not seeing the proper balances once loaded into wallet.

In the coming releases, we can now add remote daemon and wallet solutions for users with non-default setups.


app/components/views/GetStartedPage/DaemonLoading/Form.js
```diff
@@ -0,0 +1,60 @@
...
+import ReactToolTip from "react-tooltip";
+
+const DaemonLoadingFormHeader = ({
+  startupError,
+  getDaemonStarted,
+  getCurrentBlockCount,
+}) => (
+  <Header getStarted
+    headerTitleOverview="Setting up Decrediton"
+    headerMetaOverview={getDaemonStarted ? getCurrentBlockCount == null ? "Daemon started, waiting for rpc..." : "Blockchain syncing" : "Starting daemon"}
```
sndurkin: I would capitalize "rpc" here as it's an acronym, and to be consistent with the other startup message.

app/main.development.js
```diff
@@ -175,11 +179,80 @@ const installExtensions = async () => {
...
 
+const {ipcMain} = require("electron");
+
+ipcMain.on("start-daemon", (event, arg) => {
+  if (dcrdPID) {
+    logger.log("info", "dcrd already started " + dcrwPID);
+    event.returnValue = dcrdPID;
+    return;
+  }
+  logger.log("info", "launching dcrd with " + arg);
```
sndurkin: The output for this is: `info: launching dcrd with [object Object]` so we should maybe use `JSON.stringify()` here.

app/main.development.js
```diff
@@ -175,11 +179,80 @@ const installExtensions = async () => {
...
+  event.returnValue = !daemonReady;
+});
+
+ipcMain.on("start-wallet", (event, arg) => {
+  if (dcrwPID) {
+    logger.log("info", "dcrwallet already started " + dcrwPID);
+    event.returnValue = dcrwPID;
+    return;
+  }
+  logger.log("info", "launching dcrwallet at " + arg);
```
sndurkin: Same comment as for dcrd.

app/actions/DaemonActions.js
```diff
@@ -0,0 +1,95 @@
...
+            minutesLeft = "<1";
+          }
+          var updateTimeLeftEstimate = "Estimated time remaining: " + minutesLeft + " minutes" ;
+          dispatch({currentBlockCount: parseInt(updateCurrentBlockCount), timeLeftEstimate: updateTimeLeftEstimate, type: DAEMONSYNCING_PROGRESS});
+        } else if (updateCurrentBlockCount !== 0) {
+          var time = new Date();
+          dispatch({currentBlockCount: parseInt(updateCurrentBlockCount), timeStart: time, blockStart: parseInt(updateCurrentBlockCount), type: DAEMONSYNCING_START});
+        }
+      }
+      await sleep(10000);
```
sndurkin: 10 seconds may be a little long here; I can see when dcrd starts listening and sometimes it happens right after a check.

app/main.development.js
```diff
@@ -346,8 +346,8 @@ const launchDCRWallet = () => {
 
   dcrwallet.on("close", (code) => {
     if (code !== 0) {
-      logger.log("error", "dcrwallet closed due to an error.  Check dcrd logs and contact support if the issue persists.");
-      mainWindow.webContents.executeJavaScript("alert(\"dcrwallet closed due to an error.  Check dcrd logs and contact support if the issue persists.\");");
+      logger.log("error", "dcrwallet closed due to an error.  Check dcrwalet logs and contact support if the issue persists.");
```
sndurkin: Typo

PR #644, opened 2017-09-06
Send all functionality

### LINKED ISSUE #603: Create a send all from account button
Currently it's hard for users to easily create a transaction that consumes all of the funds from an account. There should be simple way to create this transaction for users

alexlyp:
Now if a user wants to send everything from a given account then they can select Send All and then enter in the destination address.  It will then construct a transaction that will consume all utxos in the account and the user will then have to enter the private passphrase as with normal transactions.

Closes #603


go1dfish:
Testing looks good here

PR #645, opened 2017-09-07
Split Ticket Page into multiple components

go1dfish:
Stake pool interactions have not been heavily tested yet, everything else should be.

This is a continuation of https://github.com/decred/decrediton/issues/528


alexlyp:
When changing voting preference, the button should grey out when updated.  But seems to stay enabled after updating the choice and is still disabled on the original setting

Issue #133, opened 2017-01-06
Allow for graceful window resizing across various size groups


alexlyp:
Similar to bootstrap xs, sm, md, lg col grouping.  When width degrades past certain px range then re-render to accommodate new size

alexlyp:
With the incoming refactor in #528 and the reintroduction of .css we should be able to accomodate 3 different media groups, tiny, medium and extra large.

alexlyp:
@matheusd has done an initial pass for this issue, but we will still keep it open until we have done proper styling for the various sizes.

vctt94:
This is being worked with responsive views

PR #670, opened 2017-09-18
Allow redimensioning of app

### LINKED ISSUE #649: Make decredition fit smaller screen sizes
A lot of people (myself included) use laptops and monitors that are 1366x768. Due to this the Decredition application is larger than the desktop size by about 20-40 px. I would recommend making the total application shorter vertically by 40 px so that it nicely fits on those desktop sizes.

Another method would be to change what happens when you resize the application vertically. Right now when I resize the window and make it smaller the "total balance" at the bottom gets hidden, along with the "balance" and logo at the top.

matheusd:
This is the first set of improvements to allow dynamic resizing of the app's main window.

All pages have been changed to use the same set of CSS classes for the general layout, and those classes were rewritten to allow redimensioning. Most of the in-page classes also had to be modified to allow better reflow on resizing.

There are still a large number of small issues on resizing, but the app is at least usable.

Fixes #649. Works towards fixing #133.


alexlyp:
Looks really good so far.  Couple issues I see at first glance:

1) GetStarted content background is white instead of the dark blue as it is on master.
2) unit descriptor (DCR/Atoms) on send page is in front of Amount instead of behind.

I'll start to review codewise.

alexlyp:
So looks like with the addition of page-content to the DaemonLoading  makes the logo hidden and I still see some of the later steps as having a white background.

dajohi:
needs rebase.

matheusd:
There is still one last loading screen with white background (the one shown when loading transactions) but that one seems to be white on master as well (should I change that to the dark background as well)?

@alexlyp,  I don't see the logo hidden on any page. Is it hidden on a single page or on all pages for you?

alexlyp:
/home/user/git/src/github.com/decred/decrediton/app/components/views/GetStartedPage/DaemonLoading/Form.js

I inspected elements and it looks like 

        <DecredLoading hidden={!props.isProcessing} className="get-started-loading" />

is on top of 

    <div className="get-started-content-new-seed page-content">

PR #679, opened 2017-09-19
Move remaining redux connectors out of components

go1dfish:

With this PR, no redux connector logic remains in components


matheusd:
Don't know if it's just me, but it seems to have broken wallet creation. I get a TypeError on the console.

alexlyp:
Same result for me on wallet creation

matheusd:
Seems ok now

PR #684, opened 2017-09-20
Add rescan progress to sidebar

### LINKED ISSUE #681: Put rescan progress in sidebar instead of overview
- Also fix the rescan on import script and instead issue an independent rescan request that can be captured as a stream

alexlyp:
Also add explicit rescan request to import script request. This now allows us to capture rescan progress updates and give feedback to the user.
Fixes #681

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -199,7 +199,7 @@ export function importScriptAttempt(passphrase, script, rescan, scanFrom, voting
   var request = new ImportScriptRequest();
   request.setPassphrase(new Uint8Array(Buffer.from(passphrase)));
   request.setScript(new Uint8Array(Buffer.from(hexToBytes(script))));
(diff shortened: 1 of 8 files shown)


matheusd:
I'm testing it, but the rescan from the "import script" is not working (the rescan is not triggered after importing and the modal is not closed).

I'm still trying to figure out if I'm doing something wrong in the process here.

matheusd:
Got it!
I had to add the following for the rescan to proceed. On line:

https://github.com/decred/decrediton/blob/591fd91fd699b9d04bb2f7159840e1cba0715712/app/components/views/TicketsPage/index.js#L76

Add
`onImportScript: null,`

Otherwise the dispatcher is called directly, without going through the component's onImportScript callback.

matheusd:
Seems fine now.

PR #686, opened 2017-09-21
Show pending transactions on history page

### LINKED ISSUE #640: When receiving a payment and it is pending, it shows up in overview but not history.
I just received my first payment, I checked the history tab to see it, nothing was there.  Other parts of the application however reflected that a pending payment had arrived.

I would say it might be a small usability issue, but one for UX people if we have them to think about.  You can see in this example, I have a balance of 2.69 DCR, but no history of transactions.

image

matheusd:
Fixes #640

### DIFF
--- a/app/components/TxHistory.js
+++ b/app/components/TxHistory.js
@@ -12,7 +12,9 @@ const TxHistory = ({ mined=[], unmined=[], showTxDetail }) => (
       {mined.map(tx => (
         <TxHistoryRow
           key={tx.txHash}
-          {...{ showTxDetail, tx }}
+          pending={tx.txTimestamp ? false : true}
+          showTxDetail={tx.txTimestamp ? showTxDetail : false}
+          {...{ tx }}
           date={tx.txTimestamp ? dateFormat(new Date(tx.txTimestamp*1000), "mmm d yyyy, HH:MM:ss") : null}
         />
       ))}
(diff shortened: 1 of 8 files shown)


alexlyp:
And Warning that I noticed in the past that also seems to cause issues when I select All in History: 

Warning: flattenChildren(...): Encountered two children with the same key, `0:$b4df8ea063399995c82b3837fcca8e13ae6ab99c86a9950067f36cd9410290d4`. Child keys must be unique; when two children share a key, only the first child will be used.  And then it only shows 4 txs and seems to mangle subsequent page overs.

alexlyp:
Strangely though it doesn't seem to happen every time.  I'll try and investigate that issue more and let you know about it.

alexlyp:
Looks like it works great for tickets, revokes and regular.  Only issue I see is that account [...Pending] floats a bit to the left when the page gets expanded past normal width so they are offline from [Confirmed].

alexlyp:
Could you also now add all unmined txs to the overview page instead of just regular?

matheusd:
On the warning: I encountered that a few times while developing, but (in the cases I got) was because I was concatenating the unmined transactions twice (which shouldn't be happening now but I'll try and double check). Maybe some specific pending transaction is giving that problem? I only got a vote pending transaction once to test.

I'll work on the alignment and on showing all types of unmined on overview. I didn't add *all* before because I was maintaining consistency between the overview and the default view on the history (which shows only regular).

matheusd:
Question: Should the overview list all (recently) mined transactions as well (instead of just the regular mined)?

alexlyp:
Yes, sorry I should been clearer.  I think we want all types of transactions + mined and unmined on overview

Issue #669, opened 2017-09-15
Add features/design to About(version) popup window


alexlyp:
The first iteration of the About window currently just shows hard coded html to display the versions of the various software.  This should be done dynamically in 2 ways: 

1) Pull the version numer from package.json some how and then use that for decrediton versions.
2) Query dcrd and dcrwallet --version to get actual binary version and use there.

We should also try and spruce up the design.

alexlyp:
Version numbers are now properly pulled, Still would like someone to spruce up the design a little bit.

alexlyp:
I believe everything we want is currently available in the design, but I'll bump this to the future milestone since this is completely acceptable for release.  I'll wait until @linutee has a chance to do a mockup.

matheusd:
Done via #1528

PR #688, opened 2017-09-21
Read version from binaries and display in about window

matheusd:
Fixes first part of #669.

During development, Decrediton version will **not** show up correctly because of [issue 7085 of electron package](https://github.com/electron/electron/issues/7085).

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -413,7 +413,7 @@ const readExesVersion = () => {
       continue;
     }
 
-    let decodedLine = versionLine.match(/\w+ version ([^\s\+]+)/);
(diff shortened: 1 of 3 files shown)


app/main.development.js
```diff
@@ -390,6 +390,40 @@ const launchDCRWallet = () => {
   return dcrwPID;
 };
 
+const readExesVersion = () => {
+  let spawn = require("child_process").spawnSync;
+  let args = ["--version"];
+  let exes = ["dcrd", "dcrwallet", "dcrctl"];
+  let versions = {
+    decrediton: app.getVersion()
```
alexlyp: This is pulling the electron versionl, which explains why it is showing 1.4.15 

Should be showing 1.1.0 instead

PR #707, opened 2017-09-25
Change to vX.Y.Z version label

alexlyp:


oktapodia:
The leading `v` was removed into the [semver 2.0.0](http://semver.org/)

alexlyp:
gotcha good point.  closing

Issue #709, opened 2017-09-26
Disabled buttons still work


matheusd:
Clicking any disabled buttons still make them call the onClick() handler. 

I'll fix this together with issue #706.

linnutee:
Similar case with Send button, i'll just mention it here. 
- Send is currently visually disabled until information is entered and user clicks elsewhere (e.g. outside amount field). 
- Instead it should change to enabled state once the information is entered (w/o having to click elsewhere).

Thoughtscript:
I believe this ticket got closed per: https://github.com/decred/decrediton/pull/710/commits/6fcecb9f8aeb63be0c8715758b7b9cab8c5f4a1f

PR #710, opened 2017-09-26
Disable rescan button

### LINKED ISSUE #706: Need to grey out rescan button during rescan attempt

matheusd:
Fixes #706 and #709.

### DIFF
--- a/app/components/KeyBlueButton.js
+++ b/app/components/KeyBlueButton.js
@@ -35,7 +35,9 @@ class KeyBlueButton extends React.Component {
     if (this.props.disabled) {
       return;
     }
-    this.props.onClick(e);
+    if (this.props.onClick) {
+      this.props.onClick(e);
+    }
   }
 }
 
--- a/app/components/KeyBlueButton.js
+++ b/app/components/KeyBlueButton.js
@@ -32,10 +32,7 @@ class KeyBlueButton extends React.Component {
   }
 
   onClick (e) {
-    if (this.props.disabled) {
-      return;
-    }
-    if (this.props.onClick) {
+    if (!this.props.disabled && this.props.onClick) {
       this.props.onClick(e);
     }
   }
--- a/app/components/KeyBlueButton.js
+++ b/app/components/KeyBlueButton.js
@@ -1,7 +1,9 @@
 // @flow
 import React from "react";
+import { autobind } from "core-decorators";
 import "../style/MiscComponents.less";
 
+@autobind
 class KeyBlueButton extends React.Component {
   render() {
     let className = !this.props.disabled ? "key-blue-button"
@@ -22,11 +24,18 @@ class KeyBlueButton extends React.Component {
           style={style}
           type={this.props.type}
           disabled={this.props.disabled}
-          onClick = {this.props.onClick}
+          onClick = {this.onClick}
           hidden={this.props.hidden}>
         {this.props.children}
       </div>
     );
+  }
+
+  onClick (e) {
+    if (this.props.disabled) {
+      return;
+    }
+    this.props.onClick(e);
   }
 }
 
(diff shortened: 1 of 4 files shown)


sndurkin:
Is the Rescan button the only case where a disabled button is still clickable?

matheusd:
No. Try clicking the button on the settings page, it will dispatch the settings_changed event.

sndurkin:
Okay, are you planning to address the other cases in this PR or should we leave #709 open for now?

matheusd:
The fix is global on keyBlueButton, so this already fixes everywhere it is used. Traditional buttons are not affected by the bug.

What I didn't do but will try and find additional instances is of divs being used as buttons (instead of native html buttons).

app/components/KeyBlueButton.js
```diff
@@ -22,12 +24,21 @@ class KeyBlueButton extends React.Component {
...
         {this.props.children}
       </div>
     );
   }
+
+  onClick (e) {
+    if (this.props.disabled) {
+      return;
+    }
+    if (this.props.onClick) {
```
oktapodia: The `disabled` and `onClick` can be done in a single `if` statement

app/components/PurchaseTickets/Form.js
```diff
@@ -216,10 +217,18 @@ const PurchaseTicketsForm = ({
         </div>
       )}
 
-      <KeyBlueButton
-        className="stakepool-content-import-script-button"
-        onClick={onShowImportScript}
-      >Import Script</KeyBlueButton>
+      <div className="stakepool-content-import-script-button"
+        data-html={true}
+        data-tip-disable={rescanRequest ? false : true}
```
sndurkin: Shouldn't it be disabled if `rescanRequest` is `true`? Also, it can be simplified to `data-tip-disable={rescanRequest}`.

app/components/views/HomePage/Page.js
```diff
@@ -35,9 +36,9 @@ const HomePage = ({
           <div>
             <Balance amount={spendableTotalBalance} />
             <div className="home-rescan-button-area" data-html={true} data-tip="Rescanning may help resolve some balance errors.<br><br><b>Note:</b> This scans the entire blockchain for transactions,<br>but does not re-download it.">
-              <KeyBlueButton onClick={() => rescanAttempt(0)}>Rescan Blockchain</KeyBlueButton>
+              <KeyBlueButton disabled={rescanRequest} onClick={() => rescanAttempt(0)}>Rescan Blockchain</KeyBlueButton>
             </div>
-            <ReactToolTip place="left" type="info" effect="solid"/>
+            <ReactToolTip disable={rescanRequest ? true : false} place="left" type="info" effect="solid"/>
```
sndurkin: This can be simplified to `disable={rescanRequest}`.

matheusd: It can't because react tooltip requires a boolean and rescanRequest is an object.

sndurkin: Ah, I see... what about `!!rescanRequest`?

app/components/PurchaseTickets/Form.js
matheusd: The tip is to be shown while the rescan is running. Though just the tip there is probably not ideal (maybe a "?" icon so the user knows something is up during the rescan).

matheusd: Though here the test *can* be simplified to just {rescanRequest} (no type check on data-tip-disabled)

sndurkin: Ah sorry, I should have looked a little closer.

app/components/views/HomePage/Page.js
matheusd: Well... I personally dislike double negations but if more people think it's better, then sure I can change it.

sndurkin: Nah, I'm just nitpicking.

PR #712, opened 2017-09-26
Feature/add clickable snackbar notification

oktapodia:
I'm not confident about reselect and had a lot of problem to get only one transaction, I decided to take the `transactions` like the `HistoryPage` and do a `find` on it.

- Each transaction can be accessible via `/transactions/:txHash`
- The `TxDetails` page can be accessible during the pending state
- The snackbar transaction is now clickable

### DIFF
--- a/app/components/views/TransactionPage/index.js
+++ b/app/components/views/TransactionPage/index.js
@@ -1,28 +1,13 @@
-import React, { Component } from "react";
-import { autobind } from "core-decorators";
-import { find } from "lodash";
+import React from "react";
 import ErrorScreen from "../../ErrorScreen";
 import TransactionPage from "./Page";
 import transactionPageConnector from "../../../connectors/transactionPage";
 
-@autobind
-class Transaction extends Component {
-  render() {
-    return  !this.props.walletService ? <ErrorScreen /> : (
-      <TransactionPage
-        {...{
-          ...this.props,
-          ...this.state,
-          transactionDetails: this.getTransaction(),
-        }}
-      />
-    );
-  }
-
-  getTransaction() {
-    const { transactions, params } = this.props;
-    return find(transactions.All, { txHash: params.txHash });
-  }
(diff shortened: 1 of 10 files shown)


go1dfish:
Haven't tested this yet, but the code looks good.

Reselect is not actually used for the most part because the redux state gets mutated in ways it shouldn't, but once I move the selectors over to it your code here should still be fine.

Like the approach of just searching the transaction list for now as that avoids extra rpc work.  Ideally I'd like to move that find to the selectors code instead of the component (they get passed (state, ownProps) and each function in selectorMap({..}) gets passed the same.  But I don't think that should block this merge.

app/components/views/TransactionPage/index.js
```diff
@@ -0,0 +1,28 @@
...
+        {...{
+          ...this.props,
+          ...this.state,
+          transactionDetails: this.getTransaction(),
+        }}
+      />
+    );
+  }
+
+  getTransaction() {
```
go1dfish: If this gets moved to a selector we can probably eliminate this component entirely.

something like:

    viewedTransaction = createSelector(
      [transactions, (state, { params: { txHash }}) => txHash]
      (transactions, txHash) => find(transactions.All, { txHash })
    )

The find pulled into selectors is the functional one though so you will probably use find({ txHash }, transactions.All) instead.

alexlyp:
Over all, works pretty well.  Though noticed a couple things:

1) It doesn't quite respond to my click as I'd expect, he to kind of chase it to fire the view change.
2) Now that this is essentially revealing pending transactions details pages there is some special casing that needs to be done to reflect that it is an unmined transaction.  Should be pretty clear what needs fixing in there.

oktapodia:
Updated the transaction details view when the transaction is in pending to:
- Hide the block section
- Hide the timestamp (upright)
- Set confirmation to 0

PR #714, opened 2017-09-26
Change `glide i` to `dep ensure` in the README.md

oktapodia:
The documentation is not up to date according to https://github.com/decred/dcrwallet#installation-and-updating


matheusd:
Should also add the step to install dep (same as with glide)

PR #716, opened 2017-09-28
Fix/windows build

### LINKED ISSUE #713: Including win32ipc.node into packaged build
We currently have a Node Addon module at app/modules/win32ipc.  

node-gyp properly builds it at creates a build/Release/win32ipc.node  

When working locally this is typically fine to use the absolute path: 

 https://github.com/decred/decrediton/blob/master/app/main.development.js#L306

But this won't work for builds that are used on other machines.  From what I've researched we need to include with node-loader then we can then get a relative path to be used here.  But so far unsuccessful.  

https://stackoverflow.com/questions/32832179/using-webpack-with-nodejs-native-modules

oktapodia:
Fixes #713 appveyor can now build the windows package for us

Still have an issue, the electron app does not close gracefully, the `dcrd` process still running in the background, but think it was already the case, @alexlyp can you confirm please?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,28 @@
+version: 0.1.{build}
+
+platform:
+  - x64
+
+cache:
+  - node_modules
+  - app\node_modules
+  - '%APPDATA%\npm-cache'
+  - '%USERPROFILE%\.electron'
+
+init:
+  - git config --global core.autocrlf input
+
+install:
+  - ps: Install-Product node 6 x64
+  - git reset --hard HEAD
+  - npm install npm -g
+  - npm install electron-builder@next # force install next version to test electron-builder
+  - npm install
+  - npm prune
+
+build_script:
+  - node --version
+  - npm --version
+  - npm run package-win
+
+test: off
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -53,7 +53,9 @@
       "package.json"
     ],
     "extraResources": [
-      "bin/*"
+      { "from": "bin/dcrd.exe", "to": "bin/dcrd.exe"},
```
oktapodia: Was not working with a wildcard :(

appveyor.yml
```diff
@@ -0,0 +1,52 @@
...
+
+install:
+  - ps: Install-Product node 6 x64
+  - git reset --hard HEAD
+  - npm install npm -g
+  - go version
+  - go env
+  - go get -u -v github.com/decred/dcrd
+  - go get -u -v github.com/decred/dcrwallet
+  - go get -u -v github.com/Masterminds/glide
```
alexlyp: you can remove glide references if you have the dep there.

oktapodia: But the glide is needed for `dcrd` isn't it?

alexlyp: derp you're right.  someone is slow with it.... :)

Issue #265, opened 2017-02-21
--extrawalletargs can't handle options with spaces in them


jcvernaleo:
```
--extrawalletargs='--appdata=/some/crazy/path\ with\ spaces\ in/it'
```

oktapodia:
Fixed, can you close it?

jolan:
FYI, you put "fixes issue #..." it needs to be just "fixes #..." [one of the keywords](https://help.github.com/articles/closing-issues-using-keywords/).

Closed via #729.

oktapodia:
oh! Thanks :D

PR #729, opened 2017-10-02
Fix extrawalletargs can handle spaces

oktapodia:
Fixes issue #265 Not sure the arguments splitting is needed, and this solve the issue you can have


alexlyp:
This doesn't work for me.  --extrawalletargs='--ticketbuyer.balancetomaintainabsolute=10 --enablevoting'  Causes wallet to close most likely due to malformed arguments.

oktapodia:
Weird, I'll double check that thanks!

app/main.development.js
```diff
@@ -394,11 +395,8 @@ const launchDCRWallet = () => {
   }
 
   // Add any extra args if defined.
-  if (argv.extrawalletargs != undefined) {
-    var extraArgs = argv.extrawalletargs.split(" ");
-    for (var i = 0; i < extraArgs.length; i++) {
-      args.push(extraArgs[i]);
-    }
+  if (argv.extrawalletargs !== undefined) {
+    args.concat(stringArgv(argv.extrawalletargs));
```
alexlyp: `args.concat` just isn't working for me.  i tested with `args.push` and that seems to work as expected with multiple spaced args.  Maybe it's a OS/env issue?  linux amd64 debian8 VM

oktapodia: hum, I'm on mac but should not be an issue because the concat method is a part of [es2015](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/Array/concat) I will try to add a second verification `isString` and to use the `concat` from lodash, is it working for you?

alexlyp:
Cool looks like it's working as expected now, thanks!

Issue #331, opened 2017-04-12
Add GUI for signing/verification of messages


jzbz:
GUI for `signmessage` and `verifymessage` would be a very useful feature.

alexlyp:
Ya we should be able to add something for that in the future.  Mostly just need to have it implemented in the dcrwallet rpc then it wouldn't be too bad at all to implement in the GUIs

oktapodia:
Fixed, can you close it?

PR #740, opened 2017-10-03
[Feature] Sign message & Verify message

oktapodia:
Adding the sign message #331


alexlyp:
Can you update this PR to include VerifyMessage as well now that it is available for you?  Thanks

app/components/views/SecurityPage/Page.js
```diff
@@ -0,0 +1,30 @@
...
+import "../../../style/Layout.less";
+import "../../../style/StakePool.less";
+
+const SecurityPage = ({
+  onToggleSecurityMessage,
+  children,
+}) => {
+  return (
+    <div className="page-body">
+      <SideBar />
```
alexlyp: Need to apply the same Sidebar changes that recently went in.

package.json
```diff
@@ -204,12 +204,12 @@
     "react-intl": "^2.4.0",
     "react-intl-po": "^2.1.1",
     "react-redux": "^4.4.5",
-    "react-redux-form": "^1.2.4",
     "react-router": "^3.0.4",
     "react-router-redux": "^4.0.6",
     "react-select": "^1.0.0-rc.5",
     "react-tooltip": "^3.3.0",
     "redux": "^3.6.0",
+    "redux-form": "^7.0.4",
```
alexlyp: I'd like to get @go1dfish's approval on this redux-form pkg change and the new boilerplating you've used for the sign/verify forms.  If he likes it then we can open an issue to apply these same changes to every form.

app/style/MiscComponents.less
```diff
@@ -77,9 +78,10 @@
 
 .key-blue-button:active {
   box-shadow: 0 0 0 0 rgba(0, 0, 0, 0.2);
+  background-color: #0025cc;
 }
 
-.key-blue-button-disabled {
+.key-blue-button-disabled, .key-blue-button[disabled] {
```
alexlyp: Not sure where the issue is arising from but the buttons' text looks to be larger or possibly bolded compared to existing

alexlyp:
Obviously, needs a quick conflict resolution.

alexlyp:
But confirmed it to be working as expected.  Mostly want to get @go1dfish's okay on the redux-form impl and a few design tweaks and it should be good enough to merge. We can have @linnutee or @sander to do a quick clean up prior to release.

app/style/MiscComponents.less
oktapodia: I did that because all the buttons in decrediton are not a real buttons like `<button></button>` or `<input type="submit" />`, they are a simple `div` with an `onClick` event.

Because for my form I'm using the `onSubmit` event, with a reduxform, I have to use a real button, and then I can apply the disable style with a `disabled` key inside the input. Easier than toggling the className I think

go1dfish: I think we should switch to buttons as much as possible.

I am a favor of semantic markup, at least at a basic level.

package.json
go1dfish: https://youtu.be/vYEXzx-TINc?t=10

alexlyp:
I'll open an issue concerning styling of the page and the button streamlining as discussed.  Great work!

oktapodia:
fixes #331

PR #746, opened 2017-10-04
Translating decrediton

### LINKED ISSUE #572: Add internationalization support
Decrediton is lacking support for internationalized messages; currently everything is hardcoded in English. Either we should derive the user's language from the OS through Electron (if that's possible and the translations are available) and display that automatically, or provide an option on the Settings page which displays all available translations and lets the user dynamically switch.

We can most likely use this webpack plugin: https://github.com/webpack-contrib/i18n-webpack-plugin

matheusd:
Fixes #572.

Opening PR for discussion. Should be close to final. After this is approved and merged I'll work with @peterzen and the rest of the translation managers to get translation started.

There's a readme inside the `/app/i18n` dir for general instructions and templates for adding translated items.


matheusd:
I commented the tests because we'll need to rework how to unit test with translation.

alexlyp:
Looks great to me, but I want to get everyone's input on this before we proceed with this i18n impl.

peterzen:
Looks good to me as well.

matheusd:
Free to merge when possible.

PR #752, opened 2017-10-06
writeCfgs only if there are no configs located in the decrediton appdata dir

### LINKED ISSUE #749: Use default dcrd.conf
Currently decrediton is starting dcrd with a dcrd.conf that it writes.  This does not give users any opportunity to make changes to their dcrd settings that aren't exposed by decrediton (such as the datadir to change the location of the blockchain data).

Paymetheus handled this nicely and I think decrediton will be able to do the same.  In Paymetheus, the launcher application would still start dcrd with nonstandard options, but it would do so using the command line arguments instead of using a config file.  This allowed users to modify the config file, and any flags passed by the launcher would override the user's settings if they conflicted.  Otherwise the user's settings were used.

alexlyp:
Fixes #749

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -304,21 +304,11 @@ export function dcrctlCfg() {
 export function writeCfgs(dcrd, dcrwallet, dcrctl) {
   var cfg = getCfg();
   if (dcrd) {
-    var net = 0;
-    var autobuy = 0;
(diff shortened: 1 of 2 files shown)


jrick:
I'd remove the testnet option from the written config and rely entirely an argument for that.

Issue #756, opened 2017-10-06
Add button to show BalanceOverviewInfoModal.js once again


alexlyp:
It was created to explain the various fields user's accounts' balances.  It should have roughly the same look and feel as the information button on the Tickets page for PurchaseTicketsInfo.js reveal

On the accounts pages just below the Header, underneath Add New button, upper right corner.

alexlyp:
Closed

PR #759, opened 2017-10-07
Feature/add balance overview info modal

oktapodia:
Fixes the issue #756


alexlyp:
Looks good, just noticed that the `Close` button on the modal needs to float right, otherwise it's gtg

oktapodia:
@alexlyp Updated

Issue #594, opened 2017-08-17
Clean up ticket buyer configuration fields


alexlyp:
- Make sure each field is checked for errors before entry and fix ways in which it currently works.
- Add text to make it clear what type of entry is expected (atoms, DCR, % etc)
- Add back in instructions for fields to explain what each does.

go1dfish:
We want to add a generic DCR Input component for this.

Like how Balance works but for DCR inputs.

have it take 2 extra params, 1 for “required”

and 1 for “showErrors”

There are a lot of cases in the UI where we don’t want to show validation errors until some attempt is made

i.e. required fields shouldn’t show as an error until the person tried to submit a form, so the field could always validate and only show errors when showErrors is true

and the form would set showErrors true when an attempt had been made


It may make sense to make a similar component for non-dcr numeric inputs as well

vctt94:
".Make sure each field is checked for errors before entry and fix ways in which it currently works."

Right now I am checking for errors on onChange and if it found any error, I change the state of the app/Components/PurchaseTickets/index.js. Would that work for making sure the field is checked? 

And By adding instructions for fields, you mean like a tooltip? 

Is this FeeInput I created for showing fees a good idea? Or should I let just the Decred and Numeric Inputs? 

Thanks

PR #763, opened 2017-10-09
Clean ticket buyer fields

vctt94:
This is a pull request for solving the issue https://github.com/decred/decrediton/issues/594

Also this pull request is refactoring the inputs for creating a general DecredInput and NumericInput


app/components/DecredInput/DcrInput.js
```diff
@@ -0,0 +1,40 @@
...
+        requiredMessage= {requiredMessage}
+        disabled={disabled}
+        readOnly={readOnly}
+        placeholder={placeholder}
+        required= {required}
+        invalid= {invalid}
+        value={value}
+        onChange={onChange}
+      />
+      <Balance
```
go1dfish: I don’t think we need to show the balance again but we should show the unit type and eventually a drop down for atoms vs DCR or other future units.

app/components/DecredInput/index.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import DcrInput from "./DcrInput";
+import { autobind } from "core-decorators";
+import balanceConnector from "../../connectors/balance";
+
+@autobind
+class DecredInput extends React.Component {
+  constructor(props) {
+    super(props);
```
go1dfish: No need for this constructor if you are not going to do anything with it

app/components/DecredInput/DcrInput.js
vctt94: What drop down would be that?

alexlyp:
default Ticket fee, Tx fee overflow the inputs in the expanded ticket purchase area

alexlyp:
when an error is given on ticket fee and tx fee i see none, but it seems to push the Expiry field to the next line

alexlyp:
AutoBuyer information button needs padding to match the Ticket Purchase information button above

alexlyp:
Empty fields in Auto buyer area need to throw errors as well.  Currently I'm able to turn on the autobuyer with empty fields

alexlyp:
You should be able to have a balance to maintain higher than your balance

alexlyp:
Ticket Fee and Tx Fee should also have sensible max values (0.1 DCR is probably good there)

alexlyp:
Max Fee field needs to respond to DCR --> atoms change as well.

alexlyp:
Errors for fields in the Auto Buyer config should be below the label and the field.  Currently they are just under the field and cause bad shifting when shown

alexlyp:
Ok looks good except for one issue.  When you first go to the purchase tickets form and enter an invalid value into ticket fee or tx fee, you don't get any errors displayed until AFTER clicking purchase.  Once purchase is clicked it works as expected.  Fix that this PR should be good to go.

app/components/views/TicketsPage/Page.js
```diff
@@ -0,0 +1,41 @@
+import StakeyBounce from "../../StakeyBounce";
```
alexlyp: This entire Page.js can be removed.

Issue #755, opened 2017-10-06
Tooltips don't show after adding stakepool api


alexlyp:
After successfully adding a stakepool api key no tooltips show up until you click away to another Page then back to the Tickets page.

Thoughtscript:
@alexlyp - now have an answer for this that I'll be checking in shortly. Interestingly, the TicketsPage/Index.js component doesn't go through the normal componentDidMount() method the first time someone passes in an API Key - rebuilding the tooltip in componentDidUpdate() will solve the problem - finishing this up since I had a chance to look at it a little last night and this morning. PR coming soon. Thanks!

alexlyp:
closed

PR #768, opened 2017-10-10
tooltip rebuild needed in componentDidUpdate()

Thoughtscript:
Issue #755 - componentDidMount() is not called when this component is first viewed by a user. That's where ReactToolTip is usually initialized and built. Here we explicitly call it in componentDidUpdate() and place a check to prevent unnecessary rebuilds since this is a high-volume method (one that is called frequently).


alexlyp:
this needs `npm run lint-fix`

Issue #757, opened 2017-10-06
Time since last block 


alexlyp:
After the recent changes with the i18n implementation, I've noticed some strange counting/ changing on page turn of the time since string at the bottom of the sidebar.

I don't think it was implemented correctly previously so no need to revert.

I'd like to open up discussion about what should be the proper functionality of the "time since block" string here.

We have the new block's timestamp to compare against the current time.  Ideally this would be a component that calculates the number of seconds since the last block then uses a time parser to make it into readable text.  I was never able to figure out a method that could constantly update the component and not have a huge impact on performance

matheusd:
Ok, let me run through the implementation details of that one.

Before the i18n conversion, that label was recalculated every 10 seconds from a dispatched call on the component. That call bubbled all the way up to an action that extracted the last block's timestamp and calculated the relative time (in minutes) until now.

But react-intl has a `<FormattedRelative>` component which is able to automatically translate **and** update relative time strings ("now", "10 seconds ago", "45 seconds ago", "1 minute ago", etc). You just pass it a `Date` object as value (say, date/time of the last block's timestamp) and it will set up triggers and auto update at the "interesting" times at a maximum rate defined by a prop (currently 30 seconds for the block time component in the sidebar).

The problem of it updating when changing pages is (AFAIK) because the sidebar gets umounted/remounted since it is defined on each page. What you see after a page turn is actually the initial value of the component after a remount.

We should recheck this issue after the refactor to a single page container, as that will prevent the sidebar getting remounted all the time and should automatically solve it. If, after that refactor, there are still issues, then we can take another look.

alexlyp:
This was fixed in #772 and i believe it is acting as expected now

Issue #732, opened 2017-10-02
Update pagination UX


alexlyp:
Update pagination look and feel to match designer's suggestions.

linnutee:
See [UI components](https://xd.adobe.com/view/82325302-2eb8-4684-9b46-7ef1bc6b227a/ ) and specs: [UI components specs 031017.pdf](https://github.com/decred/decrediton/files/1356155/UI.components.specs.031017.pdf)

Context examples are in "history" as well in the new "my tickets" views:
[decrediton ui v012 - transactions and history - history](https://xd.adobe.com/view/5ca79364-9459-4017-b0b1-7e26f90059fd/)
[decrediton ui v01 - tickets - my tickets](https://xd.adobe.com/view/7707b586-62f0-4c93-afaf-1c5b9432b8eb/screen/5514ce96-bfb4-4eae-b37f-2c87e0d45d1b/My-Tickets)

<img width="717" alt="screen shot 2017-10-04 at 17 36 17" src="https://user-images.githubusercontent.com/17774057/31181665-911c90c8-a92a-11e7-9213-da8c97a5fbf9.png">

-------------------------------------------------
Past releases and references: https://www.notion.so/eeter/Decrediton-UI-knowledge-base-5cc97bc87cc74e80bb4ef42e996fab2d

matheusd:
Preview.
![paginator mp4](https://user-images.githubusercontent.com/1752493/31445457-76df4e18-ae74-11e7-8309-3a452b649a8e.gif)

linnutee:
Interactions look right. Would only recommend adding another hover state for the arrows. Instead of showing the blue backdrop, have the arrows change color. 

E.g. 
```
default: #8997A5 
hover: #0C1E3E 

```

I can't find a good reason why to use a different arrow style for these. Sorry about including those in the first place. Don't mind if they are left as they are, but may be cleaner to just have the same full arrow as used elsewhere.

<img width="772" alt="screen shot 2017-10-12 at 03 06 59" src="https://user-images.githubusercontent.com/17774057/31472964-6de88032-aefa-11e7-885a-58583f160696.png">

alexlyp:
Added in #772

Issue #731, opened 2017-10-02
Dropdowns for account selection restyling


alexlyp:
Update css to match designer's suggested style.

matheusd:
See also PR #728 and issue #663.

alexlyp:
Sorry I should have been more explicit.  The @linnutee has come up with some new designs for various current assets.  This is a placeholder issue for that work.  I'll try and link with images to be more clear.

linnutee:
<img width="873" alt="screen shot 2017-10-04 at 18 00 56" src="https://user-images.githubusercontent.com/17774057/31182986-fe2624ba-a92d-11e7-9741-dcedc8c60172.png">

[UI components](https://xd.adobe.com/view/82325302-2eb8-4684-9b46-7ef1bc6b227a/ ) and specs: [UI components specs 031017.pdf](https://github.com/decred/decrediton/files/1356155/UI.components.specs.031017.pdf)

Ok i've gathered the different fields and dropdowns together and fine tuned them a little bit. 
These are based on the initial release assets, the updates are primarily in small details of how each states interactions works – what gets highlighted etc.

Summary:
- Default states should either include a description of content for normal fields or
 type of amount (if it's DCR, DCR/Kb, Blocks, Tickets…). 
- On hover there's a color change to slightly darker 
- On activation/click the stroke changes to blue in most cases and in dropdowns the dropdown field includes a blue stroke in the bottom of it. Text color remains the same `#0c1e3e` as on hover.
- After entering the data, the stroke color changes back to "default" and text color remains the same `#0c1e3e`.
- Inactive/disabled = greyed out. 

Here is the previous version from February 2017, just for a reference.
[decrediton - v004 - 0217.zip](https://github.com/decred/decrediton/files/1356274/decrediton.-.v004.-.0217.zip)

-------------------------------------------------
Past releases and references: https://www.notion.so/eeter/Decrediton-UI-knowledge-base-5cc97bc87cc74e80bb4ef42e996fab2d

matheusd:
Preview for new select and accounts select styles. 

@linnutee, could you please double check if the colors look right? I'm having a bit of trouble following some of the specs (eg: on the latest answer you mentioned the text color as #0c1e3e but on the specs, text color is specified as #091440 in the input, #dadfe2 in the select items; stroke on the select items is listed as #596d81 but that seems a bit dark on my end, etc).

[...]

Issue #734, opened 2017-10-02
Update send all styling


alexlyp:
Update to use an icon and also refine the look and feel for when it is in "send all" mode.

matheusd:
@linnutee do you have an icon for the send all button?

linnutee:
Not yet, taking this and a few other issues on tomorrow + thu.

linnutee:
Preview:https://xd.adobe.com/view/5ca79364-9459-4017-b0b1-7e26f90059fd/screen/02b7aeb3-2ba4-47e0-9ad4-39edae81af1a/Send-

Icon 20px
[send-all.svg.zip](https://github.com/decred/decrediton/files/1380854/send-all.svg.zip)

alexlyp:
Fixes in #772

PR #772, opened 2017-10-11
UI Improvements

### LINKED ISSUE #730: Dropdowns for type selection restyling
Update all dropdowns for various type selection to match designers prototype.

matheusd:
Fixes #730, #731, #732


alexlyp:
Overall looks great.

Thoughts on what to do with the Tx Type label next to the dropdown?  I'm thinking we should just nuke it since I feel like the dropdown is more more clear on purpose currently.

alexlyp:
Also I find the vertical padding for the transaction type options is a bit high, should a bit more wider and not as tall per option imo.

matheusd:
Ok, I've made the improvements listed here and by @linnutee on the individual issues.

There are likely to be further issues after the designers get their hands on a built version and can play around, but this should be close to done.

matheusd:
Latest push fixes the icon as specified in #734

matheusd:
I've also fixed #757 now on my branch for this PR, but github is not showing the new commits. May be due to the fact that I had to rebase the work.

If this doesn't show up correctly, I'll close and open a new PR.

alexlyp:
I'll be opening another issue that i noticed while testing this.  but this is good to go.

PR #778, opened 2017-10-13
React16

Jascha-Sundaresan:
Changes:

bumps both react and react-dom to 16.

removes react-addons-test-utils as that's part of react-dom now

removes react-tap-event-plugin because we're not supporting iOS devices

bumps enzyme to v3 (many breaking changes to enzyme from v2 to v3 but since there are literally zero tests, this doesn't matter)

bumps material-ui from 0.17.4 to 0.19.4. We will eventually remove material-ui altogether

switches from babel-preset-es2015 to babel-preset-env. This is the standard now.

moves two eslint deps to devDeps

also, bundled with yarn and lockfile committed to repo.


alexlyp:
please remove the yarn.lock from this PR.  I am going to do that myself in a seperate PR.

Jascha-Sundaresan:
done

PR #779, opened 2017-10-13
Fix imports

Jascha-Sundaresan:
this adds the modules field to the base webpack config which allows us to refer to the subdirectories within /app the same way we refer to them in /node_modules

an example given using TicketsPage/index.js

also, an index was created in connections for importing convenience.


Jascha-Sundaresan:
This also works for stylesheets
`@import '../../style/sheet'` becomes `@import ~'style/sheet'`

alexlyp:
Good stuff, thanks!

PR #780, opened 2017-10-13
Freactal

Jascha-Sundaresan:
this is the smallest freactal demo I can easily show.  I took one component, StakeInfo, and refactored it.

It was previously two components, a stateful container component and a stateless functional presentational component.

It is now still two components, a stateful container component, and a stateless functional presentational component, but with less boilerplate. (23 LOC)

the stateful component is created by provideState which returns a HOC to wrap the stateless one with.


alexlyp:
I would like to discuss the possibility of switching to use Freactal.  Please open an issue where we can debate the pros/cons of such a system.

PR #774, opened 2017-10-11
New tabbed page & snackbar components

matheusd:
Fix #733, work towards fixing #632.

UI is probably not yet 100%, but I'm opening up the PR to get feedback on possible improvements.

The new global snackbar is currently being used only on the new tabbed transactions page. Depending on comments, I can either change the snackbar on all other pages in this PR or incrementally change as we refactor the pages to the new header style.

PR #781, opened 2017-10-13
Remove react-router-redux

Jascha-Sundaresan:
@go1dfish confirmed that this would be OK to do since we're not using it and it's preventing us from upgrading to react-router v4

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,45 +65,30 @@ decrediton --extrawalletargs='-d=debug'
 ## Developing
 
 Due to potential compatibility issues, for now, all work should be
-done with node v6.9.5 and electron 1.4.15.  The recommended way to install
-node is using nvm.
+done with electron 1.4.15.
 
-This has primarily been tested on Linux at the moment although OSX
-should work similarly.
-
-``` bash
-git clone https://github.com/creationix/nvm.git ~/.nvm && cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`
-. ~/.nvm/nvm.sh
-nvm install v6.9.5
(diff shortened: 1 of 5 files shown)


go1dfish:
Correct, react-router-redux is only necessary if we want to access or manipulate routing state in the redux state.

We're not doing that, and have no plans to do that in the short term, and react-router v4 is very nice.

Jascha-Sundaresan:
I'm almost done w/ the react-router-4 upgrade.

There were a few instances of this.context.router being used and I switched to the <Redirect to="/path" /> pattern that's encouraged.

There are still a few bugs to iron out with the NavLinks in the sidebar and the txDetails

go1dfish:
Turns out this is going to conflict with some new code here that does take advantage of react-router-redux


https://github.com/decred/decrediton/pull/774/files

So that will need to be addressed somehow to make this work.

Jascha-Sundaresan:
the common workaround is to use react-router-redux@5.0.0-alpha

matheusd:
On my PR I updated the react-router-redux, but I don't think the transition component actually needs it. The react-router-transtions may be incompatible with react-router v4, so depending on which PRs get merged in which order they'll need appropriate updating

dajohi:
@Jascha-Sundaresan please rebase.

Jascha-Sundaresan:
This work should be shelved until we tackle the react-router-4 upgrade.

Feel free to close this and I'll create a new PR that rolls this work into the upgrade.

alexlyp:
Closing until react-router-4 upgrade is discuss for this.

PR #783, opened 2017-10-13
Yarn updates

alexlyp:


Jascha-Sundaresan:
Why isn't there also an app/yarn.lock?

alexlyp:
There is, but just no updates to it in this PR

Issue #567, opened 2017-08-08
Consolidate uses of <ReactToolTip />


sndurkin:
`<ReactToolTip />` really only needs to be used once within Decrediton, and all elements that have tooltips can share the instance (except for specific instances where an element may want to display tooltips separately from normal ones).

We should add `<ReactToolTip />` to `<App>` (or whatever highest level component makes sense), then use `data-tip` and the other attributes to customize each tooltip on a per-element basis.

Thoughtscript:
This is related to both #570 and potentially #755. The method used to resolve #570 can maybe generalized for all instances?

Thoughtscript:
Current uses of ReactToolTip (as a direct **import**):

1. **components/views/SecurityPage/Page.js** - This is actually an unused ReactToolTip that can be removed completely. I think the intent was to have this here since it's a higher-order component for both **views/SecurityPage/Header.js** as well as **Header/index.js**.

2. **components/views/AccountsPage/AccountRow/Row.js** - "Accounts" from main selection sidebar. Used on Accounts that are not default or imported. This needs getContent...
 
3. **components/views/AccountsPage/AccountRow/index.js** - "Accounts" from main selection sidebar. Used on Accounts that are not default or imported.

4. **components/views/ReceivePage/Page.js** - "Receive" from main selection sidebar. Used in account list dropdown.

5. **components/views/TicketsPage/Page.js** - "Tickets" from main selection sidebar. Used after API Key is loaded in.

6. **components/views/SendPage/Page.js** - "Send" from main selection sidebar. Used only on the wallet/account-tip icon/button.
 
7. **components/views/HomePage/Page.js** - "Overview" from main selection sidebar. Used only on the "Rescan Blockchain" button.

8. **components/views/GetStartedPage/DaemonLoading/Form.j**s - Used when you first start up the app and it's loading the Decred chain. The tip is displayed on the skip sync button that provides an option to skip the sync as a new user.

9. **components/CopyToClipboardButton.js** - Used to show a tooltip on a copy button after something has been copied. The two places where this is used are the **components/views/ReceivePage/Page.js** and **components/SignMessage/index.js**.

Thoughtscript:
Current uses of ReactToolTip (as a direct **import**):

1. **components/views/SecurityPage/Page.js** - This is actually an unused ReactToolTip that can be removed completely. I think the intent was to have this here since it's a higher-order component for both **views/SecurityPage/Header.js** as well as **Header/index.js**.

[...]

PR #786, opened 2017-10-14
Icons

Jascha-Sundaresan:
This refactors the TicketsCogs component.

I've created a general purpose Box, Flex, Text, and Heading put them in a shared components directory that's directly accessible via `import from "shared";`

I also created an Icon component, and ported a handful of icons so far.


Jascha-Sundaresan:
Todo for next style PR:

Deduplicate layouts for SignMessage and VerifyMessage

Thoughtscript:
Recommend using a custom tooltip library. Issue #567 has been consolidated in PR: https://github.com/decred/decrediton/pull/788 which is great from a "if we keep using ReactToolTip, here's  a strictly better use of it" but, we can simplify cases in which we use tooltips (no more state rebuild() commands, etc.) - it will mean that we'll have to refactor all existing tooltip elements (not just the actual ReactToolTip DOM instances but the elements that specify through "data-tip" what is to be displayed). Not a worry - I'd actually enjoy doing that since I'm looking into consolidating tooltips already. 

@Jascha-Sundaresan approach looks like a great way forward though I haven't had a chance to test it much.

app/components/PurchaseTickets/Form.js
```diff
@@ -100,8 +99,8 @@ const PurchaseTicketsForm = ({
       <div className="stakepool-voting-title-area-name">
         <T id="purchaseTickets.title" m="Purchase Titckets" /></div>
       <div className="stakepool-purchase-ticket-input-buttons">
-        <PurchaseTicketsInfoButton onClick={onShowTicketsInfo}/>
-        <TicketsCogs opened={!isShowingAdvanced} onClick={onToggleShowAdvanced}/>
+        <Icon i="info" onClick={onShowTicketsInfo}/>
+        <Icon i="cog" pt={ 5 } pl={ 10 } active={isShowingAdvanced} onClick={onToggleShowAdvanced}/>
```
Jascha-Sundaresan: pt and pl stand for "padding-top" and "padding-left"

padding and margin props accept numbers, strings, and arrays.

numbers 0-1 gets converted to percentage strings `1/2` to `"50%"`
integers < theme.space are theme.space[number]
larger numbers are converted to px strings

Strings are literal.

Arrays provide values for responsiveness set by breakpoints at theme.breakpoints.

go1dfish:
I tend to be instinctively repulsed by CSS in JS solutions as well, but I'm open to the idea of using them for page layouts and as a means of making flexbox easier to work with.

Jascha-Sundaresan:
@sndurkin as a react developer, this makes it FAR easier for me to maintain styles. This is nothing like radium or how it was implemented.  We are NOT using inline styles here, nor are we manually stying every layout div tag w/ a style object (or under the hood).

styles are completely deduplicated here and what I'm doing is building out base components that accept all the necessary props for being laid out according to design's specs.  

For a react developer this is far easier to maintain. I know how things are going to affect each other w/o having to grep the repo.  Facebook is the hugest platform when it comes to styling and the push for css-in-js came from them.  Admittedly the field had some growing pains, most notably with the phase of it that you were initially exposed to, but this area has matured and proper practises have emerged.

`Base` `Box`, `Flex`, and `Text` represent how to build components in ways that have styling abstracted away in sensible layers, independent from your markup itself.  We can do `<Header is="h1" color="altBlue2">Text here</Header>`

This keeps us entirely in the "react" way of thinking while being able to handle styling and markup semantics independently, as properties of the react component.

matheusd:
TLDR: I think there are possible advantages to using cxs, but I think there's too much stuff going on at once   here and if we choose to go the cxs route that should be done like the inline style removal or the translation work was done: change every component to the new way.

Biggest advantage I see with cxs is having scoped classes and possibly reducing some of the css class names nonsense (eg: _receive-content-nest-copy-to-clipboard-icon_). *But* I see that as using a modular style of the first example of the official package (https://github.com/jxnblk/cxs): 

```javascript
// on style/sidebar.js
import cxs from "cxs"
export const menuLogo = cxs({ position: "relative", height: "60px", ... })
  
// on bar.js
import { menuLogo } from "style/sidebar"
<div class={menuLogo}>...</div>
```

Now, this is one issue independent of everything else we can decide separately: change all styling from css classnames to cxs classes. I'd probably vote yes on this issue, with the caveat that I think we should **not** do this exactly at this moment: I'd do this after refactoring more stuff into individual components (ie.: after moving all headers out of individual pages, after consolidating the uses of the password modal, after unifying all uses of the input component, etc).

A second issue is refactoring from <divs> into other types of components (either using cxs or not). While I appreciate being more specific than using "div", I **don't** like this: https://github.com/decred/decrediton/pull/786/files#diff-052c44a3d62c256534bc81ee363feae4R23

```javascript
 <Box w="20em" ml="auto" f={ 16 }>
```

This seems way more cryptic at first glance than
```javascript
import { selectArea } from "style/stakepool";
//...
<div class={selectArea}>
```

Or maybe this:
```javascript
import { selectAreaStyles } from "style/stakepool";
const selectArea = cxs('div', selectAreaStyles)
//...
<SelectArea>
```

Between these three styles, at this moment I'd prefer number 2.

Jascha-Sundaresan:
thanks for chiming in @matheusd.

The direct classname API that cxs exposes is useful as a framework agnostic application but the export from `cxs/component` is significantly more powerful because it accepts functions in addition to objects, allowing us to do things like:

`const mayClick = ({ onClick }) => onClick ? clickableStyle : null;`

as to the use of Box that you pointed out, that's only because I hadn't refactored the component its containing to accept those props.

```
          <Box w="20em" ml="auto" f={ 16 }>
            <SelectStakePool
              options={unconfiguredStakePools}
              value={selectedUnconfigured}
              onChange={onChangeSelectedUnconfigured}
            />
          </Box>
```

will eventually just be:
```
          <SelectStakePool w="20em" ml="auto" f={ 16 }>
              options={unconfiguredStakePools}
              value={selectedUnconfigured}
              onChange={onChangeSelectedUnconfigured}/>
```

What these levels of abstraction allow is for us to encapsulate styles where they're most appropriate.

Within the `SelectStakePool` component, it only cares about how it looks.

when the Component is consumed, you place it as needed using its layout properties.

Passing in a cxs class as a className would not allow for this.

Also, the properties are really powerful:

`w` accepts a number, string or array.  Numbers between 0 and 1 get converted to a percentage, strings are accepted literally, numbers are converted to px values, and arrays are responsive values on breakpoints. The same applies for the other props.

Jascha-Sundaresan:
also, 
```
import { selectAreaStyles } from "style/stakepool";
const selectArea = cxs('div', selectAreaStyles)
//...
<SelectArea>
```
The correct syntax is `cxs("div")(selectAreaStyles)

 is literally what I'm doing right now, and what I encourage. The power of this is what I already stated. `selectAreaStyles` can be either an object or a function that accepts props and returns an object. 

So far, I'm simply not abstracting things out to a separate directory until/unless we come to a point where some style rule is actually consumed by two different things. It just doesn't happen very often when everything is a component w/ properties and you can just reuse the components.

dajohi:
please rebase @Jascha-Sundaresan

Jascha-Sundaresan:
will rebase this afternoon. one of the conflicts is best addressed by adding tooltip functionality to my `Box` component.

alexlyp:
Due to the epic amount of conflicts that have arisen from this and the age of the PR I'm closing this.  If you'd like to open up discussion about changing this please open an issue with your proposed refactor.

PR #788, opened 2017-10-14
Consolidate tooltip

Thoughtscript:
Issue #567 - All pre-existing, changes, and remaining DOM tooltip instances documented.

### DIFF
--- a/app/components/CopyToClipboardButton.js
+++ b/app/components/CopyToClipboardButton.js
@@ -41,6 +41,7 @@ class CopyToClipboardButton extends Component {
         data-place="bottom"
         data-type="info"
         data-effect="solid"
+        data-for="appToolTip"
         data-tip={this.state.showTooltip ? formatMessage(messages.copied) : ""}
         onClick={this.onClick}
         onMouseLeave={this.onMouseLeave} />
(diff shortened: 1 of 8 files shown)


Jascha-Sundaresan:
I created a tooltip component in #786 that obviates the need for this.

As soon as that gets merged in you can use it to remove react-tooltip altogether.

Thoughtscript:
Per our discussion @Jascha-Sundaresan and for the obvious reason that it makes sense to first bring all of our React Tool Tip instances up to optimal use before we switch over, refactor, and rewrite all tooltip instances: we should merge this first.

Then we also get: (1) a stable and consolidated ReactToolTip commit, (2) make it slightly more efficient to switch to the new tooltip, (3) have a place to roll-back to (not saying that your tooltip won't work - I've seen it and we've been chatting about it - looks good).

But yes, longer-term, the approach that I'm behind is to build a better tooltip that manages the actual tip via CSS (no more state rebuilding thank you), is componentized and reusable in a more logical way - ReactToolTip requires BOTH an independent DOM element as well as data-for attributes all, and is primarily React-based from a development though not a runtime side.

alexlyp:
Due to the moving/combining of the Receive/Send this will require conflict resolution

Thoughtscript:
Great I'll take a look and make the required merges.

alexlyp:
Well I'd hold off on any work here.  @Jascha-Sundaresan is currently refactoring the tooltip system so this PR most likely will end up being obsolete.  I'll keep it open until it's clear that his work will supercede this.

Thoughtscript:
Just saw the comment on the main issue.

alexlyp:
Closing in favor for @Jascha-Sundaresan recent PR

Issue #771, opened 2017-10-10
Dcrd/wallet selection on startup


alexlyp:
We should add functionality to accomplish the folllowing items:
- Use an already running instance of dcrd (and manually enter rpcuser/pass/cert location)
- Select a different file location for wallet.db usage
- Further steps to ensure that dcrd/dcrwallet binaries are shutdown (if they are child-processes)

alexlyp:
dcrd advanced is working.

PR #793, opened 2017-10-15
Dcrd selection on startup

vctt94:
Pull request for solving issue https://github.com/decred/decrediton/issues/771


matheusd:
Haven't run the code, but eyeballing it: You are asking to connect to the running dcrd at every startup, correct?

I think it would be better to ask whether to use the built in daemons based on a config setting, which would be unset on install and only ask for a choice on the first run, possibly before running the daemons and with a clear indication that the default behavior for inexperienced users should be to just continue on to the next step.

Something to the effect of "Customize startup options yes/no? If you are unsure or don't know what this means click no".

Another option would be to have a config select on the settings page "customize startup options" and only show the login form if that variable is set.

vctt94:
Yes, right now I ask every start to keep running with daemon running or login in a different dcrd. I agree with you, I think I will create a config argument for advanced dcrd setup. Any suggestions are welcome though.

vctt94:
RIght now I am getting the rpcuser, rpcpass and rpccert from the form and starting a different daemon from the config file. I am starting dcrd with the command: 

dcrd -u test -P test --rpccert=/home/victor/.dcrd/rpc.cert --testnet --rpclisten=127.0.0.1:19119

and the command for checking the daemon I am running is the following:
dcrctl getblockcount -u test -P test --rpccert=/home/victor/.dcrd/rpc.cert --testnet --rpcserver=127.0.0.1:19119 --walletrpcserver=127.0.0.1:19110

But always I am getting an unauthorized message: 

"[WRN] RPCS: RPC authentication failure from 127.0.0.1:58166
error: 401 Unauthorized." 

If I run these commands on command line on my terminal, it works just fine, but when I am using the spawn on the launchDcrd method, it gives me the unauthorized message.

What can I be doing wrong?

alexlyp:
So after successfully starting up dcrd and dcrctl receiving proper creds/appdata/cert, it just sits at the Login to RPC view with the DecredLoading spinner.

alexlyp:
Also there needs to be an option to connect to an already running dcrd, kind of like skip_daemon_loading option.  So you should be able to enter in any remote dcrd information and it'll go out and communicate with that one and won't even try to start it's own.  Maybe use a check box or something?  Default is unclicked, and click would be to "connect to already running dcrd"

alexlyp:
And I'd like some input from @go1dfish and @matheusd about UX for this feature.   Typically most users shouldn't have to see this screen.  Thoughts on adding a config option to let users see this screen otherwise it'll start up as it does now?

app/components/views/GetStartedPage/Page.js
```diff
@@ -3,14 +3,20 @@ import DecredLoading from "../../DecredLoading";
...
+    <div className="page-view inverted-colors">
+      <Header {...props} />
+      <div className="page-content-fixed">
+        {
+          isAdvancedDaemon ?
+            isSubmited && !hasErrors ? <DecredLoading hidden={!props.isProcessing} className="get-started-loading" /> : null
+            : <DecredLoading hidden={!props.isProcessing} className="get-started-loading" />
+        }
+        <Body {...props} />
+      </div>
```
alexlyp: You shouldn't be making any changes here.  If you need it to take those props into account add a new selector and then add that to the isStartupProcessing selector (refer to isPrepared) for an example of this.  But this Page.js should be unchanged for your needs.

app/actions/DaemonActions.js
```diff
@@ -1,49 +1,89 @@
...
+  const {rpcuser, rpcpassword, rpccert } = args;
+  const rpchost = RPCDaemonHost();
+  const credentials = {
+    rpcuser: rpcuser,
+    rpcpassword: rpcpassword,
+    rpccert: rpccert
+  };
+  daemon.startDaemonAdvanced(args)
+  .then( () => {
+    dispatch(syncDaemon(credentials, rpchost));
```
alexlyp: I'd prefer you store these credentials in state for use else where instead of continually passing them further and further down.

alexlyp: In the end, seems like you are passing these all the way down just to have them accessible for the startRpc request.

alexlyp:
After you accomplish the 2 scenarios we discussed on slack you can make 2 forms on the same page.  We'll open another issue to add proper styling to the forms and startup page.

app/main.development.js
```diff
@@ -337,13 +417,22 @@ ipcMain.on("grpc-versions-determined", (event, versions) => {
...
 
-const launchDCRD = () => {
+const launchDCRD = (credentials) => {
   var spawn = require("child_process").spawn;
-  var args = ["--configfile="+dcrdCfg()];
+  let args = [];
+
+  if(credentials){
+    const {rpcappdata} = credentials;
+    const rpccert = `--rpccert=${rpcappdata}/rpc.cert`;
```
alexlyp: you can remove this, setting appdata then makes dcrd use the .cert/.key found in the specified directory.

app/main.development.js
```diff
@@ -302,14 +362,34 @@ ipcMain.on("start-wallet", (event, arg) => {
...
+    host = RPCDaemonHost();
+    port = RPCDaemonPort();
+  } else {
+    host = RPCDaemonHost();
+    port = RPCDaemonPort();
+    args.push(["--configfile=" + dcrctlCfg()]);
+  }
+
   var spawn = require("child_process").spawn;
-  var args = ["--configfile="+dcrctlCfg(), "getblockcount"];
```
alexlyp: you need to re-add this for startType == 2

alexlyp: Specifically: 

`args.push(["--configfile="+dcrctlCfg()]);`

Issue #691, opened 2017-09-21
Use insight to determine current top block


alexlyp:
https://mainnet.decred.org/api/status  will return current info  and we can just use "info.block"  to see what the top block is.

matheusd:
Is this to replace the neededBlocks estimation? Or somewhere else?

alexlyp:
Yup that's it exactly.

PR #800, opened 2017-10-17
Run executables from /bin instead of /app/bin on development

### LINKED ISSUE #770: Allow for dev to use the same bin/ directory
Instead of having to use 2 different directories for binaries for production and dev (`bin/` and `app/bin/` respectively).  We should be able to consolidate into using the same `bin/` for both.

matheusd:
Fix #770

### DIFF
--- a/app/components/views/GetStartedPage/DaemonLoading/index.js
+++ b/app/components/views/GetStartedPage/DaemonLoading/index.js
@@ -1,29 +1,40 @@
 import React, { Component } from "react";
 import { autobind } from "core-decorators";
+import ReactTimeout from "react-timeout";
 import {
   DaemonLoadingFormHeader as DaemonLoadingHeader,
   DaemonLoadingFormBody
 } from "./Form";
 
 @autobind
-class DaemonLoadingBody extends Component {
+class DaemonLoadingBodyBase extends Component {
   constructor(props)  {
     super(props);
     this.state = this.getInitialState();
   }
 
-  componentWillUnmount() {
-    this.mounted = false;
-    this.resetState();
-    if(this.timeoutId) {
-      clearTimeout(this.timeoutId);
-    }
-  }
-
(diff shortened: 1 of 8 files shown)


matheusd:
Latest push also fixes #691

alexlyp:
Strange on sync/blockchain DL it displayed what appeared to be the correct required block height but then when it got to what appeared to be the estimated height it moved on (156K, 149K respectively)

matheusd:
Should be all right now. Bug was a misplaced closure.

app/components/views/GetStartedPage/DaemonLoading/index.js
```diff
@@ -1,38 +1,49 @@
...
+    this.mounted = true;
+    this.timeoutId = this.props.setTimeout(() => {
+      if (this.mounted) {
+        this.setState({ showLongWaitMessage: true });
+      }
+    }, 2000);
+    const neededBlocksInterval = this.props.network === "mainnet"
+      ? 5 * 60 * 1000
+      : 2 * 60 * 1000;
+    this.props.setInterval(this.props.determineNeededBlocks, neededBlocksInterval);
```
alexlyp: Can you explain why you set and interval here based on network target block time?  Not sure it's necessary.  Really just need it once on startup

matheusd: During initial block download (on a first time run with a clean block database) the initial sync may take minutes/hours. During that time the block height will change, so until the daemon is completely synced to the network it should try and find the needed block height.

I understand that the difference may be small (I added the interval check mainly for completeness), so if you feel the block difference is small we can take it off.

alexlyp: oh very nice!  didn't think about that situation.  nope that's totally understandable now.  nice work!

PR #810, opened 2017-10-19
Add rpc version mismatch alert

### LINKED ISSUE #769: No error given when dcrwallet version is not appropriate
There is a dcrwallet rpc version check on start up to ensure that the expected built api.proto is being used on a valid verison of dcrwallet bin.  When the error is tripped you can see `VERSION_NOT_VALID` on the console, but decrediton just sits at the Check Wallet Loading screen.

matheusd:
Fix #769

### DIFF
--- a/app/actions/VersionActions.js
+++ b/app/actions/VersionActions.js
@@ -1,6 +1,8 @@
 // @flow
 import { loaderRequest } from "./WalletLoaderActions";
 import { getVersionService, getVersionResponse } from "../wallet/version";
+import { push as pushHistory } from "react-router-redux";
+
 export const GETVERSIONSERVICE_ATTEMPT = "GETVERSIONSERVICE_ATTEMPT";
 export const GETVERSIONSERVICE_FAILED = "GETVERSIONSERVICE_FAILED";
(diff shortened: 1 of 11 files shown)


alexlyp:
So the screen looks good and properly catches the rpc version mismatch.

But I'm thinking the text on the page is a little off, mostly due to the version shown on the About screen is the app version and not the rpc version.  Maybe we can add the rpc version there for wallet  as well?

`dcrwallet {app version} (rpc version {rpc version})`

matheusd:
Do you want me to remove the version from the initial error screen and leave it only on the about page? Or no problem leaving it in the two?

alexlyp:
I think having it in both places is fine

matheusd:
Preview
screenshot_2017-10-19_15-05-44

alexlyp:
cool looks good.  I have an open issue for 1.1.2 that requests some better styling for the about screen so we can shift/adjust layout later.

matheusd:
Feels kinda hacky the way I had to extract the data from the react hierarchy to pass to the about page, but since it (the about page) is outside of the main app store, I couldn't think of a simpler way to do that.

PR #819, opened 2017-10-20
Transitions

Jascha-Sundaresan:
This gets us off react-addons-css-transition-group which has been deprecated.
This also removes two other dependencies: react-router-transitions and react-transitions-group.

All that remains is react-motion, which was previously a dependency of react-router-transitions, but is now in our package.json explicitly instead.

I needed to do this to get us back up to react 16. These transitions will also be much easier to convert to react-router-4 when we make that upgrade.  Getting us back to react 16 fixes the regressions in my work on removing react-tooltip.

Thank you


app/components/views/TransactionsPage/Header.js
```diff
@@ -0,0 +1,79 @@
+import { spring, Motion } from "react-motion";
+import { Link } from "react-router";
+import { injectIntl, defineMessages } from "react-intl";
+import "style/Header.less";
+import headerConnector from "connectors/transactionsPage";
+
+const opts = { stiffness: 150, damping: 15 };
```
alexlyp: I feel like it's a bit bouncy now, after testing a bit `damping: 20` feels good.

app/components/views/TransactionsPage/index.js
```diff
@@ -1,41 +1,28 @@
...
+import Header from "./Header";
+import { RouteTransition } from "shared";
 
-const propTypes = {
-  router: PropTypes.object.isRequired,
-  isTestNet: PropTypes.bool.isRequired,
-};
+const enterLeft = { atEnter: { offset: -100 }, atActive: { offset: 0 }, atLeave: { offset: 100 }, mapStyles: (styles) => ({ transform: `translateX(${styles.offset}%)` })};
+const enterRight = { atEnter: { offset: 100 }, atActive: { offset: 0 }, atLeave: { offset: -100 }, mapStyles: (styles) => ({ transform: `translateX(${styles.offset}%)` })};
+const opts = { stiffness: 150, damping: 15 };
```
alexlyp: Same `damping: 20`

app/style/Header.less
```diff
@@ -1,123 +1,114 @@
...
+
+.tabbedheader-tab {
+  display: inline-block;
+  vertical-align: middle;
+  text-align: center;
+  padding: 0px 4px 3px 4px;
+  margin: 0px 10px 0px 10px;
+  box-sizing: border-box;
+  color: inherit;
+  text-decoration: none;
```
alexlyp: I see the text-decoration set to none, but looks like the tabs have default `<a>` styling.

Jascha-Sundaresan:
I'm closing this PR and opening another one from a fresh branch w/ a cleaner commit history.

app/components/views/TransactionsPage/Header.js
Jascha-Sundaresan: sure.. I got the numbers 150 and 15 from design but I'll change them

app/components/views/TransactionsPage/index.js
Jascha-Sundaresan: I can set defaults inside RouteTransition if you want.

app/style/Header.less
Jascha-Sundaresan: this is from before... I didn't add or change any styling rules.

alexlyp: Well no matter the cause, it needs to be fixed

alexlyp:
No need to close PRs.  Just do you cleaning then force push to this branch

app/components/views/TransactionsPage/index.js
alexlyp: sure

PR #820, opened 2017-10-20
reworks transitions

Jascha-Sundaresan:
Here is a cleaner version of #819


app/components/shared/RouteTransition.js
```diff
@@ -0,0 +1,45 @@
+import { createElement as h } from "react";
+import { spring, TransitionMotion as TM } from "react-motion";
+import { object, string, func, bool } from "prop-types";
+
+const e = (s, o) => Object.keys(s).reduce((a, b) => { a[b] = typeof s[b] === "number" ? spring(s[b], o) : s[b]; return a; }, {});
```
oktapodia: Can you add some explicit variable names, please? Not possible to understand what is doing without spending time on it

app/style/Header.less
```diff
@@ -1,123 +1,120 @@
-@import (reference) "./main.less";
```
oktapodia: Can you revert this file, please ? (encoding issue I suppose)

package.json
```diff
@@ -199,19 +199,17 @@
     "node-addon-loader": "alexlyp/node-addon-loader#master",
     "prop-types": "^15.5.10",
     "qr-image": "^3.2.0",
-    "react": "15",
-    "react-addons-css-transition-group": "^15.6.2",
-    "react-dom": "15",
+    "react": "16",
```
oktapodia: Can you specify the patch version as well like `16.0.x` just to have the security/minor fix, please?

app/components/views/TransactionsPage/Header.js
```diff
@@ -0,0 +1,79 @@
...
+const messages = defineMessages({
+  "transactions.title":               { id: "transactions.title",               defaultMessage: "Transactions" },
+  "transactions.description.testnet": { id: "transactions.description.testnet", defaultMessage: "Testnet Decred addresses always begin with letter T and contain 26-35 alphanumeric characters (e.g. TxxXXXXXxXXXxXXXXxxx0XxXXXxxXxXxX0)." },
+  "transactions.description.mainnet": { id: "transactions.description.mainnet", defaultMessage: "Mainnet Decred addresses always begin with letter D and contain 26-35 alphanumeric characters (e.g. DxxXXXXXxXXXxXXXXxxx0XxXXXxxXxXxX0X)." },
+  "transactions.tab.send":            { id: "transactions.tab.send",            defaultMessage: "Send" },
+  "transactions.tab.receive":         { id: "transactions.tab.receive",         defaultMessage: "Receive" },
+});
+
+@autobind
+class Header extends React.Component {
```
oktapodia: Can you use the propTypes please?

app/style/Header.less
Jascha-Sundaresan: "./main.less" is still there.  this is just a strange bug in git thinking I did more than I did in this file. You'll see it in the green below

app/components/shared/RouteTransition.js
Jascha-Sundaresan: sure.. it'll add about 15 lines of code to make this whole thing explicit.

app/style/Header.less
oktapodia: Yes I saw that^^ that's about your encoding I think

app/components/views/TransactionsPage/Header.js
Jascha-Sundaresan: sure

package.json
Jascha-Sundaresan: done

Jascha-Sundaresan:
Heh. so I figured out the source of that bug.

One of the main features of react motion is that it won't unmount something until it's done animating. Unmounting animations are one of the main challenges w/ react.

In this particular case, when you click back to the transactions page that quickly, the caret hadn't finished its movement, so it's returns to the place it moved last.

I'll figure out how to fix this.

Jascha-Sundaresan:
done.

app/components/views/TransactionsPage/index.js
```diff
@@ -1,41 +1,28 @@
...
-        ]}
-        children={this.props.children}
-      />
-    );
-  }
-}
+const Transactions = ({ children, location }) => {
+  const tabs = ["send", "receive"];
+  const page = "transactions";
+  const pathname = location.pathname.split("/")[2];
```
go1dfish: This is slightly hacky IMO (could break tests?) we should mark it as such and come back around later.

I expect this can be handled better once we upgrade to router v4.  I'm ok with it for now but would like to avoid direct accesses to browser globals as much as we can.

Jascha-Sundaresan: Yes.. this will be gone once we upgrade to react router 4

oktapodia: Can you had a TODO and create and issue when this PR will be merge just to not forget, please?

package.json
oktapodia: Sorry my comment was not accurate enough, can you add `16.0.x` or `~16.0.0`

app/components/views/TransactionsPage/index.js
Jascha-Sundaresan: sure.

PR #825, opened 2017-10-22
Remove react tooltip

Jascha-Sundaresan:
Still got a couple styling issues to fix, but this is almost done.


Jascha-Sundaresan:
This is now complete.

It addresses #567

app/components/shared/index.js
```diff
@@ -1,3 +1,6 @@
 export { default as CopyToClipboard } from "./CopyToClipboard";
 export { default as RouteTransition } from "./RouteTransition";
 export { default as Tooltip } from "./Tooltip";
+export { default as LinkToAccounts } from "./LinkToAccounts";
+
+export const Aux = ({ children }) => children;
```
oktapodia: What is the benefit of this helper?

Jascha-Sundaresan: It's a self immolating component.

This allows us to build components using a wrapper that doesn't appear in the final website markup.

I used it on the main page views so that they're placed directly into the markup without a wrapper "div".

This resolves styling issues without the need to create extra classnames to propagate parent layout styles (like height: 100%).

This is one of the new patterns from react 16. It's really powerful. :)

alexlyp:
needs a rebase and the yarn.lock looks like it has extra changes outside of just the react-tooltip removal

app/components/PurchaseTickets/Form.js
```diff
@@ -249,56 +208,56 @@ const PurchaseTicketsForm = ({
...
 
-      {canAffordTickets ? null : (
+      { !canAffordTickets &&
         <div className="stakepool-purchase-error">
           <T id="purchaseTickets.errors.insufficientBalance" m="Insufficient spendable account balance to purchase tickets." />
+        </div> }
+
+      <Tooltip warning disabled={ !rescanRequest }
+        text={ <T id="purchaseTickets.importDisabledRescan" m="Importing scripts is disabled during a rescan."/> }>
+        <div className="stakepool-content-import-script-button">
```
alexlyp: Something wacky is going on here and causing Import Script button to get pushed to a lower line.  Should be in line with Purchase button at all times.  I see no changes to css of the import script button so must be some other issue

Jascha-Sundaresan: fixed

PR #834, opened 2017-10-24
Startup improvements

### LINKED ISSUE #723: Decrediton cannot be restarted immediately after it's shut down
There's a small amount of time after closing Decrediton where you cannot reopen it, because dcrd/wallet are still running.

@jrick proposed a good solution: Hide the Decrediton window when the user closes it and don't close it until the background processes are killed. If the user tries to start Decrediton while the previous instance is still running, show an error.

### LINKED ISSUE #727: Cannot start decrediton twice
Bug report from user willywoo on slack. Running two instances of decrediton thrown an error due to already running daemons. It should report a better error.

Maybe it can be worked with #723.

matheusd:
Fix #727 
Fix #723

A few notes:

electron-debug has been removed, as all functionality was already manually implemented (except for reloading the main window).

Regarding reloading: I've added the default F5 shortcut for activating it (as electron's main menu only supports one shortcut). If the other combinations are needed (cmd+r, ctrl+r, etc), then I can re-add them manually as well. Shouldn't be an issue as it is only supposed to be used in ~production~ development.

Instead of hiding the window as suggested in #723, I'm sending the user to a shutdown page until the daemon finishes quitting. This is because, in some instances, the shutdown may take a long time and so just hiding the window may make an user think decrediton has completely shut down when it is in fact running in the background.

This will require testing in all platforms, to make sure the shutdown procedure is working correctly.


alexlyp:
tACK on linux deb8.  code looks good but i'll hold off on merging until we can get someone to verify everything works as expected on macOS and win

alexlyp:
Looks like some minor conflicts to resolve.  Just make sure you add daemonStopped to the other possible daemon start actions.

matheusd:
I've rebased and resolved the code conflicts, but I seem to be having some issues after the #793 merge.

Running out of master, if I press F5 to reload UI I get a dcrd daemon error. Is that happening to you too?

Also, that PR broke the daemon_skip_start flag (or I guess superceded it)?

matheusd:
Ok, I've noticied now you've done some work after 793, I'm revewing the changes on the startup procedure

PR #838, opened 2017-10-25
Added instructions for install via brew cask

josser:


oktapodia:
When I'm trying to install it via cask I have the following error message `Error: Cask 'decrediton' is unavailable: No Cask with this name exists.` Do we have a cask decrediton version?

josser:
@oktapodia you have to do brew update before

josser:
Here is formula: https://github.com/caskroom/homebrew-cask/blob/master/Casks/decrediton.rb
so it should just work, just update your local brew

PR #841, opened 2017-10-26
Add styling to handle long account names

### LINKED ISSUE #806: Long account names need to be abbreviated
We need abbreviated account names on dropdowns and history rows.  Possibly look for other places to change as well

alexlyp:
Fixes #806


Jascha-Sundaresan:
heh.. I had no idea _where_ you wanted that logic placed.

Glad you were able to jump on it.

Issue #798, opened 2017-10-16
Update main navigation motion design


linnutee:
![decrediton - motion - navigation](https://user-images.githubusercontent.com/17774057/31638925-c6a7245c-b2de-11e7-9c65-08cdbe778424.gif)

Motion design specs:
Hover state on text: 
- Change the text color/opacity from `#c4cbd2` to` #ffffff`(Quart InOut) -> 130 ms

Unhover: 
- Same thing in reverse

Clicked:
- Selection rectangle moves in Y to new position below pressed item -> 210 ms (Expo InOut)
- Text color on previous item changes back to `#c4cbd2` -> 130 ms (Quart InOut)

alexlyp:
Nice looks great.

Jascha-Sundaresan:
I'll look into this.

Jascha-Sundaresan:
I think this issue can be closed now

PR #843, opened 2017-10-29
adds animation to sidebar

Jascha-Sundaresan:
fulfills #798

I used the same spring as the enterLeft/Right of the tabbed pages for a consistent flow


alexlyp:
Motion looks and feels good, possibly a touch stiff, but we can tweak that.  

I'm seeing a x-scroll currently though, once that's fixed I'll do a review and get this in.

PR #844, opened 2017-10-30
New tickets page

### LINKED ISSUE #735: Initial My Tickets mock up
First pass can be totally static information while the backend is updated to provide the required ticket status information.

matheusd:
Fix #735 

Sending early for initial review (still a WIP). Notes:

~- The information is not yet correct, as it depends on [dcrwallet's #981](https://github.com/decred/dcrwallet/pull/981).~ (merged)
~- There are some console.log() sprinkled around, these are being used to debug the info (see above point).~ (shouldn't have any more)
- No page transitions yet (edit: transactions don't have transations between pages yet either, we can add later)

Need input on whether to keep or remove the initial per-type ticket count. 

screenshot_2017-10-30_10-02-46
screenshot_2017-10-30_10-03-02


alexlyp:
on a first pass seems to work great.

matheusd:
Next round of improvements. Now filling with the proper data. I'll update the PR in a bit.

screenshot_2017-10-31_16-27-25
screenshot_2017-10-31_16-27-35

matheusd:
If you try and run this without the relevant dcrwallet PR, you won't get the vote and pool fee information. The other information should be available.

app/selectors.js
```diff
@@ -278,6 +291,109 @@ export const viewedTransaction = createSelector(
...
+      const decodedSpenderTx = hasSpender ? (decodedTransactions[spenderHash] || null) : null;
+
+      // effective ticket price is the output 0 for the ticket transaction
+      // (stakesubmission script class)
+      const ticketPrice = ticketTx.getCreditsList()[0].getAmount();
+
+      // ticket tx fee is the fee for the transaction where the ticket was bought
+      const ticketTxFee = ticketTx.getFee();
+
+      // missing: split tx fee
```
matheusd: I'm not entirely sure if I should include the fee for the transaction before the ticket or not.

app/selectors.js
```diff
@@ -278,6 +291,109 @@ export const viewedTransaction = createSelector(
...
+
+      // ticket change is anything returned to the wallet on ticket purchase.
+      // (don't know if it is currently possible to have change due to split transactions)
+      const ticketChange = ticketTx.getCreditsList().slice(1).reduce((a, v) => a+v.getAmount(), 0);
```
matheusd: I **think** this is not needed as the ticket transaction should only produce the ticket commitment output, but I'm also not entirely sure, so I decided to include this here.

app/components/views/TicketsPage/MyTicketsTab/TicketInfoCard.js
```diff
@@ -0,0 +1,68 @@
...
+      poolFee: <Balance amount={ticket.ticketPoolFee || 0} />,
+      reward: <Balance amount={ticket.ticketReward || 0} />,
+      roi: ticket.ticketROI
+    }} />;
+
+  let timeToLeaveTipText;
+  if (ticket.leaveTimestamp) {
+    const days = Math.ceil((ticket.leaveTimestamp - ticket.enterTimestamp) / 86400);
+    timeToLeaveTipText = <T id="ticket.daysToLeave"
+      m="~ {days, plural, one {# day} other {# days}} from buying until {status}"
```
matheusd: Not really happy with this copy, if anyone has a better suggestion.

app/components/views/TicketsPage/MyTicketsTab/TicketInfoCard.js
```diff
@@ -0,0 +1,68 @@
...
+
+const TicketInfoCard = ({ ticket, onClick, expanded }) => {
+
+  const className = "ticket-info-card" + (expanded ? " is-expanded" : "");
+  const rewardTipText = <T id="ticket.rewardCalc"
+    m={`Investment: {investment}
+    Transaction Fee: {txFee}
+    Pool Fee: {poolFee}
+    Ticket Reward: {reward}
+    ROI: {roi, number, precise-percent}`}
```
matheusd: Maybe this should make more clear that the ROI is per-ticket? Or no need for that?

app/helpers/tickets.js
```diff
@@ -0,0 +1,80 @@
...
+    return null;
+  }
+
+  let vote = outputScript.slice(2, 4).reduce((a,v,i) => a|(v<<(i*8), 0));
+  let version = outputScript.length > 4 ? outputScript.slice(4,8).reduce((a,v,i) => a|(v<<(i*8)), 0) : 0;
+
+  // TODO: currently hard coded because dcrwallet doesn't return all
+  // agendas (only the active ones). All agendas are needed for historical
+  // votes.
+  let agendas = {
```
matheusd: I'm not sure whether this is a known issue with dcrwallet's grpc or not. It's not ideal to have all the agendas duplicated here, but as long as they aren't being created too fast, shouldn't be an immediate problem.

app/components/views/TicketsPage/MyTicketsTab/TicketInfoCard.js
alexlyp: I think it's fine for now, we'll be adding everything up for clear statistics on the next pass. So for now it's ok to display ROI on a ticket by ticket basis.

app/helpers/tickets.js
alexlyp: Yah not much we can do at the present moment, i think we are still thinking about a way to "save" past agendas for use later.  Fine for now

app/selectors.js
alexlyp: don't worry about the split fee, may be going away entirely soon.

app/selectors.js
alexlyp: Yah this will need to be smarter coming up due to the possible removal of the splittx and use of the ticket change.  But should be fine for a first pass.

app/components/views/TicketsPage/MyTicketsTab/TicketInfoCard.js
alexlyp: I think it's clear what you're trying to say.  Can you add a note for the 86400 though?  Like `// 86400 seconds = 1 day`  and maybe even do `24 * 60 * 60` instead of 86400

alexlyp:
overall looks great and works great.  only issue i see is i'm not sure what the value that the revoke tickets are showing.  is that the revoke value?  it should do the same calculation that vote - ticket, so it'll be negative as it was a net loss due to pool fees and ticket fees.

matheusd:
It's showing the original ticket price, same as vote. 

What maybe I should add is a tip with the fees and ROI calculation, what do you think? That way it's symmetrical between vote and revoke

alexlyp:
Right it should be the same as Vote minus the icon that is shown i believe

app/helpers/tickets.js
```diff
@@ -0,0 +1,80 @@
...
+        { mask: 0x06, name: "DCP0001", choices: {0x02: "no", 0x04: "yes"},
+          defaultChoice: "abstain"}
+      ],
+      6: [
+        {mask: 0x06, name: "DCP0002&3", choices: {0x02: "no", 0x04: "yes"},
+          defaultChoice: "abstain"}
+      ],
+    },
+    mainnet: {
+      5: [
```
alexlyp: Mainnet should show 4 for dcp0001 and 5 for dcrp0002&3.  And would this parse correctly if we added the proper multiple agendas for Mainnet-4 or Testnet-5?  

https://github.com/decred/insight-api/blob/23d38a0bfcc3cafebe9318d288891512fff33eb1/lib/Rpc.js#L65

Masks and choices are slightly different for the lnsupport choice.

app/components/views/TicketsPage/MyTicketsTab/TicketInfoCard.js
```diff
@@ -0,0 +1,75 @@
...
+        txFee: <Balance amount={ticket.ticketTxFee || 0} />,
+        poolFee: <Balance amount={ticket.ticketPoolFee || 0} />,
+        reward: <Balance amount={ticket.ticketReward || 0} />,
+        roi: ticket.ticketROI
+      }} />;
+  } else if (ticket.status === "revoked") {
+    returnTipText = <T id="ticket.revokeCalc"
+      m={`Investment: {investment}
+      Transaction and Revocation Fees: {txFee}
+      Revocation Relay Fee: {relayFee}
```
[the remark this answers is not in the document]
matheusd: Will do. I actually wrote initially as Pool fees, but I was checking the dcrwallet code and somewhere was written as "relay fee" so I ended up using that.

PR #848, opened 2017-11-01
Swap inputs to semantic versions

### LINKED ISSUE #777: Update input fields styling
screen shot 2017-10-13 at 01 32 00

Just about the same thing as for dropdowns: https://xd.adobe.com/view/82325302-2eb8-4684-9b46-7ef1bc6b227a/

[UI components specs v002.pdf](https://github.com/decred/decrediton/files/1380915/UI.components.specs.v002.pdf)

Since last version I've updated the default states type color to `#C4CBD2`
- Input field A = most basic field with normal bodycopy … e.g destination address, account names, passwords etc. 
- Input field B = large numbers + type size script, this is used for displaying Amounts in DCR. The full numbers and two first decimal spots are with a larger typeface (19px), while all following decimal spots are with smaller (13px). 
- Input field C = has only large numbers, no sizing script. (Not sure atm if this may become obsolete or not, but I included it just in case)
- Input field D = small numbers.

matheusd:
Fix #777.

Haven't touched the sign/verify page as that is being worked on another PR.

This doesn't add some stuff that was discussed on slack (validation on amounts/address/etc, ability to change the unit for a given input), it just converts all inputs into their semantic versions for future improvements.


app/components/TicketAutoBuyer/Form.js
```diff
@@ -130,7 +129,7 @@ const TicketAutoBuyerForm = ({
             </div></div>
               <div className="stakepool-purchase-ticket-num-input">
                 <div className="stakepool-input-form-purchase-ticket">
-                  <DecredInput
+                  <DcrInput
```
alexlyp: This should be a FeeInput

Issue #797, opened 2017-10-16
Update headers styling


linnutee:
Preview: https://xd.adobe.com/view/c09a14cb-4f5d-42fe-9226-a453cb0cbd80/
Icons: [decrediton - header icons - 171017.zip](https://github.com/decred/decrediton/files/1389141/decrediton.-.header.icons.-.171017.zip)

Typesetting and colors on the headers and bodycopy needs to be gone over through-out. Included specs below.

**Overview**
- Update font size for the balance numbers (current one is smaller)
- Update Rescan button position, aligned with the type baseline (32px from bottom)
- Rename button to "Rescan"

**Transactions**
- Add icon
- (?) As far as I remember the last decision was to include History under Transactions? If that's not the case I will also add a History icon here. @alexlyp can you verify? 

**Tickets**
- Add icon
- Re-align/add current ticket price to left. 

**Accounts**
- Add icon
- Add bodycopy

**Security Center**
- Add icon
- Add submenu
- Optional: write a relevant copy to wrap up the function of the page.* 

**Settings**
- Add icon

**Help**
- Add icon
- Add bodycopy

<img width="746" alt="screen shot 2017-10-17 at 00 47 27" src="https://user-images.githubusercontent.com/17774057/31636849-d55ec392-b2d4-11e7-8975-9cfc75830c6e.png">
<img width="880" alt="screen shot 2017-10-17 at 00 22 40" src="https://user-images.githubusercontent.com/17774057/31636850-d57f0b7a-b2d4-11e7-8b05-4b6987154867.png">
<img width="664" alt="screen shot 2017-10-17 at 00 51 07" src="https://user-images.githubusercontent.com/17774057/31636974-51824d22-b2d5-11e7-83f4-871772a0ac9e.png">

Jascha-Sundaresan:
I can jump on this and do it as soon as @sndurkin and @oktapodia sign off on the toolset I'd use to do it in #786

Jascha-Sundaresan:
how do we want the ticket price to look?  like the balance on the overview page or the balance on the history tab?

Jascha-Sundaresan:
also, do we want the ticket price visible on all 4 tabs or just the purchase tickets one?

Issue #767, opened 2017-10-10
Styling of the revoke notification on the HomePage


alexlyp:
With the newly added feature of the revoke notification on the home page we need to add proper styling for it.  This should include a check for revoke tickets request attempt being true/false to show a loading/feedback fot the user.

alexlyp:
Please do not attempt this issue without explicit input from @linnutee or @sander

Jascha-Sundaresan:
@linnutee, can you chime in here w/ some input?

PR #855, opened 2017-11-05
Headers

Jascha-Sundaresan:
This addresses #797 and #767


alexlyp:
We briefly discussed the breadth of this PR on Slack, but I want to leave a comment here for posterity. PRs need to be much, much smaller and more focused than this.  Typically you should only be attending to 1 particular issue per PR.  Obviously, there are circumstances that don't allow for this, but most times you should split out work between 3-4 smaller PRs.  I will be letting it slide this time, but from here on out I'll be requiring this and nACKing any PRs that don't abide by this rule.

alexlyp:
Error/success message on accounts page pushes everything down. should float on top.

alexlyp:
Add new account button no longer works.  "dispatch not defined" in console

alexlyp:
When going from: Transactions/History to: Transactions/Receive it shows the y-scroll for a brief moment then disappears

app/helpers/index.js
```diff
@@ -0,0 +1,22 @@
+import { createElement as h } from "react";
+export { tsToDate } from "./dateFormat";
+export { addSpacingAroundText, restrictToStdDecimalNumber } from "./strings";
+
+export const kidCheck = C => {
```
alexlyp: Can you please add comments for the less than clear helper funcs please?

alexlyp:
Tickets page needs `Current Price: ` as requested.

app/components/shared/Icon.js
```diff
@@ -0,0 +1,21 @@
+import icons from "iconMap";
```
alexlyp: I'd like some input from @go1dfish, @matheusd  @oktapodia concerning the iconMap usage.

app/helpers/index.js
Jascha-Sundaresan: added

app/components/shared/Icon.js
Jascha-Sundaresan: this pattern allows us to add new icons by only adding a couple lines of code and reusing the same component instead of adding file, adding a line referencing the file in Icons.less, then creating a classname that sizes a div and places the icon in as a background image, centers it, sizes it, and makes sure it doesn't repeat, which still necessitating we put extra padding around things next to icons because they don't actually take up space.

Jascha-Sundaresan:
message styling has been fixed, as has the new account button. need styling direction to add the ticket price, and I've added comments to the HOCs 

I still need to address the y-scroll. I don't see a scroll bar when switching back and forth but if I scroll down on the history tab and then switch back to one of the other ones i'm still scrolled down until the animation finishes.  is that what you mean?

app/components/shared/Icon.js
matheusd: I'm not completely against it and I agree with Jascha that there is somewhat a bit of work involved in adding a new icon today (add the svg file, add the variable referencing it then apply the correct styles into a div).

But, I don't like embedding the icon contents inside js. I'd prefer the svg file be separated as that makes it easier to see the available icons on the file explorer, edit the icon if needed, replace it, etc.

My other concern is the consistency of the codebase. We should decide whether we want this style of adding icons and if we do, then we should open an issue and work to replace all icons to this new style. Otherwise, if we decide that we will *not* replace the existing icons with this new style, then this should be reworked to follow the existing conventions. Having half the icons coded in one way and half in another is bad IMO.

oktapodia: Like @matheusd said, I prefer to keep the svg files (and that's better for performances) instead of full JS stuff

For today, I think we should improve the SVG loading by doing something like that (quick example):
```
import iconTest from '../icons/iconTest.svg';

<img src={iconTest} />
```

Jascha-Sundaresan: Keeping svg files is not better for performance.

This solution is as fast as it can get. Nothing is faster in js than object lookup.

If you want a convenience for knowing the icons available, we can use the iconmap directly like so

`<Icon i={ iconMap.testnet } />`

This manner enables autocomplete when filling in the icon type.

Oktopedia's proposed solution doesn't allow for sizing or programmatically changing color of a monochrome icon.

Jascha-Sundaresan: As far as consistency of the codebase, I'd be happy to convert the other icons after 1.1.2 release

Jascha-Sundaresan: As far as consistency of the codebase, I'd be happy to convert the other icons after 1.1.2 release

Keeping it in js also benefits performance in that we don't require an additional step of loading assets

oktapodia: Agree, my example was bad but I still prefer a separated files, it will be easier to maintain and to update.

Interesting conversation about the splitting files [https://github.com/facebookincubator/create-react-app/issues/1388#issuecomment-272476528](https://github.com/facebookincubator/create-react-app/issues/1388#issuecomment-272476528)

alexlyp: ok after hearing other points on this, i'm going to ask for this portion be reverted and the svgs be used instead.  this can be put up for discussion after release, but it sounds like the current consensus is nACK

Jascha-Sundaresan: I added a bunch of new icons for the headers and the testnet logo, so it wouldn't be a reversion, but an addition of code. @oktapodia agreed that his example was bad. Maintaining and updating are significantly easier when only 2 lines of code need to be added instead of a file which still needs to be read to resize correctly when declaring an icon className for it.

I wrote the Icon component because it was faster for me to do than adding the icons the current way, and now it's even faster because the component is already built. When @linnutee gave me the Testnet icon to add to the Sidebar, it took me less than 30 seconds to add it.

As far as @matheusd's concerns about 
> easier to see the available icons on the file explorer, edit the icon if needed, replace it, etc.

I addressed those by suggesting the pattern of importing the constant when consuming the component. That benefits from autocomplete which allows you to easily see what's available.

Opening the iconMap file in your editor while working w/ a command-T and autocomplete on filenames is faster than switching to a file explorer and opening tabs (that's sublime's shortcut but I imagine all editors have an equivalent).

Replacing and adding icons is significantly faster in the paradigm I'm proposing. That's the _only_ reason I did it. This enables a faster turnaround time w/ design's deliveries.

@alexlyp can we at least wait until @matheusd chimes in to my response to his concerns before calling it a nACK? We haven't heard from @go1dfish yet either.

Regardless, I won't be able to get to this until after 5. The workday has started.

Jascha-Sundaresan: Also, @oktapodia, the conversation you linked goes straight to someone suggesting that inline svg is the best solution.

This is what my implementation does.

alexlyp: I legit have no idea how to add icons via this method, so it's a pretty strong nACK for me.  I get that you think this is a great addition, but for everyone else it'll be a struggle to do things that were previously simple.

Jascha-Sundaresan: I can teach you in under 5 minutes.  Can it wait until after work please?

Jascha-Sundaresan: Also, here's where I got the idea of importing the iconMap constant for developer experience.

https://medium.com/@david.gilbertson/icons-as-react-components-de3e33cb8792

oktapodia: > I wrote the Icon component because it was faster for me to do than adding the icons the current way, and now it's even faster because the component is already built. When @linnutee gave me the Testnet icon to add to the Sidebar, it took me less than 30 seconds to add it.

Please @Jascha-Sundaresan stop thinking about yourself but think as a team, something faster for you is not necessarily easier/faster for anyone else and like you mention in the other pull request

> Regardless, this is a discussion for after we release 1.1.2

Jascha-Sundaresan: > write a one time tool to pick up the code of all icons and insert them into the IconMaps.

This may be pretty straightforward. if we wanted to completely standardize it, we could always take the whole viewbox instead of using height and width when the viewbox is a simple rectangle or square. That would still allow us to set the height by by the exposed prop.

Right now, I just used svgo to minify it and pull out everything except the surrounding svg tag.

The only catch I see to this is that we'd want to handle our monochrome icons differently so we still have the ability to paint them programmatically instead of relying on a separate icon for each color (like the ticketscogs open/closed). That requires the extra step of pulling out the fill values, or, ideally, design squashes the svg so that the export is a single path whenever possible.

A simple answer for this would be to have the monochrome icons separated into a subfolder `icons/monochrome` and then have a different script for them.

As to "thinking as a team", I'm trying to do exactly that by introducing new tools that admittedly have a non-zero learning curve but have the ability to increase productivity tremendously.

Your script idea is a good one and I'd be happy to build both the colored and monochrome versions if that will improve the experience of the others on the team.

oktapodia: Personally, I don't want to overstuff the application, SVG files are good, used and known by many peoples, I prefer to keep the thing simple, to be honest.

alexlyp: I'm with @oktapodia here, if we were planning on adding tons and tons of more icons, i could see the utility in this.   we have a very small number of icons that aren't going to be changing very often.  spending time on trying to perfect every little part in the code is not a productive use of time.

Jascha-Sundaresan: This isn't just about the size of our icon library. It's about the interactivity of the user interface. 

Right now, in order to do something a simple as change an icon's color on hover, we need to store two different icons and swap out a div's background image from one icon to the other. We can't treat the icon like a real component and simply set its color. 

There's no possibility for transition animations w/ this approach either.

alexlyp: I'll close this PR and just do it myself if you don't want to, why do you need to be this stubborn over something so small?  I shouldn't be wasting my time commenting on things like this when there are real problems and bugs to be fixing.  This will be my last comment in this PR.

alexlyp:
small travis issue for unused Icon

Issue #862, opened 2017-11-07
Advanced daemon on reload


alexlyp:
Currently doing a F5/ctl-R/location.reload()  doesn't work quite correctly when using the advanced daemon settings.  

Ideally, if a user uses either remote daemon or a different appdata, that selection/data should be retained for use if/when the user reloads the electron instance.

alexlyp:
Case #1:  User changes advanced daemon to true in config and chooses to use an alternate appdata directory.  Upon reload attempt (ctrl-r/F5), the fact that an alternate appdata was used should be known (via config possibly) and the user should not be sent to the advanced daemon screen.  This alternate appdata should then be used for the appData store information for startRpcConsensus.  When decrediton is fully shut down, this saved appData can be retained and possibly used to pre-populate the appData field on the advanced daemon form.  But the form should be shown nonetheless.

Case #2:  User changes advanced daemon to true in config and choose to use a remote daemon.  Upon reload attempt (ctrl-r/F5), the fact that a remote dameon was used should be known (via config possibly) and the user should not be sent to the advanced daemon screen.  This remote credentials should then be used for the rpcCreds store information for startRpcConsensus.  When decrediton is fully shut down, this saved rpcCreds can be retained and possibly used to pre-populate the fields on the advanced daemon form.  But the form should be shown nonetheless.

PR #867, opened 2017-11-08
Advanced daemon reload

vctt94:
fixes on #862 

case 1 working. 
I created a must_open_form which I set for true when daemon is synced.

### DIFF
--- a/app/components/views/GetStartedPage/index.js
+++ b/app/components/views/GetStartedPage/index.js
@@ -8,6 +8,7 @@ import { FinalStartUpHeader, FinalStartUpBody } from "./FinalStartUp";
 import { DaemonLoadingHeader, DaemonLoadingBody } from "./DaemonLoading";
 import { AdvancedStartupHeader, AdvancedStartupBody } from "./AdvancedStartup";
 import { walletStartup } from "connectors";
+import { getMustOpenForm, getAppdataPath } from "config.js";
 
 @autobind
 class GetStartedPage extends React.Component {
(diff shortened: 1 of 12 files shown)


app/components/views/GetStartedPage/index.js
```diff
@@ -31,6 +32,11 @@ class GetStartedPage extends React.Component {
   }
 
   render() {
+    const skipForm = getMustOpenForm() === false && getAppdataPath().length>0;
+
+    if(skipForm) {
```
alexlyp: You need to move this check up into `componentWillMount` above.  `render()` causes multiple incorrect requests to onStartDaemon.

Also you should put skipForm into the selectors:  feel free to diff my branch for the changes https://github.com/alexlyp/decrediton/tree/ayp_startup_fix

app/components/views/GetStartedPage/index.js
```diff
@@ -19,7 +20,18 @@ class GetStartedPage extends React.Component {
...
+    const hasAppData = getAppdataPath().length > 0;
+
+    if(hasAllCredentials && hasAppData)
+      this.props.setCredentialsAppdataError();
+
+    if (!this.props.openForm && hasAppData) {
+      this.props.onStartDaemon(null, getAppdataPath());
+    } else if (!this.props.openForm && hasAllCredentials) {
+      this.props.onStartDaemon(getRemoteCredentials());
+    } else if (!this.props.isAdvancedDaemon) {
```
alexlyp: After doing some testing it's clear that this should be the first check done.  On !advanced reload it is using the previously entered rpc credentials I had entered.  I'd just put:

`if (!this.props.isAdvancedDaemon) { this.props.onStartDaemon(); break; }` at the top of componentDidMount()

Issue #815, opened 2017-10-19
Pagination refactor/rewrite for GetTransactions/GetTickets


alexlyp:
Currently the GetTransactions and GetTickets are requested on start-up for the entire set of transactions and tickets.  This is obviously not ideal, though it appears that it still runs without much noticeable slowdown on large wallets we should be requesting the transactions on a page by page basis.  

With the current method of providing start/end block for GetTransactions/TicketsRequests it's not trivial to just ask for page M of N.  I'd like to open up discussion here about possible ways to fix this and implement an elegant pagination solution.

jrick:
FYI, the wallet will need some work done to incrementally return these transactions in a stream before this change makes sense in the client.  Currently all of these transactions are being loaded at once before being streamed to the client.

matheusd:
The work jrick mentioned is referenced in this issue:

https://github.com/decred/dcrwallet/issues/768

PR #875, opened 2017-11-13
Switch to paginated GetTransactions

### LINKED ISSUE #854: Implement "Sorting" to Transaction History
screen shot 2017-11-04 at 23 29 56

Preview: https://xd.adobe.com/view/2f6756f6-5479-40af-9db3-9202d5029d36/

This applies for the Transaction History view. Don't think there's much point adding it to My Tickets also?

Sorting button uses the same "Context Menu" as seen in #730 for Display filter (see specs there also). 
- On hover it should display tooltip with message "Sort by". 
- Allows to sort **visible transactions** by "Newest", "Oldest", "Smallest Amount", "Largest Amount". 

Also if needed update the Display filter: 
- On hover displays tooltip with message "Show"
- Allows users to show: "All", "Received", "Sent", "Invalid (yes/no @alexlyp?)"

This partially addresses #785 – but I would not add that complex amount selector feature atm. That should be addressed together with adding a proper Search function, which will should go hand-in-hand with filtering/sorting + possibly needs these to be re-worked also.

Icon: [sort-by.svg.zip](https://github.com/decred/decrediton/files/1442867/sort-by.svg.zip)

matheusd:
This commit switches the transactions from being loaded all at once during startup to being loaded on demand, on a paginated basis.

Fix transactions part of #815.
Fix #854 

Depends on https://github.com/decred/dcrwallet/pull/988


matheusd:
Remarks on the code so far follows.

Still needed:
[ ] Support unmined transactions
[ ] Support new transactions (received from notifications)
[ ] Improve UI for loading older transactions (either by adding infinite scrolling or styling the "more transactions" button).
[ ] Cleanup code

Sample demonstrating the new paginated nature of `getTransactions`. Each click of the button initiates a new `getTransactions` grpc call, filters the returned results until `maximumTransactionCount` is received, and then stops requesting new transactions.

In the demo, I'm using a page size of 2 transactions for demonstration purposes. Notice that (specially when filtering the revocations) more than 2 transactions are shown - that is due to the fact that dcrwallet iterates by block, and a block may return an amount of txs that makes it pass the limit. Don't think it will be a problem, but can be reworked with a bit more smarts on the transaction fetching logic if needed.

For the moment, this also prevents the use of a full paginator, since we lose the information of how many transactions exist in the wallet (and how many fulfill the selected filter). If that information is needed, then new grpc calls are going to be needed on dcrwallet, either to implement filtering there or to return the total amount of transactions for the wallet (either all or by type).

streamed-transactions mp4

alexlyp:
Started doing some testing on this PR with the required dcrwallet PR.  

Overall, I think it works well.   Obviously still some UX issues to work out to for the "more transactions/paginations" but otherwise works as expected.

alexlyp:
Maybe we can at least up the request number to something sensible like 6, then possibly fix the tx within a block

matheusd:
Yeah, I left at 2 as a try out, to make it obvious what was going on. 

I think the best interaction would be an infinite scrolling: load enough transactions to fill the screen (probably around 6-10), when the user scrolls enough that the "more transactions" button is visible, start loading additional transactions. I didn't implement that because this PR was already growing big enough.

alexlyp:
Yah agree on that.  Once dcrwallet PR is in, I'm ok with this going in and for you to pursue endless scrolling.

Issue #879, opened 2017-11-14
Output row error expanding even when no errors are seen


alexlyp:
It's causing a slight bump after onBlur from the fields.

Instead it should already have an error area designated and just populate those errors if necessary.

alexlyp:
Fixed

PR #891, opened 2017-11-21
ui: Send button validation

### LINKED ISSUE #542: validateaddress
If decred/dcrwallet#885 gets done then a way to validate addresses from the decrediton gui needs to be added.

### LINKED ISSUE #822: Update Send button – state change detection
When entering amount, users need to do an extra click somewhere before the Send button changes to enabled state (blue). 

This should happen instantly when valid tx address and some amount are entered.

matheusd:
This makes the validation for sending a transaction to happen when the address/amount changes, instead of requiring an onBlur event on the inputs.

Fix #822 
Fix #542 

Requires https://github.com/decred/dcrwallet/pull/1001


alexlyp:
So looks good from my testing.

Only input I have is to possibly do some pre-validation to avoid needing to send every onChange to ValidateAddress.  base58 hashed addresses must be at least 25, but can be as long as 35 iirc.

matheusd:
Ok, I added a simple validation for length and network prefix on the addresses. I can probably cook up a full base58 validation if you'd prefer that (there are a couple of js snippets for bitcoin address validation around, I could probably adapt one for decred address decoding).

Also, could you please check whether this fixes #879? I believe it does, but I wasn't aware of that issue before working on this (@vict94 pointed it out on slack).

alexlyp:
Yah looks like it fixed #879 I'll do some further stress testing to confirm that

alexlyp:
I'll open an issue to do the full base58 validation, probably overkill but worth being thorough I suppose

PR #895, opened 2017-11-21
Show full tx details in transaction details page

### LINKED ISSUE #870: Use DecodeMessageService to decode transaction details
Currently we just use what wallet has provided us to render the transaction details page.  Now we can get the full transaction information decoded on demand to show all the information about a given transaction.

matheusd:
Fix #870 

Transaction decoding is requested on demand, only when the transaction details are viewed.

### DIFF
--- a/app/actions/DecodeMessageActions.js
+++ b/app/actions/DecodeMessageActions.js
@@ -16,6 +16,7 @@ export const getDecodeMessageServiceAttempt = () => (dispatch, getState) => {
 };
 
 export const DECODERAWTXS_SUCCESS = "DECODERAWTXS_SUCCESS";
+export const DECODERAWTXS_FAILED = "DECODERAWTXS_FAILED";
 
 // decodeRawTransaction requests decodification of a list of hex transactions.
 // Dispatches the event when all transactions have been decoded. Better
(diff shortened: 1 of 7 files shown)


matheusd:
I've noticed something funny with the non-wallet inputs/outputs: The transactions which were moved from unmined to mined while the wallet was running don't have the value of the input recorded. Transactions which were received while the wallet was offline (meaning, when the wallet was restarted it found only the mined transaction, **not** the unmined) **do** have the amounts for the inputs recorded.

Seems to be something on the way dcrwallet is recording the transactions into the db, but I haven't been able to nail down the exact cause yet.

Also, addresses for non-wallet outputs don't seem to be decoded.

Here's a preview on how it looks, when the transaction has the input amounts.

screenshot_2017-11-21_19-10-59

alexlyp:
Seeing the undefined addresses as noted, but unable to reproduce the offline/online unmined/mined transactions issue you mention. 

Thought I did recently see an error about duplicated div keys for regular transactions today.  Believe when a transaction was supposed to go from unmined to mined.

matheusd:
Do the inputs of a received tx show their value?

This is how I got inputs without value:

- Open Decrediton. Let it sync.
- Go to testnet faucet. Request coins into wallet.
- Wait a few blocks to let tx go from unmined to mined (with decrediton open).


I'll try to recreate the wallet later today, see if it makes any difference.

PR #902, opened 2017-11-23
Settings page view design 

### LINKED ISSUE #835: Implement "Settings" view visual design
https://xd.adobe.com/view/15b5fb9d-ebde-4efc-919c-61dba3706e8c/
Visual design + specs

Input fields: #777

[changepasswordDefault.svg.zip](https://github.com/decred/decrediton/files/1412308/changepasswordDefault.svg.zip)

vctt94:
fixes #835


alexlyp:
So looks like you didn't rebase this quite right.  Not sure where there are changes from master included into this commit.

app/components/inputs/SettingsInput.js
```diff
@@ -0,0 +1,59 @@
...
+
+  render() {
+    const {
+      className, valueKey, labelKey, multi, clearable, options
+    } = this.props;
+
+    return (
+      <div className={className}>
+        <Select
+          clearable={clearable}
```
alexlyp: Duplicated props.  But I think you should just remove clearable and multi as options for the SettingsInput.  I can't see an situation that they would not always both be false for a settings dropdown.  If we did need to add a special case we would just create a specialized one off Input in that case.

alexlyp:
The change passphrase button needs hover styling.  Use this background color:

  background-color: #1b58ff;

PR #907, opened 2017-11-27
Refactor Control Actions

go1dfish:
Continuation of #528

Splits off rpc communication to code in wallet

### DIFF
--- a/app/wallet/service.js
+++ b/app/wallet/service.js
@@ -1,10 +1,6 @@
 import Promise from "promise";
 import * as client from "middleware/grpc/client";
-import {
-  NextAddressRequest,
-  DecodeRawTransactionRequest,
-  ValidateAddressRequest,
-} from "middleware/walletrpc/api_pb";
+import * as api from "middleware/walletrpc/api_pb";
 
 const promisify = fn => (...args) => new Promise((ok, fail) => fn(...args,
   (res, err) => err ? fail(err) : ok(res)));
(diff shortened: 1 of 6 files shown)


app/wallet/service.js
```diff
@@ -18,26 +14,29 @@ export const getDecodeService = promisify(client.getDecodeMessageService);
 
 export const getNextAddress = (walletService, accountNum) =>
   new Promise((resolve, reject) => {
-    const request = new NextAddressRequest();
+    const request = new api.NextAddressRequest();
     request.setAccount(accountNum);
     request.setKind(0);
+    request.setGapPolicy(api.NextAddressRequest.GapPolicy.GAP_POLICY_WRAP);
```
go1dfish: @alexlyp One thing to note about this PR is that for all cases of the client using getNextAddress it now sets an explicit gap policy.

I don't think this should be a problem, but wanted to call attention to this just in case it is.

alexlyp: Yah that should be fine, maybe we can let the user toggle that in the future, but I believe explicit setting to WRAP by default is the way to go.  We've really only seen some exchanges and other use cases need to disregard the policy or error, otherwise I prefer the WRAP.

alexlyp:
Hit a few issues during testing:

1) ChangePassphrase:  TypeError: Cannot read property 'PRIVATE' of undefined
2) Rescan:  Just like issues I saw with handling streams for transactionNotifications, looks like it's not working correctly here for rescanProgress. 
3) Constructing transaction: res.getEstimatedSignedSize is not a function
4) ImportScript: hitting the same issue from rescanProgress not being properly captured.

alexlyp:
@go1dfish any chance you can attend to these issues?  typically don't like to have PRs sit for so long

alexlyp:
Closing in favor of a rebased version.

Issue #880, opened 2017-11-15
Detect & allow user to change language on first run after install


matheusd:
User should be able to select the desired language after install, the first time Decrediton is run (before starting wallet creation). 

Ideally, the app should detect the OS language and suggest that, but allow the user to change. After the first run, the user can change the language in the settings page.

alexlyp:
Yah that's a great idea, I'm thinking we can do this while the blockchain is downloading?  I'm trying to think of other initial startup settings/options we could add during blockchain download, maybe timezone setting?

matheusd:
I guess timezone could be relevant (to see all timestamps in utc/local timezone/custom timezone). I'll try to think of additional options.

Maybe instead of cherry picking which configs to show, we could show the config page (or all relevant options in the config page that don't depend on having the wallet created).

PR #909, opened 2017-11-27
ui: Autodetect Language

matheusd:
This enables the app to detect the current user OS and use that on the first run after installing.

This is needed for users of languages other than english because wallet setup and seed information is one of the most critical steps for the app's usage.

Partial fix of #880.


app/i18n/translations/dev.json
```diff
@@ -214,10 +215,12 @@
   "notifications.type.send": "Saída",
   "notifications.type.ticket": "Tíck",
   "notifications.type.transfer": "Transferência",
+  "passphrase.info.header": "Private passphrase",
+  "passphrase.info.message": "This passphrase ensures that no funds may be spent from this wallet without it.  It is used to encrypt your private keys associated with your addresses.  These private keys are required to sign transactions that spend from the addresses.  Make sure to use a high-security passphrase and keep this passphrase in a secure place.   With your wallet file and this private passphrase an attacker would have full control over any funds associated with this wallet.",
```
oktapodia: I don't know if that's needed but you always have 2 spaces at the beginning of a sentence after a `.`

matheusd: The string on dev.json is copied from the source string (in this particular case, that string is in PassphraseInfoModal, last modified by ayp) so we'd need to check with him.

But really, this was probably just a typo... :)

PR #914, opened 2017-11-27
ui: change settings during startup

### LINKED ISSUE #880: Detect & allow user to change language on first run after install

matheusd:
This adds a page during startup to allow changing all non-wallet-dependent settings (language, unit, etc).

Fix #880


matheusd:
I've refactored the sidebar handling from the getStart page into the actions, as changing the current locale caused the sidebar to unmount/mount and show up at the wrong time.

I have also refactored the general settings (settings that don't depend on the wallet) into their own component to allow easier embedding. All settings added there in the future (eg, timezone) will automatically be accessible during startup.

Also, as commented on #880, I've added the button to go to the settings page during the block download phase (as that is the longest operation on a new wallet). That can be added to other places if needed.

Preview:
settings-startup mp4

alexlyp:
Looks good from a quick pass.  Though I'd remove the network settings toggle from there.  I think that needs to be rethought in general.  Possibly even selecting/confirming would cause decrediton to reload.

matheusd:
Ok, I'll remove it. On my machine (dev environment) just changing the network isn't sufficient when I want to switch between mainnet/testnet: I need to also edit {dcrwallet,dcrd,dcrctl}.conf. Don't know if that's happening in production as well.

PR #931, opened 2017-12-01
Fix issue with transaction details and new menu links

alexlyp:
With the changes to MenuLinks, transaction details crashes due to having no corresponding activeLink to use.  This mostly fixes that issue and changes the selected menu to Transactions, but it doesn't highlight the link, because it isn't the "active" route.

### DIFF
--- a/app/components/SideBar/MenuLinks/index.js
+++ b/app/components/SideBar/MenuLinks/index.js
@@ -24,7 +24,8 @@ class MenuLinks extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     const activeLink = getPage(nextProps.routes);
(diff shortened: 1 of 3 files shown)


alexlyp:
@Jascha-Sundaresan I'd like your input on fixing this further.  

I tried to added "transactions/details/:txhash" to be a subroute or "transactions"

`<Route path="transactions">
   ....
   <Route path="details/:txhash" noTab/>
</Route>`

This mostly worked with a filter in getTabs to avoid creating another tab for the details, but it would crash when actually going to the details page due to it not being able to resolve the active tab I believe.  I'm open for suggestions for the best way to fix this issue.

Jascha-Sundaresan:
I'll take a look after work this evening.

Issue #920, opened 2017-11-28
Add a new loading indicator


alexlyp:
In some places imagine a bit less invasive of a loading than the spinning DCR logo could be nice.  

I was thinking that we could just add another state to the button for loading.  Another color than enabled or disabled.  Also add some sort of basic animation to indicate that something is processing:  three blinking dots? small spinning DCR logo in the button?

linnutee:
This one is for replacing the large loader (rotating dcr symbol) during setup (and in any places currently within the app). 

Example + specs: https://xd.adobe.com/view/6efed3df-eadb-4b04-aa12-538c00ba2c59/

[decrediton - setup loader - alpha gif.zip](https://github.com/decred/decrediton/files/1545032/decrediton.-.setup.loader.-.alpha.gif.zip)

alexlyp:
Hrm not sure how I would add this gif in place of the css animated decred loading currently.  @matheusd any thoughts?

alexlyp:
https://github.com/decred/decrediton/pull/973

Here's a quick spin of it.  Still needs proper spacing but looks relatively smooth. Just wondering if using a GIF is good for performance etc.

matheusd:
I'll take a look at the PR in a moment. Performance wise, I don't think it will make any noticeable difference (for the case of a single animation playing).

Having it as an svg animation would be better from the pov of display quality and responsiveness but I believe we can roll with this one for now.

linnutee:
This one might be too tricky as an svg due to the 3d. Is there any other output format that's more optimal, spritesheet?

alexlyp:
https://github.com/decred/decrediton/blob/master/app/style/icons/stakeySprites.png

This is what I had Kyle make for the stakey bounce animation, I suppose that would be best for consistency.

alexlyp:
OK the hexagonal loader is in master

PR #936, opened 2017-12-04
Add infinite scroll on transaction list

### LINKED ISSUE #930: Infinite scrolling on transaction history
To be done after #875.

Implement infinite scrolling and other UI/UX improvements on the transaction history tab.

matheusd:
This implements infinite scrolling of the transaction list page using the https://github.com/CassetteRocks/react-infinite-scroller component.

I've used a small stakey bouncing as indicator during transaction loading. That indicator (or a more generic version) may be used as basis for #920.

Fix #930.

### DIFF
--- a/app/components/StakeyBounce.js
+++ b/app/components/StakeyBounce.js
@@ -1,6 +1,0 @@
-import React from "react";
-import "../style/Stakey.less";
-
-const StakeyBounce = () => <div className="stakey" />;
-
-export default StakeyBounce;
--- a/app/components/indicators/LoadingMoreTransactions.js
+++ b/app/components/indicators/LoadingMoreTransactions.js
@@ -1,0 +1,10 @@
+import StakeyBounceXs from "./StakeyBounceExtraSmall";
+import { FormattedMessage as T } from "react-intl";
+import "style/Loading.less";
+
+const LoadingMoreTransactionsIndicator = () =>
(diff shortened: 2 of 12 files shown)


alexlyp:
Nice really like the UX for this, huge improvement.

For the stakey loading div, I was thinking we could avoid showing unless it is taking longer than some amount of time, like 1 sec.  I only rarely see it appear at the bottom and typically it just flashes and goes away.  Not really sure what the proper way to handle this is.  Maybe put in a slight artificial delay?

matheusd:
Good idea. I can also increase the distance from the bottom where the loading is triggered (so you usually never reach the end of the list while scrolling)

alexlyp:
Ah yah I like that idea.  Something to give a little more feedback would be nice.

matheusd:
Feedback to what operation? The loading? Something in addition to the stakey indicator?

alexlyp:
Ah let me be a bit more clear. 

The way I see it goes 2 ways: either we give more feedback and slow down if the height is near the bottom or up the height and it becomes essentially transparent to the user that it is getting more transactions to show.  In the 2nd option, no feedback is required.

matheusd:
Right. Just pushed one increasing the height, as that is easiest. See if that is enough.

alexlyp:
Feels really good now.

PR #943, opened 2017-12-05
Consolidate passphrase confirmation modals

alexlyp:
This PR removes the need to do things such as "requestPassphrase" to show/hide modals in the parent components.  Instead, they are fully contained into the added PassphraseModalButton components.  

Now you can simply pass your passphrase modal form content and the onSubmit action that the content will perform on submission.  I believe the Content modals themselves could be reduced further and simply pass the passphrase form any additional fields you'd like to capture along with the standard passphrase (or even possibly a modal with no passphrase and only additional fields).

Passphrase confirmations that were updated:
- Send
- Purchase Tickets
- Revoke Tickets
- Import Script
- Add StakePool

Once we settle on this being solid change, we can apply the change to the remaining ones and ensure we are also using these passphrases confirmations for all private passphrase capturing.


app/components/modals/PassphraseModalContent/Modal.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import React from "react";
+import KeyBlueButton from "KeyBlueButton";
+import { FormattedMessage as T } from "react-intl";
+import { PasswordInput } from "inputs";
+import "style/Modals.less";
+
+const Modal =({
```
matheusd: As I mentioned on another file, I'd move (or insert by reference) this inside PassPhraseModal (below the current title/content), as all PassphraseModals are supposed to ask for the password.

alexlyp:
Ok confirmed Ticket Purchase and Import Script to work now.  Can you give them another try?

I'm still looking to further reduce the modal content usage into a clean HOC but can't seem to quite get it right.  But I believe this should be good enough for a first pass then we can refine further.

matheusd:
Btw, I've noticed a strange bug: if you import a script and then purchase a ticket, it tries to import the script again. This is also happening in master, but it's more noticeable on your branch (the snackbar message for import script success appears on your branch, probably due to some timing/ordering difference from master).

PR #963, opened 2017-12-07
Fix height of selects and inputs

### LINKED ISSUE #960: Destination input shifts on send to account click
The account select for the destination account shifts down along with the bottom of the div and constructed transaction information.

matheusd:
Fix #960 

Had to tweak the styling of other pages as the changes in the select affect them as well.


alexlyp:
Problem looks fixed.

Only issue I see is a slight vertical alignment issue for the purchase ticket account select/num tickets row.

PR #969, opened 2017-12-08
Add hide show divs motion on Account Page

vctt94:
Hide/show divs more gracefully when showing account details


alexlyp:
Feels pretty good.  

When hiding the details it looks like it catches about 3px below fully shut, then after 1-2s it disappears.

Also it needs to reduce height when going to rename account option.  If that is tough to do, I'm ok with just opening an issue and getting this in.

app/components/views/AccountsPage/Accounts/AccountRow/index.js
```diff
@@ -54,14 +58,89 @@ class AccountRow extends Component {
...
+          updateRenameAccountName,
+          renameAccount,
+          hideRenameAccount,
+          intl,
+          renameAccountNameError,
+        }}
+      />,
+      key: "output_0",
+      style: {
+        height: spring(280, {stiffness: 110, damping: 14}),
```
alexlyp: Changing this height to 140 fixes the Rename account issue.

PR #972, opened 2017-12-09
Adding hide/show div to Stake Info Details on Purchase Tickets tab

vctt94:


app/components/StakeInfo/index.js
```diff
@@ -13,6 +16,76 @@ class StakeInfo extends React.Component {
...
+        }}
+      />,
+      key: "output_0",
+      style: {
+        height: spring(180, {stiffness: 170, damping: 15}),
+      }
+    }];
+  }
+
+  getStakeInfoRowStyles() {
```
matheusd: I'm not a fan of this and the next one being called `*Styles` when they contain both style and data/elements.

vctt94: would it be better if it was stakeInfoRowComponent?

vctt94:
This still a WIP, I was trying to change the opacity for avoiding these hard flickers, but it seemed a bit tricky, but I can get a better effect without changing the opacity, I think so. I will finish to work on it today

app/components/StakeInfo/index.js
matheusd: I think so, yeah.

matheusd:
One thing you can try and see how it looks: instead of 2 columns on the expanded details, set it as 3 columns, with the first row == to the collapsed version. So that it will look like it's expanding, instead of switching + expanding.

alexlyp:
Agree about the columns.  Feel free to make any changes to make the change from short to expanded as smooth as possible.  I always felt like the styling of the StakeInfo area was lacking anyways.

vctt94:
I am keeping always the first line now but that is a row with just one data. I believe the effect looks good, but some opinion with the design is welcome

alexlyp:
Motion feels much better.  I'll probably touch up the expanded styling a bit, but this will be great to start.

One thing that I find jarring, is the scroll-y pop-in/out when expanding/hiding the area.  Now with our styled scrolls I think it's worth just doing "scroll-y: scroll" for tab-content so it will just enable/disable as needed instead of suddenly appearing.

vctt94:
I am not sure what you mean by that. Are you saying for adding "overflow-y: scroll" on tab-content layout.less?

alexlyp:
derp yes, that what I meant. sorry.

vctt94:
I did not notice difference between this and auto, do I did that right?

alexlyp:
When it's on auto for me, when you expand the area there is a slight (5px) bump of everything towards the left.

PR #983, opened 2017-12-12
Rescan cancel and final styling updates

### LINKED ISSUE #627: Improve UI for Rescan button
Please add an "Are you sure?" with description of function, or something similar, to the Rescan button in the wallet.

I'm a new user and just waited overnight for blockchain to load up. Then I did a test transfer into the wallet. Waited a moment, and then saw this juicy button labeled "Rescan". I thought it might scan for current transactions so I clicked it. Horrified that it started full blockchain scan again. Wish it had warned me or been clearer. Even just renaming button "Rescan Entire Blockchain" would be an improvement, though overly long.

alexlyp:
Fixes #627


alexlyp:
Ok removed other logging, but left the .on("error") one because I'd like to consume that message some how.

PR #988, opened 2017-12-14
Hide show motion Advanced purchase ticket

vctt94:


vctt94:
I am having an annoying bug, when typing on input, it loses its focus because the component is rendering again. I believe one way to fix is adding redux-form instead of regular inputs. But when I am trying to call the component with redux-form, it is entering in an infinite loop calling the form until the stack is full. Does anyone have passed by this? Any other ideas on how to solve this issue? I am open to discussions.

alexlyp:
OK i'll take a look

alexlyp:
Overall looks great.  Just noticed that on expanding the area the Purchase button (and insufficient funds error) overlap the advanced options very briefly.

alexlyp:
Still seeing it a bit, can you just do the same opacity/height motion that you've done for the other expand areas?  Also you should move PurchaseTickets/*Advanced and *ToolBar into the PurchaseTickets directory under the TicketsPage/PurchaseTab view

vctt94:
I did not figure a way to change opacity between two elements that have opacity 1. In the other cases I am hiding the div. On this case I am changing between components. But should be a way,I will look into it.

vctt94:
I believe the motion now looks better, what do you think?

alexlyp:
Nice, feels good.  I'll review and get this in soon.

PR #990, opened 2017-12-14
Clean up of components and imports

### LINKED ISSUE #809: move shared components to a shared directory w/ an index
```
import SlateGrayButton from "../../../SlateGrayButton";
import KeyBlueButton from "../../../KeyBlueButton";
import Balance from "../../../Balance";
```

becomes
```
import { SlateGrayButton, KeyBlueButton, Balance } from "shared";
```

alexlyp:
Try to move remaining loose components into correct folders.  Also clean up paths and references to imported components throughout.

Closes #809


matheusd:
Are you leaving `AgendaCard` on the root of components for later refactor? Or did you just miss that one?

alexlyp:
Was debating that and the other components floating in the `components/` directory.  Think they should all go in `shared/`?

matheusd:
AFAIK, the agenda card is currently only used on the governance, so I'd move it there. Same rationale for other stuff that is **not** being used in multiple places (move closer to wherever it's being used).

For small stuff being used in multiple places, yeah I believe we should move to "shared" (even though I'm not a fan of that name, we can roll with it for now).

matheusd:
Complemeting previous response: medium-sized stuff (like txHistory) that have their own dir with multiple smaller components can live directly in the components dir (don't need to move to shared)

alexlyp:
Gotcha I'll make those changes

alexlyp:
OK I think I'm all done with moves and changes

PR #992, opened 2017-12-15
Autofocus top field in passphrase modals

### LINKED ISSUE #823: Update Password entry interaction
When a password modal opens, password input field needs to be pre-activated for typing.

alexlyp:
This required changing Input from a stateless functional component to be able to use the DOM `ref` and `componentDidMount()`.

Close #823


alexlyp:
Made that change you noted above.

PR #1013, opened 2017-12-19
Add onKeyDown to inputs

matheusd:
Fixes bug of closing app when hitting enter on GetStarted pages.


matheusd:
When you test this, could you please test inserting a seed with duplicate seed words? Something like: `spaniel quantity spaniel...`. I'm getting a bug on my dev machine on duplicate words, but on a production version of 1.1.2 this isn't happening, so wanted to confirm if this is bugging out on someone else.

PR #1019, opened 2017-12-20
Hide show motion auto buyer ticket

### LINKED ISSUE #933: Add hide/show motion to other expandable areas
Take the motion that was implemented in #915 and apply it to the following areas:
- ~Account row click to show details~
- StakeInfo area expanding
- Purchase Ticket advanced
- Ticket buyer advanced settings

vctt94:
fixes #933 with PRs 969, 972 and 988


alexlyp:
Is there any way to have the page actively scroll down with the expanding bottom area?  And I feel like there needs to be margin between the end of the expanded area and the bottom of the screen.

vctt94:
I tried with scrollTo, but it seems not work on electron, If I did nothing wrong. There is also a scrollIntoView, but it still not support by most browsers, like chrome, so also not supported by electron.

vctt94:
scrollTop works on electron. I am taking the scroll to the bottom when it is showing details, but I think it is not so smooth, what do you think?

alexlyp:
Nice I think the scrollDown feels better.  Anyway you can ease the scroll down to match the div expansion?  Feel abrupt and jarring the way it is now.

vctt94:
I think now it looks better, but it has a bug that I need to see how to work around. If you are seeing the details and trigger some redux action, it scrolls to the bottom.

vctt94:
Ok, I think now it is good to go

PR #1021, opened 2017-12-20
Validate Address form

### LINKED ISSUE #978: Show Wallet Addresses or test if address is from wallet
Got a support message on slack, about a transaction not showing up on Decrediton.

I thought it would be helpful if we could either show all used wallet addresses **or** have a way to test if a given address is from the wallet. 

dcrctl has a `getaddressesbyaccount`, but AFAIK a listing like that wasn't implemented on Decrediton due to concerns of address reuse.

Maybe an alternative to listing addresses would be to have a way of inputing an address, and then the app can check this listing whether this address has been generated by it.

alexlyp:
Adds a new form in the Security Center that allows a user to verify that a given address is:

1) Owned by them
2) Valid address, but not controlled by the current wallet.
3) Invalid address, that is unable to be used on our network.

Currently have a rough GUI, but plan on adding transactions that are related to the address (if address owned and txs found).  Also should add an info modal to explain the results.

Closes #978


app/actions/ControlActions.js
```diff
@@ -654,3 +657,13 @@ export const validateAddress = address => async (dispatch, getState) => {
...
   }
 };
+
+export const validateWalletAddress = address => async (dispatch, getState) => {
+  const { network } = getState().grpc;
+  const validationErr = isValidAddress(address.address, network);
+  if (validationErr) { return { isValid: false, error: validationErr, getIsValid () { false; } }; }
+  dispatch({ type: VALIDATEADDRESS_ATTEMPT });
+  wallet.validateAddress(sel.walletService(getState()), address.address)
+    .then(response => { dispatch({ response, type: VALIDATEADDRESS_SUCCESS });})
```
matheusd: Minor nit. 

Seems to me the only difference between `validateAddress` and `validateWalletAddress` is the try...catch and the dispatch of VALIDATEADDRESS_SUCCESS. Maybe we could move the VALIDATEADDRESS_SUCCESSS dispatch into validateAddress and get rid of validateWalletAddress altogether? Or make validateWalletAddress use validateAddress and dispatch the VALIDATEADDRESS_SUCCESS in case of no errors.

This would be mainly to avoid the duplicate code...

alexlyp: Yah I did it sloppy on this first pass due to trying to mutlicase the existing one.  I'll fix before merging this PR.

alexlyp:
I'll be opening a new issue that will cover designer tweaks and what I'd like to see the validate address area become.

vctt94:
I think would be good to before unmount the component clean the state. So the owned message leaves.

vctt94:
I found a bug, when typing on To destination address input at transactions page on send tab, I am getting this error message: Uncaught TypeError: resp.getIsValid is not a function

alexlyp:
OK this is ready for another review/testing

vctt94:
Looks good to me. Great work

PR #1023, opened 2017-12-22
Add View Logs tab to Help page

alexlyp:
This is a rough attempt at allowing the user to quickly access their dcrd and dcrwallet logging from inside of decrediton.  This only captures and displays the logging from the current process of dcrd/dcrwallet (ie it's not loading the log files themselves).  

I'll open another issue to handle adding some better styling and also to add a check whether there are daemon logging to see. (ie you can't see remote daemon logging).


app/components/views/HelpPage/LogsTab/Page.js
```diff
@@ -0,0 +1,44 @@
...
+          </div>
+        </div>
+      }
+      {!dcrwalletLogs ?
+        <div className="log-area-hidden" onClick={showDcrwalletLogs}>
+          <T id="help.logs.show.dcrwallet" m="Show dcrwallet logs" />
+        </div>:
+        <div className="log-area-expanded">
+          <div className="log-area-expanded-hide"  onClick={hideDcrwalletLogs}>
+            <T id="help.logs.show.dcrwallet" m="Hide dcrwallet logs" />
```
matheusd: id should be `help.logs.hide.dcrwallet`

app/main.development.js
```diff
@@ -503,15 +521,23 @@ const launchDCRWallet = () => {
     }
   });
 
-  if (debug) {
-    dcrwallet.stdout.on("data", (data) => {
-      process.stdout.write(`${data}`);
-    });
+  dcrwallet.stdout.on("data", (data) => {
+    if (dcrwalletLogs) {
+      dcrwalletLogs = Buffer.concat([dcrwalletLogs, Buffer.from(data)]);
```
matheusd: Should we worry about adding a maximum size for the `dcrwalletLogs` buffer here? Otherwise this will just keep getting bigger and on a long lived wallet may start to cause problems...

matheusd: Also, a stylistic (optional) note, just cause I can't help it ;P

I'd initialize `dcrwalletLogs` as an empty buffer to avoid the `if` here, if that var is never nulled.

matheusd:
Also (don't know whether you wanna do on this PR or maybe another one), this needs to be available on the getstarted pages so we can instruct users to debug wallet initialization problems.

app/main.development.js
alexlyp: Agree on the max size, just not sure how best to implement.  If greater than max then pop the front of the current dcrwalletLogs buffer and then continue with concat of new data?

alexlyp:
Ok, added an arbitrary max size of log buffer as 50000, it'll pop the first line from the previous buffer before concatenating the incoming data to ensure it's < 50000 in length

app/main.development.js
```diff
@@ -481,15 +496,21 @@ const launchDCRWallet = () => {
     }
   });
 
-  if (debug) {
-    dcrwallet.stdout.on("data", (data) => {
-      process.stdout.write(`${data}`);
-    });
+  dcrwallet.stdout.on("data", (data) => {
+    while (dcrwalletLogs.length + Buffer.from(data).length > MAX_LOG_LENGTH) {
+      dcrwalletLogs = dcrwalletLogs.slice(dcrwalletLogs.indexOf(dcrwalletLogs.slice(dcrwalletLogs.length-1, dcrwalletLogs.length),1)+1);
```
matheusd: This seems a bit more complicated and slower than it needs to be. I think this can be improved in the folowing ways:

1. Code for handling `dcrwallet.stdout.on(data)` and `dcrwallet.stderr.on(data)` is very similar. When it was 1 line for each it was ok to be duplicated, but since this is now more elaborate, this could be refactored into a local parametrized closure to unify the code. Bonus points for doing it once for all daemons. Something like:

```
const AddToLog = (destIO, destLogBuffer, data) => {
  ...  
  debug && destIO.write(data)
  return Buffer.concat(destLogBuffer, ...);
};
...
dcrwallet.stdout.on("data", (data) => dcrWalletLogs = AddToLog(process.stdout, dcrWalletLogs, data))
dcrwallet.stderr.on("data", (data) => dcrWalletLogs = AddToLog(process.stderr, dcrWalletLogs, data))
```

2. Don't call `buffer.from(data)` repeatedly. Store it on a local var then use that.

3. Use `os.EOL` instead of the slice(lenght-1, length) as it makes instantly more clear what is happening here :wink: 

4. Instead of a loop to figure out how many lines to remove from the top, search for the first EOL after the amount of bytes that need to be removed (data.length) and slice from there until the end. You can then drop the loop.

5. Current impl has an infinite bug when data.length > MAX_LOG_LENGTH. Tested by reducing MAX_LOG_LENGTH to 1000 (or other small value), when closing the app it gets stuck in an infinite loop.

alexlyp:
OK, updated with all the suggested changes.  Much more elegant than mine :P

PR #1024, opened 2017-12-22
[wip] Paginated get tickets

### LINKED ISSUE #815: Pagination refactor/rewrite for GetTransactions/GetTickets

### LINKED ISSUE #950: (1.1.1) Visual Design: Tickets/My Tickets update
screen shot 2017-12-06 at 06 25 08

Not sure why this had been added on top of the My Tickets subpage. I would recommend removing it.
- It bears resemblance to the isolated ticket types, that was meant as an overview of the various elements and not to be included as a view. 

- My Tickets should display the ticket list when clicked on. The context menu is used to filter out different ticket types. 
https://xd.adobe.com/view/7707b586-62f0-4c93-afaf-1c5b9432b8eb/screen/5514ce96-bfb4-4eae-b37f-2c87e0d45d1b/My-Tickets

- The similar statistic about all Tickets lives under the Statistics subpage:
https://xd.adobe.com/view/7707b586-62f0-4c93-afaf-1c5b9432b8eb/screen/eabc4f54-628f-413b-b28d-563557eb0476/Custom-Size-6

matheusd:
Fix #950 
Fix #815 (tickets part)

Requires https://github.com/decred/dcrwallet/pull/1026

Don't mind the code yet. It's here for testing after the dcrwallet PR gets merged.


alexlyp:
Seems to work as expected, but performance is a bit wonky with about ~500 tickets on testnet.  Once I scroll down and load in ~20-30 tickets it starts to get a bit sticky when scrolling.

matheusd:
So, I had to use quite a lot of tricks to try and bring performance up, some of which I should refactor to send as a separate PR.

But if you have a couple of minutes, try again this updated version and see if it feels better for you.

alexlyp:
Ah much better. After scroll around for a while I don't notice any slow down and except for a little run up on CPU (decoding new transactions I think) looks good.

matheusd:
I'm gonna go ahead and close this for now because it got way out of sync to master.

While dcrwallet now sends the tickets in a streamable fashion, during startup we're loading the full set to calculate statistics anyway so the original purpose of this (remove the need to load all tickets at once during startup) is lost.

We can make another pass at this if/when we start loading stats directly from the wallet instead of calculating it ourselves.

PR #1031, opened 2017-12-28
sort out the `import` declare

isLishude:
I'm confused about `require("electron-store")` twice here,It's not necessary.By the way the code include Node require style and ES6 import style,It looks like not clean.


matheusd:
I'm getting an exception ("Store is not a Constructor") running `yarn dev`  with this PR applied. Could you double check if you're also getting that?

isLishude:
I see,I fixed it just now and tested on mine.It seems that Commonjs module uses `import m from "module"` is better than `import * as m from "module"` by Babel.

PR #1037, opened 2017-12-29
Adding hex seed import 

### LINKED ISSUE #847: Allow for seed hex input as well
There should be an option for a user to enter in the hex of the seed instead of the full 33 word seed.

brunobraga95:
fix #847

### DIFF
--- a/app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/Form.js
+++ b/app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/Form.js
@@ -50,8 +50,13 @@ class ConfirmSeedForm extends React.Component{
                 <T id="confirmSeed.wordsRemaining" m="{remaining, plural, one {one word remaining} other {# words remaining} }"
                   values={{remaining: remaining}} />
                 :
-                <T id="confirmSeed.hexSymbolsRemaining" m="{remaining, plural, one {one hex symbol: your key must have at least 32 symbols and 128 max} other {# hex symbols: your key must have at least 32 symbols and 128 max } }"
-                  values={{remaining: remaining}} />
+                <T id="confirmSeed.hexSymbolsRemaining"
+                  m="{remaining, plural, one {one hex symbol: your key must have at least {hexMin} symbols and {hexMax} max} other {# hex symbols: your key must have at least {hexMin} symbols and {hexMax} max } }"
+                  values={{
+                    remaining: remaining,
+                    hexMax: SEED_LENGTH.HEX_MAX,
+                    hexMin: SEED_LENGTH.HEX_MIN,
+                  }} />
               }
             </div>
           </div>
--- a/app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/Form.js
+++ b/app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/Form.js
@@ -1,15 +1,21 @@
 import SeedEntry from "./SeedEntry";
+import { TextToggle } from "buttons";
 import { FormattedMessage as T } from "react-intl";
 import "style/CreateWalletForm.less";
 import { InfoModalButton } from "buttons";
 import { SeedInfoModalContent } from "modals";
+import { SEED_LENGTH } from "wallet/seed";
+
+const getRemaining = (seedWords, seedType) =>
+  (seedType === "words" ? SEED_LENGTH.WORDS - (seedWords.length ? seedWords.split(" ") : []).length
+  : seedWords.length);
 
 class ConfirmSeedForm extends React.Component{
-
   constructor(props){
     super(props);
     this.state = {
-      showPasteWarning : false
+      showPasteWarning : false,
+      seedType: "words",
     };
   }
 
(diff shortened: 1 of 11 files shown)


matheusd:
Hello Bruno, thank you for your PR!

A few tips for your future contributions:
- Please rebase against latest `master` before submitting, otherwise we get merge conflicts
- When listing the issue(s) that will be closed, add them as `#<issue-numbe>` (see [here](https://help.github.com/articles/closing-issues-using-keywords/) for an explanation of the syntax)
- Run `yarn lint` and fix any problems listed

I'll add a review of the PR in a few minutes.

app/components/CreateWalletForm/ConfirmSeed/SeedEntry.js
```diff
@@ -91,7 +103,14 @@ class SeedEntry extends React.Component {
...
   }
-
 }
 
+SeedEntry.prototypes = {
+  seedType: PropTypes.string,
+};
+
+SeedEntry.defaultProps = {
+  seedType: "words",
```
matheusd: Since you're using the "words" and "hex" constants (the seed type) in multiple files, I'd define them as constants (`SEED_TYPE_WORDS = "...", SEED_TYPE_HEX = ...`) somwhere and import them.

app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/SeedEntry.js
```diff
@@ -16,20 +18,40 @@ class SeedEntry extends React.Component {
...
-
+    const { seedType } = this.props;
+    if(seedType === "hex") {
+      return (
+        <Input
+          onChange={this.onChange}
+          value={this.state.currentHex}
+          onPaste={this.props.onPaste}
+          name='hexInput'
+          placeholder={formatMessage(messages.enterSeedPlaceholder)}
```
alexlyp: Please make a new placeholder for the Hex Entry like: "Enter your seed in hex..." or something like that.

app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/Form.js
```diff
@@ -20,31 +26,55 @@ class ConfirmSeedForm extends React.Component{
...
+                modalTitle={<h1><T id="confirmSeed.seedInformation" m="Seed information" /></h1>    }
+                modalContent={<SeedInfoModalContent />}
+              />
+            </div>
+            <div className="confirm-seed-label-remaining-words">
+              {seedType === "words" ?
+                <T id="confirmSeed.wordsRemaining" m="{remaining, plural, one {one word remaining} other {# words remaining} }"
+                  values={{remaining: remaining}} />
+                :
+                <T id="confirmSeed.hexSymbolsRemaining" m="{remaining, plural, one {one hex symbol: your key must have at least 32 symbols and 128 max} other {# hex symbols: your key must have at least 32 symbols and 128 max } }"
```
alexlyp: Can you explain the 32 min 128 max?  Afaik it will always be 32 bytes (64 chars when written out)

brunobraga95: I thought so too, but while coding, depending on the seed I was entering I notice the following `errors` or `warning` (that I am pretty certain were coming from the backend `decodeSeed` method)

- Warning 1:  Would be shown in case a no `HEX` symbol was entered, which makes sense, so it only accepts 0-F
- Warning 2: Would appear if the seed has an odd length
- Warning 3: Would say that the seed must have between 128 and 512 bytes, therefore a minimum of 32 and a max of 128 symbols

I will post the exactly messages as soon as I can, but that is pretty much does reason.

am I missing something?

app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/SeedEntry.js
brunobraga95: absolutely.

app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/Form.js
alexlyp: Yah those warnings are coming from dcrd/hdkeychain:https://github.com/decred/dcrd/blob/master/hdkeychain/extendedkey.go#L457-L459

So you "can" make a wallet with a seed that is between 16 and 128 bytes, but we only provide 32 byte seeds from all of our software.  I suppose we should allow people to use any length between 16 and 128.   Let me discuss with the other guys.

alexlyp: And could you also add input validation to only accept 0-F for the hex?

alexlyp: Also in terms of input validation, probably best to not send to wallet to decode the seed until it is > 32 chars and not odd.

alexlyp: OK so confirmed that we're OK with > 16 and < 128.  But maybe add a warning on seeds that aren't 32 bytes that these come from non-supported software and may have unintended consequences.

brunobraga95: should we allow the same thing for the `words` input type?

alexlyp: No, leave the words input as is.

matheusd:
Can you make it so it doesn't break the layout flow when "hex seed" is selected?
screenshot_2018-01-05_09-02-30

PR #1039, opened 2017-12-31
adding a desktop file to make it easy to launch decrediton on Linux

druonysus:


matheusd:
Thank you for the PR!

Did you test both the .deb and .rpm installers? Is this `.destkop` supposed to get automatically installed? Couldn't find the electron/electron-builder API that specifies where the .desktop should live on the source  tree.

druonysus:
you could package this .desktop file in a Linux package, be it .deb or .rpm. This is a [freedesktop.org standard](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html)

druonysus:
the desktop software on my openSUSE machines have .dekstop files that get installed in `/usr/share/applications/`... I think this the same for other distributions too, but I haven't verified this.

PR #1059, opened 2018-01-08
Maintain selected stakepool state and remove stakepool

### LINKED ISSUE #1058: Decrediton resets selected stake pool when multiple pools are configured
Decrediton version 1.1.3

Repro:
1) Have multiple stakepools configured. 
2) In Tickets tab, choose non-first pool to be used to buytickets 
3) Navigating to a different tab than Tickets (i.e. Overview) 
4) Navigate back to Tickets tab
5) Observe that the selected pool is different than the pool chosen in step 1. The selected pool appears to be reverted to the first pool on the list.

Also, Decredition does not allow to remove a stake pool configuration, so a workaround of removing all pools except one is not possible.

matheusd:
Fix #1058 

This adds the following changes:
- Move selected stakepool into global state so it is maintained between tab/page changes
- Add a remove stakepool button into the the stakepool list
- Add a DangerButton to be used in remove operations
- Add ConfirmModal to be used in operations that need confirmation but not necessarily the passphrase

Only thing I'm not 100% sure is if I need to dispatch a rescan after removing a stakepool from the configured ones.

### DIFF
--- a/app/actions/StakePoolActions.js
+++ b/app/actions/StakePoolActions.js
@@ -24,8 +24,9 @@ const updateSavedConfig = (newPoolInfo, poolHost, apiKey, accountNum) =>
         : config);
     if (!stakePoolConfigs.find((conf, idx) => conf !== currentStakePoolConfig[idx])) return;
     getCfg().set("stakepools", stakePoolConfigs);
+    let selectedStakePool = stakePoolConfigs.filter(p => p.Host === poolHost)[0] || null;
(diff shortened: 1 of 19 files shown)


alexlyp:
No rescan necessary for removing a stakepool account.  I suppose if we had the ability to "removescript" in dcrwallet we'd need to do that, but once someone importscript it's essentially permenant to that wallet.

PR #1064, opened 2018-01-09
Update electron to 1.7.10

### LINKED ISSUE #1051: Update to latest electron
We're currently at 1.4.15, which is over a year old at this point.

Updating the package and dependencies itself is easy. The tricky bit is to test on all platforms to make sure there are no regressions, so we should schedule to apply this update at the start of a new release cycle to have enough time to test everything. Major things that should be tested:

- Our custom IPC node module on win32/64
- Startup and shutdown (simple and advanced mode)
- Logging

matheusd:
Fix #1051 

I tried to update to the 1.8.x release line, but got some undebugable white error screen (electron didn't load the app, didn't throw an exception or gave me anything to indicate where it had errored), so I pulled back to the latest release of the 1.7.x (non-beta) line.

I tested the general functions (seed input, transactions, ticket purchase, simple and advanced mode) on the following platforms:

- Linux (development version with `yarn dev`)
- Linux (locally packaged version with `yarn package`)
- Windows (development version with `yarn dev`)
- Windows (locally packaged version with `yarn package`)

Don't have a macOS computer to test, so could use some help there if available.

I had to delete the `node_modules` and `app/node_modules` and execute `yarn` to rebuild the correct electron/electron-builder packages.

Also, on windows I had to manually install the npm package `windows-build-tools` as an admin, but I don't remember having to do that before (see https://github.com/felixrieseberg/windows-build-tools/issues/56#issuecomment-304404123).

I'm getting an exception on shutdown regarding the grpc module. It's a read error, probably due to trying to execute a service after the wallet process has shutdown, or due to one of the streams getting closed (my guess is the block/transaction notification). I only got this during shutdown, so I guess it can be safely ignored for now, but I can try and better pinpoint the origin if needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ bin/
 .vagrant
 dist/
 app/i18n/extracted/app
+/*.node
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -110,6 +110,9 @@ var cfg = initCfg();
 
 const logger = createLogger(debug);
 logger.log("info", "Using config/data from:" + app.getPath("userData"));
+logger.log("info", "Chrome version: " + process.versions.chrome);
+logger.log("info", "Electron version: " + process.versions.chrome);
+logger.log("info", "Decrediton version: " + app.getVersion());
 
 var createDcrdConf, createDcrwalletConf, createDcrctlConf = false;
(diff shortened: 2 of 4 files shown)


alexlyp:
tested fine on deb9, will try to test on macOS today or tomorrow.

PR #1068, opened 2018-01-09
Add log viewing to GetStarted stage

### LINKED ISSUE #1049: View logs on GetStarted stage
We should be able to view the logs not only after the wallet is loaded on the help screen, but also on the GetStarted pages (during daemon blockchain sync, wallet header download, etc) as that may be useful to debug initialization problems.

matheusd:
Fix #1049 

Besides during blockchain downloading, I also added the log button on daemon loading (before starting the blockchain download) as that seems to be the step most people get stuck.

Let me know if you think it should be added to more places.


alexlyp:
Couple of nits:
- Can you float the Logs button to the left or right?  Seems weird center aligned.
- Can you hide "Show dcrwallet logs" since dcrwallet won't have been started at this point

PR #1073, opened 2018-01-10
Add first pass action logging in WalletLoader

alexlyp:
First swing at adding logging to decrediton actions.

The thought is to log each action attempt as "info", successes as "debug" and errors as such.  I'm open to input to try and find a nice equilibrium that isn't too spammy.

Once we settle on general outline for these, I will apply the same to the rest of the actions.  After that we can debate how much logging we want above in the components, but I'm thinking we shouldn't really have any logging up above (unless some error situation).

### DIFF
--- a/app/actions/WalletLoaderActions.js
+++ b/app/actions/WalletLoaderActions.js
@@ -21,13 +21,18 @@ export const LOADER_SUCCESS = "LOADER_SUCCESS";
 
 export const loaderRequest = ( address, port) => (dispatch) => {
   const request = { address, port };
+  wallet.log("debug", "Wallet loader request at: " + address +":" + port);
   dispatch({ request, type: LOADER_ATTEMPT });
   return getLoader(request)
     .then(loader => {
       dispatch({ loader, type: LOADER_SUCCESS });
       dispatch(walletExistRequest());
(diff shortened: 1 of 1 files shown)


matheusd:
Agree on the general outline (log requests as info, successes as debug, errors as errors).

Agree on taking this as a starting point, see how spammy it gets and then adjusting as needed.

My first instinct would be to do exactly the way you did (going over individual functions and adding logging as needed) but I wanted to propose a different method of adding the logging for comparison and discussion: #1075 

The main difference is using a higher order function to log and implementing logging inside the wallet "package" so the work to manually add logging to each function is somewhat reduced. We can mix and match both approaches, improve the `withLog` to log individual parameters or create different versions that log/don't log the `..args`.

I think we should get @go1dfish opinion on this.

alexlyp:
I definitely prefer your version of this.  Much, much cleaner.  Then I suppose if there are any special case  logging situations we can just put those in as needed `wallet.log(..)`

matheusd:
Yeah, either special case there, create a different version of `withLog()` that does something different or don't use `withLog()` at all and manually call the log function.

alexlyp:
Closing in favor of the other impl from @matheusd

PR #1075, opened 2018-01-11
[poc] Logging with higher order function

matheusd:
Alternative to #1073

### DIFF
--- a/app/wallet/app.js
+++ b/app/wallet/app.js
@@ -1,5 +1,5 @@
 import { ipcRenderer } from "electron";
-import { isObject, isString, isNumber, isNullOrUndefined, isUndefined, isNull } from "util";
+import { isObject, isString, isNumber, isUndefined, isNull } from "util";
 import { isFunction } from "util";
 
 export const onAppReloadRequested = cb => ipcRenderer.on("app-reload-requested", cb);
@@ -7,6 +7,8 @@ export const onAppReloadRequested = cb => ipcRenderer.on("app-reload-requested",
 export const log = (level, ...args) => {
   ipcRenderer.send("main-log", ...[level, ...args]);
 };
+
+export const logOptionNoArgs = (opts) => ({...opts, noArguments: true});
 
 // Formats a dynamic list of log arguments
 const formatLogArgs = (msg, args) => {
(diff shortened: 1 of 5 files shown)


app/wallet/app.js
```diff
@@ -5,3 +5,20 @@ export const onAppReloadRequested = cb => ipcRenderer.on("app-reload-requested",
...
 };
+
+// Higher Order Function that wraps the promise-generating function f with log
+// calls. Calling f() **must** return a promise.
+export const withLog = (msg, f) => (...args) => {
+  log("info", "Wallet requesting %s", msg);
+  return new Promise((resolve, reject) => {
+    f(...args)
+      .then((...res) => {
+        log("debug", "Wallet received response %s", ...res);
```
alexlyp: Can we add the msg here too, so we can easily see what response this corresponds.

alexlyp: Also on most of my responses it didn't resolve the `%s` and just showed it at `Wallet received response %s`

When it did resolve it showed [Object object].

I'm guessing this is due to most of these loader requests not having any `resolve(response)`.

matheusd:
Ok, improved the logging of the arguments and response values.

Also, I changed the order of `withLog()` to pass the function first and message/options after that, so that the function signature is immediately visible on the source code. See if you prefer this version, versus passing the log message first.

I also added a variant to not log arguments (to be used on calls like signMessage which pass the private passphrase). Probably will also need a variant to not log responses (to be used on the seed functions). I added logging on signMessage as a test.

app/wallet/loader.js
```diff
@@ -24,44 +25,44 @@ export const getWalletExists = (loader) =>
...
+  }), "Create Wallet", logOptionNoArgs());
 
-export const openWallet = (loader, pubPass) =>
+export const openWallet = log((loader, pubPass) =>
   new Promise((resolve, reject) => {
     const request = new OpenWalletRequest();
     request.setPublicPassphrase(new Uint8Array(Buffer.from(pubPass)));
     loader.openWallet(request, error => error ? reject(error) : resolve());
-  });
+  }), "Open Wallet");
```
alexlyp: can you log this with no args as well?  it's logging the attempted public pass being attempted.

matheusd:
Want me to add logging to other stuff in this PR?

alexlyp:
I think this is a good start, you can go ahead and apply this to the rest of the actions if you like now.

PR #1080, opened 2018-01-12
More transaction notification refactoring

alexlyp:
In lieu of us forcing the daemon to be synced before proceeding (and upcoming SPV), I've removed all syncness checking in the transaction notification data processing.

The hope after merging this PR is to decouple new block connecting with updating all balances.  We should only have to update balances for a given account when a transaction from is observed to have been received from that account.  We can also investigate moving the getStakeInfo out from new block connects.  Roughly the same rule should be applied and only call that when a new ticket/revoke/vote is observed to have been received.


matheusd:
I just tried running this branch, but app startup gets stuck in the "Checking Wallet State" stage. Is your version working ok?

alexlyp:
Hrm actually I'm seeing that too now, I'll see what's up with it

alexlyp:
OK should be fixed now, accidentally removed grpcpport setting for wallet.

PR #1081, opened 2018-01-13
Clean up account individual updating

### LINKED ISSUE #574: Decoupling requests on successful block connect
By issuing requests when a new block notifiication hits, instances where wallet gets tons of new block connected notifications  (after a restart after a while or some other down time) and will lead to a lot of unnecessary requests being issued.

alexlyp:
Fixes #574 

Now instead of blindly requesting information about all accounts, we only update the account referred to in the transaction or account notification streams. 

I believe more work can be done on streamlining the storing of account information, plus I'm not positive that it's currently Immutable.  

Seems to be making extraneous account update calls, but otherwise appears to be working as expected.

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -19,7 +19,7 @@ export const GETWALLETSERVICE_SUCCESS = "GETWALLETSERVICE_SUCCESS";
 function getWalletServiceSuccess(walletService) {
   return (dispatch, getState) => {
     dispatch({ walletService, type: GETWALLETSERVICE_SUCCESS });
-    setTimeout(() => { dispatch(getAccountsAttempt()); }, 10);
+    setTimeout(() => { dispatch(getAccountsAttempt(true)); }, 10);
     setTimeout(() => { dispatch(getMostRecentTransactions()); }, 20);
     setTimeout(() => { dispatch(getTicketsInfoAttempt()); }, 20);
     setTimeout(() => { dispatch(loadActiveDataFiltersAttempt()); }, 1000);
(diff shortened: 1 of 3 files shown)


app/actions/ClientActions.js
```diff
@@ -71,49 +71,81 @@ export const getTicketBuyerServiceAttempt = () => (dispatch, getState) => {
...
+    }
+  });
+  updatedBalance.total = getBalanceResponse.getTotal();
+  updatedBalance.spendable = getBalanceResponse.getSpendable();
+  updatedBalance.immatureReward = getBalanceResponse.getImmatureReward();
+  updatedBalance.immatureStakeGeneration = getBalanceResponse.getImmatureStakeGeneration();
+  updatedBalance.lockedByTickets = getBalanceResponse.getLockedByTickets();
+  updatedBalance.votingAuthority = getBalanceResponse.getVotingAuthority();
+
+  console.log("get balance update", accountNumber, updatedBalance);
```
matheusd: Remove this before final merge.

app/actions/ClientActions.js
```diff
@@ -362,11 +419,23 @@ export const getTransactions = () => async (dispatch, getState) => {
 
 export const NEW_TRANSACTIONS_RECEIVED = "NEW_TRANSACTIONS_RECEIVED";
 
+function checkAccountsToUpdate(txs, accountsToUpdate) {
+  txs.forEach(tx => {
+    tx.tx.getCreditsList().forEach(credit => {if (accountsToUpdate.find(eq(credit.getAccount()))) accountsToUpdate.push(credit.getAccount());});
```
matheusd: You missed a `!` at the `{if (accountsToUpdate.find(...`. Updating the accounts balance for receiving txs is broken without it.

alexlyp: ah nice catch

app/actions/ClientActions.js
```diff
@@ -71,49 +71,81 @@ export const getTicketBuyerServiceAttempt = () => (dispatch, getState) => {
...
+          votingAuthority: resp.getVotingAuthority(),
+        };
+        balances.push(accountEntry);
+      })
+      .catch(error => {
+        dispatch({ error, type: GETBALANCE_FAILED });
+        return;
+      });
+  });
+  dispatch({balances, type: GETBALANCE_SUCCESS });
```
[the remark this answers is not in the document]
alexlyp: Yah I think I meant to make new ones, but seems like reusing makes sense here.

PR #1086, opened 2018-01-16
Fixes for startup crashing

alexlyp:
NeededBlocks was not being checked whether it was greater than 0, so it was bypassing the daemon syncing check.  This also exposed an incorrect component being imported for ShowErrors on the StartRPC screen.

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -81,7 +81,7 @@ export const syncDaemon = () =>
       return wallet
         .getBlockCount(credentials, appData)
         .then(updateCurrentBlockCount => {
-          if (neededBlocks > 0 && updateCurrentBlockCount >= neededBlocks) {
+          if ((neededBlocks == 0 && updateCurrentBlockCount > 0) || updateCurrentBlockCount >= neededBlocks) {
             dispatch({type: DAEMONSYNCED});
             dispatch({currentBlockHeight: updateCurrentBlockCount, type: STARTUPBLOCK});
             setMustOpenForm(false);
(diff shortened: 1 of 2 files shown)


matheusd:
You had removed the `neededBlocks` estimation on a previous PR and so it is only determined by the API.

I'm now wondering what happens if the api is down or the user has that dns blocked/proxied/whatever. Wouldn't that prevent the wallet from ever being synced?

alexlyp:
Yah I believe it would jam.  Not really sure how to best handle that situation.

matheusd:
I think the best way to solve would be to get the needed block height from whatever the connected dcrd instance believes is the current best block, but I don't think there's a way to get that information.

Only problem is with SPV we don't have that information (though I'm unsure if that will even be needed with SPV, so no need to worry too much about that).

Other alternative would be to return the estimation code for the case when neededBlocks === 0. The problem with **that** is that since it's always underestimated, the user can start using the wallet before dcrd has finished catching up (but that was the old behaviour anyway).

alexlyp:
Ya, I think it's best to not revert the hacky neededBlocks estimation.  

Yah I think for now if neededBlocks remains 0 we should just let the user through as long as currentBlockHeight of dcrd is > 0

app/actions/DaemonActions.js
```diff
@@ -81,7 +81,8 @@ export const syncDaemon = () =>
       return wallet
         .getBlockCount(credentials, appData)
         .then(updateCurrentBlockCount => {
-          if (updateCurrentBlockCount >= neededBlocks) {
+          if ((neededBlocks == 0 && updateCurrentBlockCount > 0) || (neededBlocks != 0 && updateCurrentBlockCount >= neededBlocks)) {
+            console.log(updateCurrentBlockCount);
```
matheusd: Log slipped through.

PR #1087, opened 2018-01-16
Fix startup actions call procedures

alexlyp:
To avoid certain calls being called multiple times, this cleans up the startup actions.  

When there are no new headers, then there is no need to rescan the chain.  Then these calls are explicitly done.  Otherwise if a rescan is required we do the same startup action calls (stakeinfo, gettransactions, ticketinfo, account balances) on that success as well.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -80,6 +80,7 @@ export function rescanAttempt(beginHeight) {
     });
     rescanCall.on("end", function() {
       dispatch({ type: RESCAN_COMPLETE });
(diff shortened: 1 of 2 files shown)


app/actions/ControlActions.js
```diff
@@ -80,6 +80,7 @@ export function rescanAttempt(beginHeight) {
     });
     rescanCall.on("end", function() {
       dispatch({ type: RESCAN_COMPLETE });
+      setTimeout(() => { dispatch(getStakeInfoAttempt()); }, 1000);
```
matheusd: Slight identation mismatch on this line

PR #1092, opened 2018-01-18
Change to use dcrdata instead of insight

### LINKED ISSUE #1091: Change tx/block links to use explorer.dcrdata.org instead of mainnet.

alexlyp:
Fix #1091


matheusd:
Doesn't seem to be working for mainnet.

The `network == "testnet" && "explorer"` outputs `false`, so the URL gets broken.

chappjc:
Oh, the /explorer part of the path was removed on master, which is deployed for mainnet, but not testnet.  We've coded redirects so that old /explorer links should work still

Issue #1084, opened 2018-01-16
Ticket Purchase just not Occurring


chadi77:
Whenever I attempt to purchase a ticket I successfully input everything, click the Purchase button, input the correct passphrase, get the "Successfully imported new script tickets" message, and Stakey just sits on the screen dancing and will not go away. No tickets show as purchased, no transactions appear on my transactions page (although a loading mor transactions animation appears but never goes away), and I never see any evidence of a ticket anywhere. I have attempted this several time over a few weeks with no success. I have to close Decrediton in order to be able to view the Purchase Tickets page and no transactions from my ticket purchase attempts appear. Rescanning doesn't do anything either. I have attempted to review the wallet logs but I am unable to to determine anything from them (because I am unsure how to read them.) I don;t know if this is an issue with Decrediton or with how I initially set it up.

Let me know what you may need me to do to help you figure out what is going on.

Thanks!

matheusd:
Which version of Decrediton are you running? There was a bug on 1.1.2, fixed on 1.1.3 that lead to that problem on some situations, so make sure you've upgraded (check the version on the Help -> About main menu).

chadi77:
I am currently on Decrediton 1.1.3, dcrd 1.1.2+release. and dcrwallet 1.1.2+release.

matheusd:
Ok, please try the following:

- Close Decrediton. Open it again.
- Main Menu -> "Advanced" -> "Toggle Developer Tools"
- Try the ticket purchase

See if there are any error messages there.

Second check:

- Main Menu -> "Advanced" -> "Show wallet log files"
- Open `dcrwallet.log`. Look from the bottom up (last lines) see if there are any messages marked as `[ERR]`, specially after a line starting with `GRPC: Unary Method /walletrpc.WalletService/PurchaseTickets invoked ...`

chadi77:
OK yeah seeing a multitude of errors in Developer Tools. 
dcrwallet.log does not contain anything with "PurchaseTickets".

Errors from Developer Tools:
> Uncaught AssertionError: Assertion failed  bundle.js:14

* goog.asserts.DEFAULT_ERROR_HANDLER | @ | bundle.js:14

* goog.asserts.doAssertFailure_ | @ | bundle.js:14

* goog.asserts.assert | @ | bundle.js:14

* jspb.BinaryEncoder.writeSignedVarint64 | @ | bundle.js:14

* jspb.BinaryWriter.writeSignedVarint64_ | @ | bundle.js:14

[...]

Issue #1042, opened 2018-01-03
Improve decrediton logging


alexlyp:
With the addition @matheusd's logging improvements in https://github.com/decred/decrediton/pull/1032 we should begin adding logging to the internal react side of decrediton.  

Off the top of my head we could begin with the following:
- All errors
- Any call out to dcrwallet (possibly log response receipt too)
- Changes to settings 
- Changes to config.json

Feel free to comment to add any additional logging that you believe will be useful to expose to the user.

PR #1101, opened 2018-01-23
Add logging to all wallet methods

matheusd:
Part of #1042 

This adds logging to all wallet actions. I tried to track down and omit private information leaking (such as balance amounts, transaction data, etc) but it should be double checked to see if I missed any.

I also added logging to the `purchaseTicketAttempt`, due to the recent #1084 issue, but before adding more logging we should see if the action refactors (#907 & future PRs) will continue, as it is easier to add logging to the normalized wallet functions rather then on each react action.

Also didn't add logging to config yet as there is work being done on that to refactor it for multiple wallet support.

### DIFF
--- a/app/wallet/client.js
+++ b/app/wallet/client.js
@@ -5,7 +5,7 @@ import {
   BalanceRequest, TicketPriceRequest, StakeInfoRequest,
   AgendasRequest, VoteChoicesRequest, SetVoteChoicesRequest, GetTicketsRequest,
 } from "middleware/walletrpc/api_pb";
-import { withLog as log, logOptionNoResponseData, withLogNoData } from "./app";
+import { withLog as log, logOptionNoResponseData, withLogNoData, withLogNoResponseData } from "./app";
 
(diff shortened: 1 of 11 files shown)


app/wallet/client.js
```diff
@@ -5,33 +5,36 @@ import {
...
 export const doPing = promisifyReq("ping", PingRequest);
 
-export const getBalance = (walletService, accountNum, requiredConfs) => new Promise((ok, fail) => {
+export const getBalance = withLogNoData((walletService, accountNum, requiredConfs) => new Promise((ok, fail) => {
   const request = new BalanceRequest();
   request.setAccountNumber(accountNum);
   request.setRequiredConfirmations(requiredConfs);
   walletService.balance(request, (err, res) => err ? fail(err) : ok(res));
-});
+}), "Get Balance");
```
alexlyp: Can you add account number for the Get Balance request logging?

alexlyp:
Needs a conflict resolution too.

PR #1109, opened 2018-01-25
Alert on seed copy and allow seed paste

matheusd:
This changes the behavior of seed copy/pasting to the following:

- On a copy attempt of the seed, display a warning message and show a confirmation modal to allow the copy (screenshot to follow)
- On a paste attempt of a **new** seed (i.e: user creating a new wallet, clicked "continue" to go to the seed confirmation stage, tried to paste the seed) paste is disallowed and a message is displayed
- On paste attempt of **existing** seed, paste is allowed but a reminder message about backing up the seed is displayed 

Things up for discussion for this PR:

- Maintain seed copy with confirmation or disable seed copy entirely? Reasoning for maintaining seed copy with a confirmation: some users **do** understand they're doing when trying to copy.
- Use a confirmation button instead of a phrase
- Also allow pasting on new seed
- Block pasting on new seed only if detected an earlier confirmed copy


matheusd:
screenshot_2018-01-25_09-00-36

alexlyp:
Very cool!  Definitely like the difference and am a big fan of the warning screen.  

Any way we can alert on seed selection too?  That is the way I typically "copy" on Linux

Agree, with the no pasting for the confirmation but allowing it for the existing creation. 
 
I like the writing out of "I understand the risk" instead of a button, gives it a bit more gravity.

matheusd:
I'm showing the alert after receiving the `onCopy` event, so I'd imagine it already supports copy-on-select.

Did you try it on your setup and it copied the seed without the confirmation modal? I'll try and see how to config my OS to behave like that and test.

alexlyp:
Yah appears selecting isn't throwing the onCopy event.  From the docs looks like it won't trigger with onCopy with selection.  I'll play around a bit to see if I can figure out a solution.

alexlyp:
So from everything I'm reading onSelect should work, but it's just not firing for me.   

Maybe instead just not allow user-select in the css then add a button to copy which shows the alert?

matheusd:
Ok, after trying some different things, it seems the intra-app clipboard on linux doesn't trigger any events.

I switched the seed to be unselectable, and added a small copy-to-clipboard icon at the bottom-right corner of the seed.

alexlyp:
Nice works great, I'll give it a review here shortly.

PR #1112, opened 2018-01-26
[poc] Update balance on maturing coin blocks

### LINKED ISSUE #1085: Proper account balance updating for maturing coins
With the new method of only updating account balances when related transactions are seen, there is a situation that needs to be handled.

When there are votes/revokes/coinbases that are currently maturing for a user, there is no trigger to update the account balance upon successful maturity.  User could just restart to fix the issue, but there should be a more graceful way to handle.

Rough idea for fix:  
- When new mined votes/revokes/coinbases have been notified OR look for votes/revokes/coinbases on startup within 256 blocks of current block height
-  Create a array in state to track block heights that balances should be requeried.  
-  On new block connected, check height against this array, dispatch account balance update if necessary.

matheusd:
Fix #1085 

I left some instrumentation for now, to help test/debug. Will remove it and the [poc] title after review.

This is basically following the proposal in the issue (tracking the blocks where coins are maturing) with a slight change to also track the related accounts (so only the accounts related to the maturing coins are updated).

Transactions are checked on startup, rescan and after new transactions are mined.

I'm checking both for votes/revokes coinbase maturing and ticket expiry (though maybe I should ignore that?)

Also, I've created a new constants file in the wallet dir, so we should later on move all "magic" numbers spread throughout the app into this single file for better organization. It is following the same convention as the dcrd go code (one chainParam object for each network).

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -80,18 +80,27 @@ export const MATURINGHEIGHTS_ADDED = "MATURINGHEIGHTS_ADDED";
 // Given a list of transactions, returns the maturing heights of all
 // stake txs in the list.
 function transactionsMaturingHeights(txs, chainParams) {
-  let res = [];
+  let res = {};
+  const addToRes = (height, found) => {
+    const accounts = res[height] || [];
+    found.forEach(a => accounts.indexOf(a) === -1 ? accounts.push(a) : null);
+    res[height] = accounts;
+  };
+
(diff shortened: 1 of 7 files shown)


alexlyp:
Looks good to me.  I like the approach

matheusd:
Let me know if when you believe it's good to go, and I'll remove the instrumentation and the [poc] tag.

alexlyp:
Yah go for it

PR #1117, opened 2018-01-29
Switch DcrInput to atoms

### LINKED ISSUE #1110: Transaction was sent missing one atom
I tried to send a transaction of 1.05635 DCR, but only 1.05634999 DCR were sent.

tx: https://mainnet.decred.org/tx/85c6ef53def0e8dcdd9113b9af4785669401de8f29bd982754d507322908f8a4

matheusd:
Fix #1110 

This changes the `DcrInput` component to manage the input amount in atoms instead of as string, implementing rounding for floating values following the same procedure as dcrd/dcrwallet (https://github.com/decred/dcrd/blob/v1.1.2/dcrutil/amount.go#L77).

It also adds `FloatInput`, `IntegerInput` (behave as expected) and `FixedDcrInput` (which behaves as the old `DcrInput`, meant for inputs that are *always* assumed to be in DCR amounts).

### DIFF
--- a/app/components/inputs/DcrInput.js
+++ b/app/components/inputs/DcrInput.js
@@ -1,7 +1,67 @@
+import FloatInput from "./FloatInput";
 import NumericInput from "./NumericInput";
+import { strToDcrAtoms } from "helpers/strings";
 import balanceConnector from "connectors/balance";
 
-const DcrInput = ({...props, currencyDisplay}) =>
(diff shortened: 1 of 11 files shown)


alexlyp:
Nice, good refactor too.

Noticed 1 thing during testing though, set to Atoms display and then fill in some large amount in the Amount field (after populating the Address field).  Giving me an AssertionError

matheusd:
Got it. Added a guard against input > max coin supply.

PR #1131, opened 2018-01-31
[getstarted] Load wallet before daemon

alexlyp:
Due to the upcoming onboarding/get started changes, it is required that wallet is loaded before the daemon is attempted to load.  

This offers a much more intuitive onboarding flow, so users can do all of the wallet creation prior to spinning up the daemon.  In a future PR we can parallelize the process as well, so the daemon can be syncing in the background while the user creates the wallet.

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -1,4 +1,4 @@
-import { versionCheckAction } from "./WalletLoaderActions";
+import { versionCheckAction, startRpcRequestFunc } from "./WalletLoaderActions";
 import { stopNotifcations } from "./NotificationActions";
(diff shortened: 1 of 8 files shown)


matheusd:
Awesome! Tested a bunch of combinations of standard/advanced daemon, with existing/new wallets and fully synced/partially synced chain and  didn't hit any snags.

Only nit is the layout for the advanced daemon form is kinda weird with the big top loading indicator. Do you plan on changing this layout on this PR or will just leave it like that and work on it during the onboarding changes?

screenshot_2018-02-02_08-47-31

alexlyp:
Oh weird that should have been fixed, let me look into that again

PR #1135, opened 2018-02-01
Added System Browsing for File/Directory

### LINKED ISSUE #850: Add a FileInput/DirectoryInput to use a file-picker
Specifically to be used for the Advanced startup form, but may be used elsewhere.

cipherzzz:
Resolves #850


cipherzzz:
See the directory home selector and rpc cert file selector in the advanced startup forms

alexlyp:
A few linting issues in PathInput.

Overall, great stuff.  Works just as I had expected it to. 

Mostly concerned with the references to electron windows/dialogs from within react. @matheusd any thoughts about that?  I don't believe there is security issue there, but want to make sure now that React is communicating to the layers above that we aren't bleeding any information.

cipherzzz:
Yeah, I think the only way to select a directory is by using the native ipc electron process... Only way I saw, anyways. I was using prettier to format my js, but it appears to not work with the editorconfig stuff? @alexlyp

alexlyp:
Mmmm should be using the `.eslintrs.js` rules

cipherzzz:
Let me know if you need anything else from me on this... I installed the ESLint extension in VSCode so I'll at least get warnings now... Giving up on prettier for now

app/components/inputs/PathInput.js
```diff
@@ -0,0 +1,56 @@
...
+class PathInput extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { path: "" };
+    this.selectDirectory = this.selectDirectory.bind(this);
+  }
+
+  componentDidMount() {
+    const self = this;
+    ipc.on("path", function (event, data) {
```
matheusd: This won't work here, since this listener will be bound forever. Try the following and see the problem:
- Start in advanced daemon mode
- Select a Daemon Data Dir
- Go to the other "tab" and select a cert file

You should see an exception due to setState below being called on an unmounted component.

One option to solve this, would be to move the whole file/directory selection logic into either an action or (probably better) a function that returns a promise. Then this promise gets resolved when the selection is done or rejected if canceled/errored and you use a `.once` listener so that only *that* particular call is triggered.

app/style/MiscComponents.less
```diff
@@ -239,6 +239,34 @@
...
+  cursor: text;
+  background: transparent;
+  outline: none;
+  align-items: center;
+  border-bottom: 1px @stroke-color-default solid;
+  font-family: @font-family-regular;
+  font-size: 13px;
+  color: white;
+  padding: 3px 4px 1px 10px;
+  width: 400px;
```
matheusd: Don't use a fixed width on the input here, let it fill the container component, as we may want to use it in different contexts with different widths.

If you want to show a component with the input+button (and if that's why you used the fixed width) then define a new component with both sub-components and style them such that the input leaves room for the button on the side.

Something to the effect of:

```jsx
const PathInputAndButton = (...) => (
  <div class="class-for-inline-pathinput+button">
    <PathInput ... />
    <PathInputButton .../>
  </div>
);
```

app/components/buttons/PathButton.js
```diff
@@ -0,0 +1,17 @@
+import { FormattedMessage as T } from "react-intl";
+import { Tooltip } from "shared";
+
+const dataDirMessage = "Select a path";
+
+const PathButton = ({ disabled, onClick }) => (
+
+  <Tooltip text={<T id="startup.dataDir.tip" m={dataDirMessage} />}>
```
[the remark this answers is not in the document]
cipherzzz: this was a relic of a refactor. I was originally showing a different tooltip message based on the type(file or directory). I'll fix it

app/components/inputs/PathInput.js
```diff
@@ -0,0 +1,56 @@
...
+const mainWindow = electron.remote.getCurrentWindow();
+const ipc = electron.ipcRenderer;
+
+import { PathButton } from "../buttons";
+
+class PathInput extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { path: "" };
+    this.selectDirectory = this.selectDirectory.bind(this);
```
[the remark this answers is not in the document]
cipherzzz: you got it

app/components/inputs/PathInput.js
cipherzzz: I was originally trying to create a utility component that didn't need to be tied to redux state, but creating an action is the right thing - thanks. I'm from react-native land and the ipc main and renderer process are new to me. I'm guessing it's like a system thread and a ui thread

app/style/MiscComponents.less
cipherzzz: once again, from the react-native land. I would have used a flex on the parent and divided it with the children. sass is a little new to me

app/components/inputs/PathInput.js
matheusd: You can still do it without going through redux by using a promise. Something like:
```javascript
const selectDir = () => new Promise((reject, resolve)) => {
  //...
  ipcRenderer.once('path'), (path) => resolve(path));
  dialog.showDirDialog(...);
  //...
}
```

Then use on the component by: 
```javascript
onSelectDir() {
  selectDir().then((path) => this.setState(...));
}
```

cipherzzz: @matheusd - I took a different approach with hopefully similar results. I am passing in an 'id' prop and using this id as a unique ipc event identifier. I am removing the listener upon the dismounting of the component to keep it tidy. I tested for multiple file/dir browsing components on the same screen and they seem to be functional. Let me know if you have more feedback upon review.

app/components/inputs/PathBrowseInput.js
```diff
@@ -0,0 +1,64 @@
...
+
+  render() {
+    return (
+      <div className={"path-input-and-button"}>
+        <PathInput
+          value={this.props.value}
+          onChange={(e) => {
+            this.props.onChange(e.target.value);
+            this.setState({ path: e.target.value });
+          }}
```
matheusd: This isn't wrong, but just to maintain a general stylistic convention, could you move all this code into an `onChange()` method for this class? 

On the rest of the project, if the event more than a simple, single method call or assignment we usually write a separate handler for it (so the `render()` method is simpler).

app/components/inputs/PathInput.js
```diff
@@ -0,0 +1,18 @@
+class PathInput extends React.Component {
```
matheusd: Is it possible for PathInput be just a functional component over the existing `Input` (or maybe `TextInput`)? What exactly exists on PathInput that doesn't in Input/TextInput?

app/components/inputs/PathBrowseInput.js
```diff
@@ -0,0 +1,64 @@
...
+
+  componentDidMount() {
+
+    let self = this;
+    pathListener = function (event, data) {
+      self.setState({ path: data });
+      self.props.onChange(data);
+    };
+
+    ipc.on(this.props.id, pathListener);
```
matheusd: Instead of requiring that the user of `PathBrowseInput` pass an id prop, you can just generate a random one (generate a random number or a random string) and use it internally (since this has no effect for the parent component).

app/components/inputs/PathBrowseInput.js
```diff
@@ -0,0 +1,64 @@
...
+class PathBrowseInput extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = { path: "" };
+  }
+
+  componentDidMount() {
+
+    let self = this;
```
matheusd: I'm pretty sure `@autobind` already covers this as well, so you don't need to store the reference in `self`. Try it out and see if it works.

app/components/inputs/PathInput.js
cipherzzz: I tried to use TextInput, but I was not able to override the flex properties by passing in a className - it insisted on taking up the whole container. I didn't feel comfortable modifying the Input class and breaking other stuff ;)

app/components/inputs/PathBrowseInput.js
cipherzzz:
```
PathBrowseInput.js:22 Uncaught TypeError: this.setState is not a function
    at EventEmitter.pathListener (PathBrowseInput.js:22)
    at emitTwo (events.js:106)
    at EventEmitter.emit (events.js:194)
```
I'll keep it how it is

matheusd: ok

app/components/inputs/PathInput.js
matheusd: ok.

app/components/inputs/PathBrowseInput.js
cipherzzz: good idea

app/components/inputs/PathBrowseInput.js
cipherzzz: got it

app/components/inputs/PathInput.js
cipherzzz: Maybe take a look at the Input class and try overriding the flex  properties if you get a second and I'll modify my code if you can get it working...

cipherzzz:
Not sure why the CI build is failing...
```
ERROR in bundle.js from UglifyJs
Unexpected token: name (win) [bundle.js:86818,6]
```

app/components/inputs/PathInput.js
matheusd: I believe we can leave that for a separate PR (unifying/improving the css of the existing input and the pathInput) as it will need a review over all input instances, so you can leave it the way you implemented for the moment.

app/components/inputs/PathBrowseInput.js
```diff
@@ -0,0 +1,68 @@
+const electron = require("electron");
+const dialog = electron.remote.dialog;
+const mainWindow = electron.remote.getCurrentWindow();
+const ipc = electron.ipcRenderer;
+
+import PathInput from "./PathInput";
+import { PathButton } from "../buttons";
+
+var pathListener;
+var key;
```
matheusd: Don't use global variables, as adding two components at the same time will make the behaviour undefined (imagine the second component ovewriting the `key`/`listener` variable)

cipherzzz:
@matheusd @alexlyp - thanks for your feedback! Glad to help with the project

alexlyp:
Can you rebase please?

PR #1136, opened 2018-02-01
Export Transactions as CSV

### LINKED ISSUE #912: Add transactional data export
Add a menu-bar item that allows users to export their transactional data over a given period of time and put it nicely into a csv for use elsewhere.

As a first pass this should be able to export information about tickets and votes that can be used for tax filing purposes.

Please update this issue with information that you'd like to see.

matheusd:
Fix #912 

POC for the moment. No fancy ui yet, only the export function on the "transactions" tab. Will create a local "test.csv" file.

### DIFF
--- a/app/actions/StatisticsActions.js
+++ b/app/actions/StatisticsActions.js
@@ -4,6 +4,10 @@ import fs from "fs";
 import { isNumber, isNullOrUndefined, isUndefined } from "util";
 import { tsToDate } from "helpers";
 
+export const EXPORT_STARTED = "EXPORT_STARTED";
+export const EXPORT_COMPLETED = "EXPORT_COMPLETED";
+export const EXPORT_ERROR = "EXPORT_ERROR";
+
 export const exportStatToCSV = (opts) => (dispatch) => {
   // TODO: get from dialog
   const csvFilename = "test.csv";
@@ -29,7 +33,7 @@ export const exportStatToCSVFile = (opts) => (dispatch) => {
 
   // called once at the start of the stats calc function
   const startFunction = (opts) => {
-    console.log("starting", opts);
+    dispatch({type: EXPORT_STARTED});
     allSeries = opts.series;
     const seriesNames = allSeries.map(s => s.name);
     const headerLine = csvLine(["time", ...seriesNames]);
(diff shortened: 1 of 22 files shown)


alexlyp:
Great stuff!  Looks really good. 
I'm trying to figure out what we should do with addresses.  I go back and forth whether those are relevant for this type of data dump, or is merely having the overall transaction in/out is enough.

matheusd:
I think for a csv export, the addresses aren't important (as there is likely to be multiple, you'd have to also signal which are from the wallet, external vs internal, etc). We could write a json export if someone wants to process more stuff (but then, they could just use dcrctl anyway).

matheusd:
I'll wait for the resolution of #1135 so I can include its file picker to select the destination file.

matheusd:
Ready for final review. Added the following:

- Daily balances stats and export (hopefully to be used as source for  the overview graphs)
- FIle input browser for the destination csv file

I had to slightly tweak the PathInputBrowser component to allow using a save dialog as well as the open dialog that @cipherzzz implemented. Also added the ability to use filters to it (so that the extension is appended by default when creating a file).

matheusd:
Had to change the transaction gathering logic due to having to decode the full vote/revoke transactions to account for the correct commitment amounts in various situations.

alexlyp:
Nice, I beileve the csv's properly match up now.  I'll do a review then should be good to go.

app/actions/StatisticsActions.js
```diff
@@ -0,0 +1,289 @@
...
+    dispatch({filename: csvFilename, type: EXPORT_COMPLETED});
+    if (fd) {
+      fs.closeSync(fd);
+      fd = null;
+    }
+  };
+
+  dispatch(calcFunction({opts, startFunction, progressFunction, endFunction, errorFunction}));
+
+  // if (fd !== null) {
```
alexlyp: looks like this can be removed

app/actions/StatisticsActions.js
```diff
@@ -0,0 +1,289 @@
...
+      {name: "direction"},
+      {name: "fee", type: VALUE_TYPE_ATOMAMOUNT},
+      {name: "amount", type: VALUE_TYPE_ATOMAMOUNT},
+      {name: "credits", type: VALUE_TYPE_ATOMAMOUNT},
+      {name: "debits", type: VALUE_TYPE_ATOMAMOUNT},
+    ],
+  });
+
+  const formatTx = (tx) => {
+    // console.log(tx);
```
alexlyp: remove

app/actions/StatisticsActions.js
```diff
@@ -0,0 +1,289 @@
...
+      direction: tx.direction,
+      fee: tx.fee,
+      amount: tx.amount,
+      credits: tx.tx.getCreditsList().reduce((s, c) => s + c.getAmount(), 0),
+      debits: tx.tx.getDebitsList().reduce((s, d) => s + d.getPreviousAmount(), 0),
+    };
+  };
+
+  const txDataCb = (mined) => {
+    //console.log(mined);
```
alexlyp: another

PR #1140, opened 2018-02-02
Adding charts components

vctt94:
Adding charts components with fake data on overview page.

Still need to synchronize with dcrwallet when data are ready to be streamed.


app/selectors.js
```diff
@@ -249,6 +250,49 @@ export const transactions = createSelector(
...
+
+export const spendableAndLockedByDay = createSelector(
+  [transactions],
+  (transactions) => {
+    let valuesByDate = {};
+    let spendableTotal = 0;
+    let lockedTotal = 0;
+    for(let i=transactions.length-1; i>=0; i--) {
+      let transaction = transactions[i];
+      const time = tsToYYYYMMDD(transaction.txTimestamp);
```
matheusd: Instead of using a fixed format, for displaying purposes I'd use a function to convert the timestamp to `Date.toLocaleDateString()` as that is the format for the user's OS.

Eventually we could move to use moment.js as it has better formatting capabilities, but that's not needed right now.

app/components/views/HomePage/index.js
```diff
@@ -61,6 +74,23 @@ class Home extends React.Component{
...
+  formatBalanceData(spendableAndLockedByDay){
+    const keys = Object.keys(spendableAndLockedByDay);
+    let data = [];
+    let newData = {};
+    for(let i=0;i< keys.length; i++){
+      const date = keys[i];
+      const values = spendableAndLockedByDay[date];
+      newData = {
+        name: date,
+        locked: values.lockedTotal/100000000,
```
matheusd: Don't use the fixed `/10000000` here. Use the selector `unitDivisor`, as it already performs the correct scaling depending on the user's settings.

vctt94:
Thanks for reviewing my PR.

I changed the data for fake data, as @alexlyp pointed out on slack.

alexlyp:
Can you rebase please?

alexlyp:
Can you update the PR title and message briefly explaining what you have done here?  Otherwise looks good to go.  Thanks!

PR #1141, opened 2018-02-04
Updated eslint and resolved new lint errors

StabbarN:
Updated eslint packages and resolved issues found by the updated eslint. See commit-wise to follow the changes more easily.


matheusd:
So, am I right in understanding that there are new linting rules in the new versions? As updating the eslint packages are now requiring fixes throughout the app?

StabbarN:
Yep, that’s correct

matheusd:
This is a big commit, and even though with a quick, very high-level overview seems fine and the changes *seem* (and that's the operative word here) to be only cosmetical, we'd prefer to have smaller changesets for review.

I think it's fine to update the packages, and the small commits like the ordering of the spread operator are also ok in themselves, but making eslint rule changes, even if it's an implicit rule that got changed upstream, should be done rule-by-rule, in individual commits, so we can make sure no extraneous code is getting lumped.

The eslint package upgrade should happen at some point, *but* the rules should remain stable across the upgrade and then we can discuss, implement and review the individual rule changes separately to lower the review burden.

StabbarN:
Sure, I'll split it into several PRs.

alexlyp:
I'd prefer if these changes happen after the next release.  Feel free to submit PRs, but I'll probably let them sit until the start of the next cycle

PR #1147, opened 2018-02-05
dockerize

dajohi:


Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+LABEL description="Decrediton builder image"
+LABEL version="1.2"
+LABEL maintainer "john@netpurgatory.com"
+
+USER build
+WORKDIR /home/$USER
+
+COPY ./ /home/$USER
+
+RUN yarn install
```
alexlyp: don't need to put install here, plus you need to `RUN yarn lint` and `RUN yarn test`

PR #1148, opened 2018-02-06
Add development binary linux target

matheusd:
This adds a yarn script to generate binaries for linux, without generating the .rpm and .deb packages.

I'm mainly using this to generate the development binaries for periodic testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ install:
 script:
   - yarn lint
   - yarn test
-  - yarn package-linux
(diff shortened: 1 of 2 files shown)


alexlyp:
Can you update .travis.yml to use `yarn package-dev-linux` as well?

alexlyp:
And I think you need to rebase as well

alexlyp:
Nice looks good, while you're in there can you take the double quotes off of the app/node_modules in travis.yml too?

PR #1149, opened 2018-02-06
Add call to rebroadcast unmined transactions

### LINKED ISSUE #1105: Implement rebroadcast of unmined transactions
https://github.com/decred/dcrwallet/pull/1033 implemented a rebroadcast api call. It should be ported to Decrediton.

matheusd:
Fix #1105 

I added the rebroadcast button on the send page for now, shown only when there are unmined transactions, but I'm not 100% certain it should remain there. Maybe in the future we should add a "troubleshooting" page with it and other support actions we generally instruct users to do.

To test this:
- run dcrd with a `--connect=1.2.3.4:5678` arg, so it's not on the network
- run decrediton and connect to the dcrd instance that is not on the network (you may need to change the url in [this line](https://github.com/decred/decrediton/blob/master/app/actions/WalletLoaderActions.js#L281) to something else so the app isn't waiting for the blocks)
- wait a block (to verify that the daemon is **not** really connected to the network; check the block explorer)
- send a new transaction
- wait a block (check on the block explorer) 
- verify that the unmined tx has **not** been published (check on the block explorer)
- close & reopen everything, now with the dcrd instance connecting to the network (remove the `--connect` arg)
- wait a block (to verify that the unmined tx has **not** been published)
- click the rebroadcast button 
- verify that the unmined tx **has** been published (check the block explorer)
- wait a block
- verify that the tx **has** been mined


alexlyp:
Nice tested great as explained.  Needs a rebase then is good to go.

PR #1157, opened 2018-02-08
Fixing overview

vctt94:
Fixing some details linnutee pointed at slack.

- Adding txHistory with no status (if not pending) at overview page
- Returning no ticket information on home recent transactions
- Fixing some chart style


matheusd:
needs a rebase

alexlyp:
lint-fix plz

alexlyp:
Too much whitespace around the bottom and left of the graphs.

The recent transactions row does not match the design doc from @linnutee

vctt94:
I believe now the transactions row are right.

Also, I found a bug, as we are using recentTransactions on selector, if we have many last transactions of only one kind (like only transactions or only tickets) the card does not show properly.
Here is an image: 
bugex

I changed recentTransactions for transactions and looks like it works, but I am not sure if this is the right approach.

alexlyp:
Yah I'm going to open an issue and discussion with @matheusd on how to fix that problem.  Basically it needs to populate recentTickets a bit differently to make sure it has something to put there.

alexlyp:
Can you cut the AM/PM and make it into 24H format for the time?

vctt94:
Yeah, sure.

Should I change back for recentTransactions in here https://github.com/decred/decrediton/pull/1157/files#diff-a7298267586e8979aeb195bb3079d9d2R267 as well until we have a solution?

alexlyp:
No that will goof up other things, I'll look into the fix for that today

alexlyp:
I'm still seeing the recent transaction's dates not have enough room and then it causes them to expand the height of the row then causes the transactions to overflow off the bottom of the screen.  And the rows are no longer clickable to see tx details.

alexlyp:
Also still seeing extra whitespace around the charts

vctt94:
I removed the bottom white space. but I am not seeing this overflow, so I added a little padding on bottom so it fixes. Does it? 

Still need to remove the commas from the date. I am working on it.

PR #1158, opened 2018-02-08
Populate Recent Tickets for Overview

alexlyp:
This is a first pass at properly displaying stake transactions on the overview.  

We are currently reusing TxHistory for these as well, but we should probably do a new TxHistoryRow based on TicketCards from the MyTicketsTab.  That change will also require usage of the normalizedTickets.

### DIFF
--- a/app/selectors.js
+++ b/app/selectors.js
@@ -242,6 +242,9 @@ export const transactions = createSelector(
 export const homeHistoryTransactions = createSelector(
   [transactionsNormalizer, get(["grpc", "recentTransactions"])], apply
 );
+export const homeHistoryTickets = createSelector(
+  [ticketNormalizer, get(["grpc", "tickets"])], apply
+);
 
 //fake data for balance chart
 export const spendableAndLockedBalance = createSelector(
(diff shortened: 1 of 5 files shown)


app/selectors.js
```diff
@@ -239,8 +239,17 @@ export const hasUnminedTransactions = compose(l => l && l.length > 0, get(["grpc
 export const transactions = createSelector(
   [transactionsNormalizer, get(["grpc", "transactions"])], apply
 );
+
 export const homeHistoryTransactions = createSelector(
-  [transactionsNormalizer, get(["grpc", "recentTransactions"])], apply
+  [transactions],
```
matheusd: If you use `transactions` instead of `get(["grpc", "recentTransactions"])`, changes to the current transaction filter will change the transactions shown in the overview page. Is that what you want to accomplish or are you going to (after this PR) change the behaviour of the transactions selector? If not, then you need to return it (the `grpc.recentTransactions` state is only filled during startup and after a successful rescan).

app/selectors.js
```diff
@@ -239,8 +239,17 @@ export const hasUnminedTransactions = compose(l => l && l.length > 0, get(["grpc
...
+  (transactions) =>
+    transactions.map(tx => {if (!tx.txType || tx.txType == "Regular") return tx; }).filter(tx => tx !== undefined)
+);
+
+export const homeHistoryTickets = createSelector(
+  [transactions],
```
matheusd: Same thing applies as previous comment. Either need to use `grpc.recentTransactions` or more likely (if we want to always fill with the 5 most recent stake txs) we'll need a separate state to track the home tickets.

matheusd:
I'm getting an error on the selectors now... are you getting that as well?

PR #1163, opened 2018-02-12
Support dynamic dcrwallet port

### LINKED ISSUE #852: Avoid TCP port conflicts with other decrediton instances
A couple recent wallet changes made it possible for decrediton or other wallet guis which wrap a dcrwallet process to run them without causing conflicting ports when multiple instances of the wallet are run:

* https://github.com/decred/dcrwallet/commit/92108ddf3a5d93e78fb26b27e4b44229c9c69f0b
* https://github.com/decred/dcrwallet/commit/b4c31ee1e67e35b16fd3f6cce7f68a998285058e

By setting the --txpipe and --rpclistenerevents flags, it is possible to use RPC listen addresses which use port 0.  This will cause the operating system to choose an unused port, and then wallet will then notify these listener addresses (including the chosen port) over the pipe using the custom IPC protocol.

A similar change could be made to dcrd if we wanted to support starting multiple instances of it, but this will require more discussion first since running multiple full nodes on the same machine is probably not desirable, and can be worked around with the new settings to choose an already-running dcrd to connect to.

matheusd:
Fix #852 

This opens up the possibility of running multiple wallet instances at the same time (not yet implemented).

On non-windows platforms, the `--rpclistenerevents --pipetx` dcrwallet arguments are used to pipe the events from the wallet daemon into decrediton. The grpc port is found when the appropriate IPC message is received.

On windows, due to various node/golang incompatibilities it's not possible to do that, so for the moment we resort to parsing the stdout output, looking for the appropriate log string. This is an ugly hack, so hopefully, we can have a better IPC mechanism on windows in the future.

Tested on linux and windows (needs testing on macOS).

To test this on an existing configured wallet, you'll need to edit the corresponding `dcrwallet.conf` (usually `.config/decrediton/wallets/testnet/default-wallet/dcrwallet.conf` or similar), modifying the grpc listen port to 0, eg: `grpclisten=127.0.0.1:0`).

Creating a new wallet and importing from an old install should automatically set the grpc port as 0.

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -21,7 +21,6 @@ export const AVAILABLE_WALLETS = "AVAILABLE_WALLETS";
 export const SHUTDOWN_REQUESTED = "SHUTDOWN_REQUESTED";
 export const SET_CREDENTIALS_APPDATA_ERROR = "SET_CREDENTIALS_APPDATA_ERROR";
 
-
 export const WALLETCREATED = "WALLETCREATED";
 export const WALLET_AUTOBUYER_SETTINGS = "WALLET_AUTOBUYER_SETTINGS";
 export const WALLET_STAKEPOOL_SETTINGS = "WALLET_STAKEPOOL_SETTINGS";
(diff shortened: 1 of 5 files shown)


StabbarN:
On latest macOS I tested setting `/Users/.../Library/Application Support/decrediton/wallets/mainnet/default-wallet/dcrwallet.conf` `grpclisten=127.0.0.1:0` and then started decrediton with an existing wallet. I don't know where to look but it seems to be working.

matheusd:
If the wallet started, it means it worked :smile: 

To verify that a dynamic port has been selected, look at the debug log on your console. It should show something like "[DCRW] gRPC server listening at 127.0.0.1:[random-port]". And a few lines down, "[DCTN] dcrwallet grpc port detected at [same-random-port]".

Issue #1121, opened 2018-01-30
Design Implementation: Launcher Create/Restore Wallet


linnutee:
Visual Design and specs: https://xd.adobe.com/spec/8a5e5ecc-9d0c-4e01-9328-e7d12d1ac652/

When a new user completes or skips the onboarding, is directed to `Create or Restore Wallet view`. 

The seed key entry has been re-worked. Instead of a single large input field, each seed word now has an individual field. 

Seed word-fields interactions are as following:
- Disabled state for displaying the seeds (cannot be interacted with)
- Empty active state (No data entered. When typing into the active field, dropdown gives a suggestion based on the typed letters (as it currently does).)
- Filled active state (Data entered and can be edited)
- Filled active state with error (User has entered a seed word that isn't part of the PGP word list or has a grammar mistake in it. User has not entered a word, but clicked continue – also triggers notification.)
Nb! Error state does not trigger when user enters a `wrong seed word`, in that case only the notification is displayed.

`X` directs the user either to the previous view (when already creating the wallet) or to the `Launcher Overview `(during create or restore view)

`Public Password`: help icon opens a modal with the information, will need to provide design input for this. 

`Confirmation`: when a new wallet is succesfully created, user is directed to the confirmation view for feedback. Also blockchain should automatically start downloading then. Download process is displayed in the bottom of the viewport. Clicking continue leads to the `Launcher Overview`.

Icons: [decrediton-walletcreation.zip](https://github.com/decred/decrediton/files/1675695/decrediton-walletcreation.zip)

See [launcher-master](https://xd.adobe.com/spec/ab537a4e-2902-4260-99f1-85c3992f9a73/) more for reference or hit me up on slack.

linnutee:
Minor update: removed the Continue button for consistency with Overview. Should just work with clicking on `Create` or `Restore`. 
![screen shot 2018-02-01 at 02 07 14](https://user-images.githubusercontent.com/17774057/35654202-a1130c00-06f4-11e8-803d-4b10293777e5.png)

Also will need to add another feature to `Restore` views "Restore from File". That will likely be a third tab.

xaur:
Does it mean user needs to TAB between 33 fields, instead of pressing space?

alexlyp:
not sure what you mean @xaur

xaur:
Sorry, I thought this:

>  Instead of a single large input field, each seed word now has an individual field.

[...]

PR #1165, opened 2018-02-13
[launcher] New seed confirmation

alexlyp:
This prepares the seed confirmation functionality to be in line with what the design document would like (as described in #1121).  Now instead of re-doing the full 33 word seed, the user is presented with the seed with roughly 10 words removed (it uses a basic PRNG to compare against 0.3, so it may vary).  The user then needs to populate each of the missing words to complete the wallet creation process.


alexlyp:
Right so this PR was more about updating the functionality of the confirmation instead of a true replacement.  I'll be doing the styling changes in a future PR.  I believe both issues you note are acceptable in the design doc in #1121

app/components/views/GetStartedPage/OpenWallet/CreateWalletForm/ConfirmSeed/Form.js
```diff
@@ -54,21 +28,12 @@ class ConfirmSeedForm extends React.Component{
...
-    }
     return errors;
   }
 
   render(){
-    const { isMatch, isEmpty, setSeedWords, seedWords, createWalletExisting } = this.props;
-    const { seedType } = this.state;
-    const remaining = getRemaining(seedWords, seedType);
-    const errors = this.mountSeedErrors();
+    console.log("rerender", this.props.seedWords);
```
matheusd: remove this

app/style/ReactSelectGlobal.less
```diff
@@ -54,21 +55,21 @@
 }
 
 .Select-control:hover:not(:disabled) {
-  box-shadow: 0 1px 0 rgba(0, 0, 0, 0.06);
-  color: @input-color-hovered;
-  border-bottom: 1px solid @stroke-color-hovered;
+  //box-shadow: 0 1px 0 rgba(0, 0, 0, 0.06);
```
matheusd: Unless you'll return this on the redesign refactor, I think you can remove the commented css.

PR #1167, opened 2018-02-13
Add ticketbuyer modal text

### LINKED ISSUE #1166: [autobuyer]  Add more explicit text for autobuying
There should be text added to the confirmation modal of the start auto buyer switch so that people know more clearly that they need to leave decrediton running for it to automatically purhase tickets

alexlyp:
Fixes #1166


matheusd:
I think we should also include a warning letting users know that the wallet remains unlocked while the autobuyer is running, which means that the users funds may be withdrawn by anyone with access to the computer.

PR #1168, opened 2018-02-14
[launcher] Update confirm modals to match design doc

alexlyp:
Also add a new "Invisible Button" type that will be used in place of the SlateGrayButton in most places

### DIFF
--- a/app/style/main.less
+++ b/app/style/main.less
@@ -80,13 +80,6 @@
   color: @error-text-color;
 }
 
-.right {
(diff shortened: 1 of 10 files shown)


app/style/main.less
```diff
@@ -80,6 +80,13 @@
   color: @error-text-color;
 }
 
+.right {
```
matheusd: I looked but didn't find these styles being used. In this actually being used somewhere?

PR #1169, opened 2018-02-14
Transaction Filtering By Address String

### LINKED ISSUE #1046: Add transaction filtering by address
Add a "search" bar at the top of the Transactions/History page that will let people see all transactions related to a particular address (either in debit or credit).  This will most likely require changes to the GetTransaction call to dcrwallet.  Feel free to drop a proposed change here before embarking on a large set of changes.

cipherzzz:
Take a look at this... It looked like the only time I would get an actual address is in the creditInputs for the transaction, so I am just filtering on those. I tested for Send/Receive/Transfer

Resolves #1046


matheusd:
I believe this is insufficient for actually fixing #1046. The address search can't be done on decrediton, because we stream transactions as needed from dcrwallet (so we usually don't have the full set of transactions in memory in decrediton).

You have to add address search capabilities to dcrwallet, to then request all transactions using the given address on decrediton.

cipherzzz:
@matheusd - I'm confused... It looks like all of the other filter criteria is happening locally after the `wallet/service.getTransactions` is called.  Are you saying that you want to create a new wallet method that will begin to pass in the filtering criteria as request parameters instead of doing it locally? Something like 
`export const getTransactions = withLogNoData((...filterParams)...`

matheusd:
Eek, you're right, forgot that the filtering by tx type was done on decrediton's side.

I'll take another look.

cipherzzz:
@matheusd - I reverted the spacing and formatting. It was being done by editorconfig, which actually formatted it correctly, but made a lot of noise as you stated. Eventually we should enforce some of the spacing standards as there are a lot of inconsistencies in the code as that goes. Should be good now, let me know if you need anything else...

app/wallet/service.js
```diff
@@ -95,6 +95,13 @@ export function formatTransaction(block, transaction, index) {
   const type = transaction.getTransactionType();
   let direction = "";
 
+
+  let debitAddresses = [];
+  transaction.getDebitsList().forEach((debit) => debitAddresses.push(debit.getPreviousAccount()));
```
matheusd: Since these aren't addresses, you should name this as "debitAccounts".

alexlyp:
Yah that's fine for a first pass, but we'll definitely want to be able to search the debits addresses as well.  As we've discussed before this will require some work in dcrwallet to accomplish.

app/wallet/service.js
cipherzzz: Done...

cipherzzz:
I am happy to look into the dcrwallet call next week it you want... I am also going to open an issue with setting the spacing for parentheses and brackets in our eslint file.

PR #1171, opened 2018-02-15
React router v4

### LINKED ISSUE #1047: Update to use react-router 4
Here's a pretty decent rundown of things that need to happen to migrate from 3.2.0 to 4:
https://codeburst.io/react-router-v4-unofficial-migration-guide-5a370b8905a

matheusd:
Fix #1047 

Big changeset, lots of lines touched but the gist of it should be reasonably easy to follow. 

The main difference introduced in react-router-v4 is that routes are now rendered *everywhere*. Nested routes in particular now work with more reacty semantics, allowing a more natural structure of the code. This PR restructures the code to take advantage of that. 

Instead of a single `routes.js`, each level of the app is now responsible for rendering itself. The structure stablished is as follows:

- `/` (`container/App.js`): top level i18n and theming init stuff
- `/getStarted` (`container/GetStarted.js`): top level get started stage (during launcher redesign we can insert the individual stages as pages instead of using a global opaque index number)
- `/[page]` (`container/wallet.js`): wallet subpages (this could be later refactored into using paths like `/wallet/[page]` for consistency)
- `/[page]/[tab]` (individual pages): each page now renders its own tabs if it is tabbed (no need for that information to be global anymore)

Since changing the react-router version needed a major rewrite of the nested routes, I ended up doing a few other cleanups:

- All wallet pages are now either `StandalonePage` or `TabbedPage` (except for overview which is different from all other pages)
- All legacy header stuff on the wallet pages has been removed and moved into the local components code
- Global messages and functions have been moved into the individual pages
- All `routes` props that were passed on tabbed pages are not needed anymore and have been removed
- Many unsued styles have been removed

There are many possible improvement refactorings that can be done in the future (eg: remove sidebar from getStarted, further unify layouts into shared components) but for now this should allow us to move to react 16 and keep up with all other major libraries.


alexlyp:
Awesome work!  First basic testing was successful, I'll dig in for a deeper review shortly.

app/components/views/TicketsPage/PurchaseTab/StakePools/List.js
```diff
@@ -11,7 +11,7 @@ const StakePoolsList = ({
   onRemoveStakePool,
   rescanRequest,
 }) => (
-  <div className="tab-card">
+  <div className="">
```
alexlyp: Replace with Aux

app/components/views/TicketsPage/PurchaseTab/Tickets.js
```diff
@@ -7,7 +7,7 @@ const Tickets = ({
   ...props
 }) => (
   <Aux>
-    <div className="tab-card">
+    <div className="">
```
alexlyp: another Aux

app/components/views/TicketsPage/StatisticsTab/Page.js
```diff
@@ -2,7 +2,7 @@ import "style/ReceivePage.less";
 import "style/MiscComponents.less";
 
 const MyTicketsPage = () => (
-  <div className="tab-card">
+  <div className="">
```
alexlyp: another Aux

app/components/views/TicketsPage/PurchaseTab/Tickets.js
alexlyp: or remove altogether

app/containers/App.js
```diff
@@ -3,21 +3,19 @@ import { IntlProvider } from "react-intl";
...
-const fade = { atEnter: { opacity: 0 }, atActive: { opacity: 1 }, atLeave: { opacity: 0 }};
-
-const wrapperComponent = props => <div className="page-view" { ...props } />;
+const topLevelAnimation = {atEnter: {opacity: 0}, atLeave: {opacity: 0}, atActive: {opacity: 1}};
 
 @autobind
 class App extends React.Component {
   static propTypes = {
-    children: PropTypes.element.isRequired,
+    // children: PropTypes.element.isRequired,
```
alexlyp: remove

alexlyp:
Just needs conflicts resolved, then should good to go.

matheusd:
Removed the duplicated `TxHistory` component on the home page after last rebase and added the reduced status component as an `overview` option to `TxHistory`.

There may still be some differences on home compared to current master or differences compared to the desired layout, but we can tweak those later if needed.

PR #1174, opened 2018-02-16
Properly populate recent regular and stake transactions

alexlyp:
Also add some various styling cleanup and refinement.


vctt94:
Good job, LGTM

The only thing I noticed is when clicking on tickets we get an error message 
`Cannot read property 'rawTx' of undefined`

alexlyp:
Hrm can't seem to replicate this issue.  All you do is click on a ticket listed in the Recent Ticket activity area?  It loads a tx details totally fine for me

app/style/HomePage.less
```diff
@@ -171,9 +177,21 @@
...
 
+.tx-history-overview-row.Vote .icon {
+  background-image: @ticket-voted-icon;
+}
+
+.tx-history-overview-row.Ticket .icon {
+  background-image: @ticket-small-icon;
+}
+
+.tx-history-overview-row.Revoke .icon {
```
matheusd: this should be `.Revocation`, otherwise the icon is not shown

vctt94:
Yes I just click on a vote actually. Tickets seems to work fine. Here is the log I am getting: 

bug

And the error is at TransactionPage: 

`if (!viewedDecodedTransaction) {
    decodeRawTransactions([viewedTransaction.rawTx]);
  }`

vctt94:
On master I have this: 

bug

I believe that is why the ticket worked. I just bought it, and it loads. If I have made many transactions I believe it would crash too. I will try it.

PR #1184, opened 2018-02-20
[launcher] Language select and checks for completion

### LINKED ISSUE #1119: Design Implementation: Launcher Language Selection
Visual Design and specs: https://xd.adobe.com/spec/8506c3f3-ebad-4872-8d8e-dc87c0eb2fed/

When a newly installed Decrediton wallet is opened up for the first time, this is the first view shown. 

`Continue` leads to the onboarding slides. 

Icons:
[decrediton-flags.zip](https://github.com/decred/decrediton/files/1675563/decrediton-flags.zip)
[decred-logo-launcher.svg.zip](https://github.com/decred/decrediton/files/1675564/decred-logo-launcher.svg.zip)

See [launcher-master](https://xd.adobe.com/spec/ab537a4e-2902-4260-99f1-85c3992f9a73/) more for reference or hit me up on slack.

alexlyp:
Closes #1119   ~Still need to figure out how to add the flags as given in the design.~

This will also begin checking to see if users need to complete these steps on startup.  If they have completed them then they will never see them unless they change the config


app/main.development.js
```diff
@@ -673,7 +673,7 @@ app.on("ready", async () => {
   if (!locale) {
     let newCfgLocale = appLocaleFromElectronLocale(app.getLocale());
```
matheusd: You might wanna copy/adjust something like this on `LanguageSelectPage.getInitialState()`, so that the language select shows by default the user's OS language, instead of being `selector.sortedLanguages[0]`.

Another option would be to add a selector such as `export const defaultLocaleName = () => appLocaleFromElectronLocale(...)`, and pass that along with the `walletSetup` connector and use that when `currentLocaleName` is empty (to make even the initial start page appear on the user's language by default).

app/main.development.js
```diff
@@ -673,7 +673,7 @@ app.on("ready", async () => {
...
     logger.log("error", `Locale ${cfgLocale} not found. Switching to locale ${newCfgLocale}.`);
-    globalCfg.set("locale", newCfgLocale);
+    // globalCfg.set("locale", newCfgLocale);
```
matheusd: you can remove from this

PR #1200, opened 2018-02-26
Overview stats bugs

matheusd:
Fixes three small bugs related to recent refactors:

- Returns the spin animation to the rescan button
- Prevents hover bug on charts on empty wallets
- Prevents hover bug on charts on an errored wallet (I had a wallet with a vote but not the related ticket, which was causing a BSOD; a rescan fixes the issue, but I couldn't easily trigger the rescan due to the BSOD).


jrick:
Decrediton was causing a BSOD?

matheusd:
BSOD - *Blank* screen of death (when an unhandled exception crashes and only a blank, white screen is shown) - not a full OS-level BSOD.

PR #1203, opened 2018-02-27
Chart styling clean up

alexlyp:
- 0 value bars were rendering fully instead of empty bars
- Tweaks to stacked bar radiuses
- Change sent to be debits and so it shows the stacked +/- chart correctly


alexlyp:
@vctt94 ah good call I'll add that

PR #1204, opened 2018-02-28
fix closing decrediton error

### LINKED ISSUE #1201: Error on Shutdown
`app/wallet/app.js:66 Uncaught (in promise) Error shutting down app`

Appears to be from the withLog wallet function, though not sure it is an actual issue

vctt94:
closes #1201


alexlyp:
I'm not seeing the error anymore, which is good.  But I'm not seeing an immediate reaction after I click the close.  It should still go immediately to the Shutdown screen while the sendSync("clean-shutdown") is occuring

PR #1206, opened 2018-02-28
[launcher] Replace old launcher designs with new

### LINKED ISSUE #1205: Advanced start daemon is with form names swapped
The Remote Daemon form is showing different location daemon form and vice versa

alexlyp:
Fixes #1205 

The launcher is nearing completion for the new design.  All the previously used forms for startup have been updated to match the design spec.


alexlyp:
Yah agree about the contrast in some areas.  Definitely worth trying to fix a few things prior to release.

PR #1208, opened 2018-03-01
[launcher] Wallet selection

### LINKED ISSUE #1132: Design Implementation: Launcher Overview – Display Wallets
Visual Design and specs: https://xd.adobe.com/spec/51a6aac8-f5a7-4620-8608-d1f4fb6b44d6/

Currently `active wallet` alongside with `other wallets` (if created, maximum 3) and `create/restore wallet` are displayed only when activated from `Settings` or if the user has not yet created a wallet (in scenario of cancelling onboarding during create/restore wallet). Otherwise if user finishes onboarding by creating a wallet and is re-directed to the `Launcher Overview`, `Display Wallets` should be set off by default. This for avoiding confusion between Accounts and Wallets.

`Create` or `Restore` wallet re-directs to the relevant views.

Clicking on a `Wallet` displays function icons: `Activate` or `Remove` Wallet, and incase of a currently active wallet - `Deactivate`. All functions are followed up by a modal for confirmation. 

In case of turning on both `Advanced Daemon Setup` and `Display Wallets` from `Settings`, Wallets should be placed below Advanced Daemon.

See [launcher-master](https://xd.adobe.com/spec/ab537a4e-2902-4260-99f1-85c3992f9a73/) more for reference or hit me up on slack.

alexlyp:
~requires #1206~

Closes #1132 

There were some changes from what the design document had suggested due to current workflow with the daemon and wallet startup.  Essentially the current way it works, the preferred network isn't known until the user picks which wallet to start.  Whereas the design document expects the blockchain to just begin loading and syncing in the background without any input from the user.  This can be done but will require a configuration setting that would be twiddled.


matheusd:
I'm getting a "two children using the same key: default-wallet" JS error here. Do you see that as well?

alexlyp:
Ah I know the issue.  Will fix.

app/components/views/GetStartedPage/WalletSelection/Form.js
```diff
@@ -1,65 +1,78 @@
...
-            </div>
-            <div className={"text-toggle-button-right " + (sideActive && "text-toggle-button-active")} onClick={!sideActive ? onShowSelectWallet : null}>
-              <T id="advancedDaemon.toggle.appdata" m="Create New Wallet" />
-            </div>
+    availableWallets && availableWallets.length > 0 && !createWalletForm ?
+      <div className="advanced-page">
+        <div className="advanced-page-form">
+          <div className="advanced-daemon-row">
+            {availableWallets.map(wallet => {
+              const selected = wallet.value.wallet == selectedWallet.value.wallet;
```
matheusd: This is making the default-wallet of both mainnet and testnet get selected, and then when starting it seems only the mainnet wallet is getting started.

matheusd: I believe you need to check by the wallet.label, which has the network appended

matheusd:
Re: error on selector, I just verified that only a one of my testnet wallets is giving me this problem on latest master, so seems to be an issue on my end.

vctt94:
@matheusd is this issue `ticketPrice of undefined`?  I got this error once, but can't reproduce it again. Ay-p got it as well. Maybe it is related with my PR.

app/components/views/GetStartedPage/WalletSelection/Form.js
vctt94: I am having the same issue. Here is what the log says: 
```
2018-03-05 09:41:52.018 [ERR] DCTN: Start RPC errored  {"code":5,"metadata":{"_internal_repr":{}},"details":"Connection to RPC server failed: mismatched networks"} 
```

PR #1225, opened 2018-03-06
Fix indexing bug on selector

matheusd:
Fix occasional `ticket is undefined` errors in selector.

The problem was the reference to the `allTickets` list by index when there was no relationship between `recentStakeActivity` and `allTickets`. `allTickets` stores ticket+spender and `recentStakeActivity` is a list of transactions, so to find the full ticket information based on an arbitrary stake transaction (which could be the ticket purchase itself, a vote or a revoke) we need to search by hash to find if we have the information.

Ordinarily, we should have that information (as it is returned by `dcrwallet.GetTickets()`) but on a few wallets I've hit a situation where the recent transactions list has a transaction *not* returned by `GetTickets` (still looking into it, but it's related to my ticket splitting tests - dcrwallet doesn't correlate the vote to the previous ticket). I've left a protection against that case for the moment.

We need to decide whether we want to show the most recent **transactions** or the most recent **tickets** on the home page: on wallets with few stake transactions, if we show by transactions we get some duplicated information: namely the ticket and vote txs - see screenshot). 

screenshot_2018-03-06_10-15-11

### DIFF
--- a/app/selectors.js
+++ b/app/selectors.js
@@ -433,19 +433,37 @@ const recentStakeTransactions = createSelector(
   [ transactionsNormalizer, get([ "grpc", "recentStakeTransactions" ]) ], apply
 );
 
-export const homeHistoryTickets = createSelector(
-  [ recentStakeTransactions, allTickets ],
-  ( homeTickets, allTickets ) => {
-    return homeTickets.map( (ticket, index) => {
-      const ticketDecoded = allTickets[index];
-      ticket.ticketPrice = ticketDecoded.ticketPrice;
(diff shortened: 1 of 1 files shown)


matheusd:
It is, but if we do it in the selector we won't always have the 5 most recent transactions (eg: in the picture simply dropping the ticket would make us show only 4 transactions).

alexlyp:
Hrm yah I think this may be related to phantom votes possibly.  Fix looks good for now until we can determine what's going on elsewhere to give these false hashes.

PR #1226, opened 2018-03-06
[launcher] Remove network toggle and autoload daemon

alexlyp:
This PR is an attempt to have the UX be more like what @linnutee had suggested in the docks.  Which network that the wallet will be using will be determined by the config instead of chosen on startup.  This now allows us to auto-load the daemon (if not in advanced mode) so the user can have the daemon syncing while they create/select their wallet.


alexlyp:
That's correct only way to get to testnet is to toggle in config.json.  We can probably think about adding back a network toggle in the settings that also has warnings and restarts everything.

PR #1227, opened 2018-03-07
Add tickets stat charts and exports

matheusd:
This:
- Adds new series to the balance stats (total historical staked amount, rewards and fees)
- Adds new vote time stats (histogram of how many tickets voted in how many days)
- Adds a new tickets stats (ticket purchases and returns with associated amounts)
- Adds exports for the above
- Adds charts for the above

Will require further tweaks after review from design team, and I think the daily roi chart is kinda useless for stakeholders with small amounts (probably weekly or monthly makes more sense in that case) but I'm releasing as is for a first iteration.


alexlyp:
Vote graph looks good.  For the ROI graph I think you should show a line graph with each plot being Rewards - Fees / Total Stake (assuming fees haven't already been taken into account).  Then on the somewhere next to the graph show those current totals.

PR #1231, opened 2018-03-08
Prevent second app instance on non-advanced mode

matheusd:
We won't have time before 1.2.0 to implement all requirements for simultaneous multiple wallets on non-advanced mode, so I'm returning the code to alert and prevent running multiple instances on that situation.


matheusd:
You can, but it isn't needed. Set the config files to use port 0, dcrwallet will pick a random port and decrediton will detect and connect to that port.

That's the default behaviour for new wallets as well.

PR #1240, opened 2018-03-12
Various proofreading issues in Tutorial

### LINKED ISSUE #1234: Proof Read Articles in "Learn The Basics"
I'm too scared of messing up the translations so here is an issue instead of a PR.
- determinstically => deterministically
- Some whitespaces missing. What sentence are responsible for starting or ending with a whitespace, the former or latter? Example: 
> period is 142 days).A succesful vote

This sentence doesn't make any sense:
> It is like a global, decentralised spreadsheet or ledger that records all the activity that happens when Decred funds **are transferred keeping track of** how many tokens are sent, and what the balance of every account is.

I had default language activated.

alexlyp:
Close #1234


matheusd:
I'm still seeing a joined string on page 3: "the mnemonic**seed is a list...**"

PR #1241, opened 2018-03-12
Identify incomplete wallets

### LINKED ISSUE #1229: When creating wallet, folders are created but not deleted if "canceled"
When we start creating a wallet, if you type a wallet's name and hit create wallet, the directory is created on the wallets directory at decrediton dir with the config.json and the conf files. If we cancel it (like refreshing decrediton), the folders are not deleted and kept at wallets dir.

We should delete the directories or create them on the next step.

alexlyp:
Close #1229 

Now user is shown incomplete wallet set ups and which will allow them to continue set up or remove on future decrediton loads.


vctt94:
Here is a picture: 

decreditonbug

alexlyp:
Not sure I understand what the problem is

vctt94:
Is it happening with you too? I checked and it is happening on master as well

alexlyp:
I'm not seeing any issue locally.

Issue #916, opened 2017-11-27
Show "New App Version Available" notification


matheusd:
Consensus has been that no auto-update feature will be implemented to the wallet. But Decrediton could show a "New Version Available" dismissable notification or always present notice (to be decided with the design team).

This involves:
- [x] Decide and implement new version detection (either querying the releases on github or some other api endpoint on decred.org)
- [x] Implement notice UI

jrick:
This should be opt-in, since it is a form of phoning home.

matheusd:
Decrediton currently makes some other requests that could also presumably be used for tracking (stakepool listing, latest block height). Do you think we should add a global opt-in to prevent requests for anything other than the wallet? Or add individual opt-ins for the other requests? Or disregard the others and add the opt-in only for the update notification?

jrick:
I'm not sure how/when decrediton is making that request, but i would prefer that to be controllable by the user as well, perhaps a button to load the stakepool list.

But that's just something we should work towards (and I can open a new issue for that) but even without that done, we shouldn't make the situation worse.

alexlyp:
Yah I believe this discussion is partially related to https://github.com/decred/decrediton/issues/908, where we should have some sort of "anonymous" mode where we don't make any outbound calls and if we do we can use a tor proxy or something.

matheusd:
@alexlyp They are related, though sort of orthogonal: I might not want to make *any* outbound connections, even if I use a traditional/tor proxy for my network. Or I may want to disable them while I'm on a trip or something.

@jrick That's why I was asking those questions: get a feel for what you believe is important then open issues to track this. Current outbound calls made are:

- Stakepool listing (called when initializing a config after install and on the stakepool page if no stakepool configs are available).

- Last Block Number (called when initializing the app and every 5 minutes when syncing the blockchain - used to determine the current block height and present a better estimate for the user when the chain has been fully downloaded).

There's also some talk about adding more stuff to the overview page (https://github.com/decred/decrediton/issues/904). And there's also stuff we could eventually add, like usage statistics, ability to send error reports/logs somewhere, etc.

So I guess the best course of action would be to start adding granular opt-ins for this stuff and have a settings tab/division/whatever holding these, with explanations of what exactly they do and where they access and a "master switch" to disable all outbound (non-wallet) connections.

Code-wise, we should restrict all outbound calls to use a custom middleware (encapsulating axios), make sure no new code uses anything other than this middleware and then that can implement the checks for the opt-ins.

Issue #908, opened 2017-11-27
Add proxy configuration


matheusd:
Most of the work done by decrediton is actually done by dcrwallet/dcrd, but a few actions, mostly related to the stakepool api (search for `axios` on the source code) are done directly by the app.

2 things are needed for this issue:

- [ ] Check whether the system proxy is used by default on electron apps in general (and Decrediton in particular) and if that is working for the requests currently made by Decrediton
- [ ] If not, implement some way to configure the proxy that Decrediton will use.

tsaiDavid:
What's the process for indicating that I'd like to try to help with this issue?

alexlyp:
that's it! :)

Roughly speaking we'd like to limit the amount of phoning home that's going on, but in the instance that we do need to communicate out, we'd like to impl Tor comms or something like that.  Feel free to open up discussion about what you're thinking here.  You can also find us on our slack/matrix/rocketchat

PR #1242, opened 2018-03-13
[WIP] Whitelist outbound calls

matheusd:
Pre-requisite for #908 and #916 

This uses electron's [WebRequest](https://electronjs.org/docs/api/web-request) facility to block all outgoing calls except for those whitelisted. 

If you run this as is, you'll start getting warnings on the console for "Cancelling Outbound request: ...". To allow requests to the external decred api to get latest block number, modify the global config.json to include the following entry:

```json
{..., "allowed_external_requests": ["EXTERNALREQUEST_NETWORK_STATUS"]}
```
Sending for early review to see if the approach is sensible. Still needed before completing:

- [x] Add startup page for enabling standard outbound calls on first setup (or to choose which calls to enable)
- [x] Allow calls to a configured stakepool
- [x] Add settings page to revoke/modify allowed outbound calls


matheusd:
Preview of the startup screen. Actual icons yet to be made...

screenshot_2018-03-13_16-15-23

jrick:
These look like official icons to me.

matheusd:
Ready for testing and review.

PR #1245, opened 2018-03-13
[WIP] Add proxy config options

### LINKED ISSUE #908: Add proxy configuration

matheusd:
Requires #1242 
Fix #908 

This uses electrons [Session](https://electronjs.org/docs/api/session#sessetproxyconfig-callback) facility to setup a global proxy for outbound connections, which opens up the possibility of using Tor for accessing third party (non-dcrwallet/dcrd) services.

For the moment, only a single proxy location can be specified as I believe that is sufficient for Decrediton's current and future planned connections.

Current proxy types supported are None (direct connection), PAC file, http, SOCKS4 and SOCKS5.

I've tested the http option using a local tinyproxy instance and the SOCKS5 using an ssh tunnel connection.


matheusd:
Also just tested using the following PAC file, using an upstream tinyproxy. I configured the PAC file location as `file:///home/user/test.pac`.

```
function FindProxyForURL(url, host) {
	// our local URLs from the domains below example.com don't need a proxy:
	if (shExpMatch(host, "localhost"))
	{
		return "DIRECT";
	}

	// All other requests go through port 8888
	return "PROXY localhost:8888";
}
```

matheusd:
The UI is in a very early (read: rough) state, as the privacy and proxy settings have added a lot of new stuff.
Let me know if you want to me to tweak before merging or leave the UI refactoring to after designers take a look.

PR #1270, opened 2018-03-20
Add blur when modal is active

matheusd:


alexlyp:
Nice really elegant fix!

Can you remove the `background-color` from `app-modal-overlay`?

And can you add `box-shadow: 0 15px 34px 0 rgba(12, 30, 62, 0.34);` to `app-modal.info-modal` and `app-modal.passphrase-modal`

matheusd:
Made the fixes. Without the background-color there's no point of having the overlay div, so I removed it as well.

PR #1271, opened 2018-03-20
Improve send confirmation modal

alexlyp:
Add addresses to confirmation modal.  Requested from `adreny` on slack.


matheusd:
I'm seeing the value unaligned to the "Total Spent" label, do you see that as well?

Another idea: when doing a transfer (send-to-self) maybe we wanna show the destination account name? Or you don't think that's needed?

PR #1287, opened 2018-03-23
Add empty state to statistics page and update styling

### LINKED ISSUE #1277: Add an empty state for Statistics page as well
We should hide the graphs on the statistics page until there is information to graph.

Also looks like the graphs aren't playing nicely with the expanding/reducing sidebar

alexlyp:
Fixes #1277


matheusd:
Tested fine. Linnutee updated the issue with a new icon, so just need to fix that.

alexlyp:
New no-stats indicator added

PR #1294, opened 2018-04-04
Create temp dcrd.conf if none is found in expected location

### LINKED ISSUE #1293: Initital start up expecting dcrd.conf to exist
In a sort of chicken and egg situation, we expect dcrd.conf to exist for starting dcrd but it won't be generated until dcrd starts for the first time.  We should have a check for conf existence and use a special case in that scenario.

alexlyp:
Fix #1293 
dcrd does not create a conf file if a conf file is specified in the options (which we do for decrediton).  So we now just create a temp dcrd.conf in appdataDirectory to be used until the user places (or generates) a config file in the default dcrd location.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -131,9 +131,7 @@ export function getWalletCert(certPath) {
 
 export function readDcrdConfig(configPath, testnet) {
   try {
-    console.log(configPath);
     if (!fs.existsSync(dcrdCfg(configPath))) return;
-    console.log("jherere", configPath);
     const readCfg = ini.parse(Buffer.from(fs.readFileSync(dcrdCfg(configPath))).toString());
     let newCfg = {};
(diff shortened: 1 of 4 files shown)


app/config.js
```diff
@@ -131,7 +131,9 @@ export function getWalletCert(certPath) {
 
 export function readDcrdConfig(configPath, testnet) {
   try {
+    console.log(configPath);
```
matheusd: remove

app/config.js
```diff
@@ -131,7 +131,9 @@ export function getWalletCert(certPath) {
...
     if (!fs.existsSync(dcrdCfg(configPath))) return;
+    console.log("jherere", configPath);
```
matheusd: remove

PR #1303, opened 2018-04-09
Translations for 1.2.0rc2

matheusd:
Adds a couple of missing translation items and updates the pt/br translations based on changes made after 1.2.0rc1


matheusd:
Fixed. Thanks for catching those!

PR #1317, opened 2018-04-19
Refactor Control Actions

alexlyp:
This should go in after v1.2.0 due to the various changes that may cause issues. 

Rescan and ConstructTransaction still require refactor due to some special cases therein.


alexlyp:
Tested most functionality, noticed I need to fix rename and get next account.

app/actions/ControlActions.js
```diff
@@ -409,184 +265,111 @@ export const SETMAXPRICEABSOLUTE = "SETMAXPRICEABSOLUTE";
...
+  if (stakePool.TicketAddress !== getTicketBuyerConfigResponse.getVotingAddress())
+    promises.push(wallet.setTicketBuyerVotingAddress(ticketBuyerService, stakePool.TicketAddress));
+  if (stakePool.PoolAddress !== getTicketBuyerConfigResponse.getPoolAddress())
+    promises.push(wallet.setPoolAddress(ticketBuyerService, stakePool.PoolAddress));
+  if (stakePool.PoolFees !== getTicketBuyerConfigResponse.getPoolFees())
+    promises.push(wallet.setPoolFees(ticketBuyerService, stakePool.PoolFees));
+  if (maxPerBlock !== getTicketBuyerConfigResponse.getMaxPerBlock())
+    promises.push(wallet.setTicketBuyerMaxPerBlock(ticketBuyerService, maxPerBlock));
+
+  return promises
```
matheusd: This seems wrong to me... `promises` is an array, it shouldn't have the `.then()` method, but the function is craping out earlier than this when trying to update the config

app/actions/ControlActions.js
```diff
@@ -409,184 +265,111 @@ export const SETMAXPRICEABSOLUTE = "SETMAXPRICEABSOLUTE";
...
+  maxPerBlock, stakepool
+) => (dispatch, getState) => {
+  dispatch({ type: STARTAUTOBUYER_ATTEMPT, });
+  return wallet.startAutoBuyer(
+    sel.ticketBuyerService(getState()), passphrase, accountNum, balanceToMaintain*1e8, maxFeePerKb*1e8,
+    maxPriceRelative, maxPriceAbsolute*1e8, maxPerBlock, stakepool
+  )
+    .then(startAutoBuyerResponse => {
+      dispatch({
+        success: "You successfully started the auto ticket buyer.",
```
matheusd: Don't need this anymore as the snackbar reducers show the correct message given the `STARTAUTOBUYER_SUCCESS` action

app/actions/ControlActions.js
```diff
@@ -409,184 +265,111 @@ export const SETMAXPRICEABSOLUTE = "SETMAXPRICEABSOLUTE";
...
-          dispatch({ success: success, stopAutoBuyerResponse: stopAutoBuyerResponse, type: STOPAUTOBUYER_SUCCESS });
-        }
-      });
-  };
-}
+export const stopAutoBuyerAttempt = () => (dispatch, getState) => {
+  dispatch({ type: STOPAUTOBUYER_ATTEMPT });
+  return wallet.stopAutoBuyer(sel.ticketBuyerService(getState()))
+    .then(stopAutoBuyerResponse => dispatch({
+      success: "You successfully stopped the auto ticket buyer.",
```
matheusd: No need for this anymore

alexlyp:
OK issues with ticketbuyer should be resolved now.

matheusd:
Fixes are working. Only one extra question: are the settings supposed to be persisted across runs of the app? I modified the "balance to maintain", updated the config, reloaded the app and the setting was back to 0.

PR #1341, opened 2018-04-27
Capture invalid seed pasting and display error

### LINKED ISSUE #1339: [launcher] Pasting seed into restore wallet disappears
When ctrl-v pasting a seed into the restore wallet form, it causes the form word selectors to disappear.  They only reappear after toggling from hex and back.  (toggling to/from hex, should also not be clearing the form as well.

alexlyp:
Close #1339 

Way to replicate error: C-V any bad words into any box in the restore seed area.  Previously, it would make the seed words disappear, now it determines whether a valid 33 mnenomic seed so it will display correctly.

### DIFF
--- a/app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
+++ b/app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
@@ -94,7 +94,7 @@ class ExistingSeedForm extends React.Component{
                 <T id="confirmSeed.warnings.pasteExistingSeed" m="*Please make sure you also have a physical, written down copy of your seed." />
               </div>}
               {this.state.showPasteError &&
-              <div className="red-error">
+              <div className="error">
                 <T id="confirmSeed.warnings.pasteExistingError" m="* Please paste a valid 33 word seed."/>
               </div>}
               {seedWords.map((seedWord) => {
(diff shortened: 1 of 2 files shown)


app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
```diff
@@ -80,9 +89,14 @@ class ExistingSeedForm extends React.Component{
           </div>
           {seedType == "words" && Array.isArray(seedWords) ?
             <div className="seedArea">
-              {!this.state.showPasteWarning ? null : <div className="orange-warning">
+              {this.state.showPasteWarning &&
+              <div className="orange-warning">
                 <T id="confirmSeed.warnings.pasteExistingSeed" m="*Please make sure you also have a physical, written down copy of your seed." />
               </div>}
+              {this.state.showPasteError &&
+              <div className="red-error">
```
matheusd: Could you change this from `red-error` to `error`?

app/style/CreateWalletForm.less
```diff
@@ -51,6 +51,12 @@
   top: 122px;
 }
 
+.red-error {
```
matheusd: and change this to `.seed-area .error`?

Just `red-error` seems a too generic name for a class with `position: fixed`.

app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
alexlyp: sure thing

PR #1355, opened 2018-05-03
[launcher] Add go back button to wallet creation

### LINKED ISSUE #1351: Launcher UX: Missing Close/Return
screen shot 2018-05-02 at 19 32 06

screen shot 2018-05-02 at 19 31 05

Create / Restore wallet views needs the` close button` added. This returns user back to launcher overview. 

Otherwise it's a potential dead-end, if user decides to not create another wallet, has no way to return to previous view w/o closing Decrediton. 

X = same icon as in content pages.

alexlyp:
Close #1351

This is a bit more of an involved process due to the required seed service for wallet creation.  We need to shutdown the previously running wallet, remove that wallet then allow the user to select/create another wallet.


app/actions/WalletLoaderActions.js
```diff
@@ -53,11 +54,27 @@ export const CREATEWALLET_NEWSEED_CONFIRM_INPUT = "CREATEWALLET_NEWSEED_CONFIRM_
...
+export const createWalletGoBackWalletSelection = () => (dispatch, getState) => {
+  const { daemon: { walletName, network } } = getState();
+  wallet.stopWallet().then(() => {
+    wallet.removeWallet(walletName, network == "testnet")
+      .then(() => {
+        dispatch({ type: CREATEWALLET_GOBACK });
+        dispatch(getAvailableWallets());
+      })
+      .catch((err) => {
+        console.log(err);
```
matheusd: Probably wanna dispatch a `{error, type: REMOVEWALLET_FAILED}` or similar here, as otherwise in case of an error the user might reach the 3 wallet limits without having any feedback an error has happened.

And then add the action type on the snackbar reducers here: https://github.com/decred/decrediton/blob/master/app/reducers/snackbar.js#L240

PR #1356, opened 2018-05-03
Refactor main.development.js launch and shutdown

### LINKED ISSUE #1365: [bug] Show wallet log files menu button does nothing
When clicking on `Show wallet log files` button at the advanced menu it does nothing.

vctt94:
This PR refactors dcrd and dcrwallet launch and shutdown and fixes #1365.

~~I am having trouble showing dcrwallet and dcrd logs right now and I need to find a solution for that.~~

~~As our log variables are at main.development and launch is an async method I can not buffer them, even when passing by reference with objects, and they stay empty buffers.~~

Thanks for your help @matheusd , I made as you said and logs are working now.


matheusd:
Eyeballing the code seems reasonably fine (haven't tested yet).

The log thing you can probably solve by moving the `dcrdLogs/dcrwalletLogs` buffers into the `main_dev/logging.js` and importing in both `main.development.js` and `launch.js`. 

You might need to slightly refactor to not use variables though: you might need to leave those two as unexported in  `logging.js` and create get/set pairs to modify their content. So, instead of:

```
dcrd.stdout.on("data", (data) => dcrdLogs = AddToLog(process.stdout, dcrdLogs, data));
```

You'd have something like
```
dcrd.stdout.on("data", (data) => AddToDcrdLog(process.stdout, data));

// in logging.js
let dcrdLogs = Buffer.from("");
export const AddToDcrdLog = (destIO, data) => { ... }
export const GetDcrdLogs = () => dcrdLogs
```

PR #1357, opened 2018-05-03
[i18n] Switch informational strings to markdown

matheusd:
This switches all text of informational modals, tutorials and major notices and warnings from embedded into the code into a markdown subsystem.

The big informational texts are easier to read, review and fix this way. Also, it considerably reduces the strings that need to be translated for the standard uses of the UI and makes it less brittle to update the source strings on transifex (as even a small typo on any of the big texts required a full refresh of sources). 

Also, this enables us to translate the app in two phases (functional UI and then informational texts) and review those text more thoroughly. And it opens up the possibility of embedding more fully featured documentation (from dcrdocs or custom made) into the app.

The number of strings to translate is altered the follow way (as measured by the `originals.json` file before and after this PR):

Before change: 715 individual strings, 4905 words
After change: 607 individual strings, 2245 words


matheusd:
Just ran another test: only 61 strings have duplicates, and most of those are only repeated once.

Only 20 have more than 2 occurrences and those are *really* simple stuff (like "cancel" or "continue"), so  I don't think we need to worry about dupes when sending to transifex.

Issue #1362, opened 2018-05-05
Finalize Overview Graphs visual design


linnutee:
Specs: https://xd.adobe.com/spec/37a39e55-d94b-43ab-a73e-d405f6b502b3/#screen/2e5abb57-0a08-4cb9-924e-f66b08ed3512/Overview%20–%20Graphs%20-%20Specs

A few more details to finalise the graphs in Overview.

1 Color of the border strokes changed to #F3F6F6
2 The smaller strokes next to each date and value can be removed. 
3 Bg color of the hover changed to #E9F8FE
4 All of the bars need rounded edges at both top and bottom ends, currently some under Tickets do not have. 
5 Overall the graph should be slightly more condensed in width, currently seems to be a fair bit wider than in the visual design example. 
6 Balance Graph data: This should only display `Available` and `Locked` funds. `Immature` belongs to Tickets category. 
7 Tickets Graph: This should only display ticket statuses, `Immature`, `Live`, `Voted` and perhaps also `Revoked` (which can be with color #FD704A)
@alexlyp can perhaps comment on the graphs data, whether another reason for showing those like that. 

<img width="999" alt="screen shot 2018-05-05 at 19 53 14" src="https://user-images.githubusercontent.com/17774057/39665517-60264600-509e-11e8-8973-4ab1e0ad6348.png">
<img width="979" alt="screen shot 2018-05-05 at 19 51 30" src="https://user-images.githubusercontent.com/17774057/39665518-63cc09de-509e-11e8-86be-5de4e1e27171.png">
<img width="896" alt="screen shot 2018-05-05 at 19 57 43" src="https://user-images.githubusercontent.com/17774057/39665528-9f77cc84-509e-11e8-8f2f-24706e6b5c47.png">

alexlyp:
So I was able to complete a few of these issues, 

Lingering issues that will require extra work:  

1) Stacked graphs in recharts don't play well with have rounded tops and bottoms.  Instead of overlayed, they are purely stacked on top so the bottom of one value is seperated from the top of another.
2) Unable to have a different axis stroke color while having a different axis label color.

linnutee:
Optics wise that out-of-ordinary grey (2.) is the definitely the biggest grind on the eyes. Any way it could perhaps be masked out then or drawn over with another element?

alexlyp:
That would be a huge undertaking.  Let me try to hotwire things to make it work.

PR #1372, opened 2018-05-08
[overview] Various styling issues on overview

### LINKED ISSUE #1363: Overview Layout: Visual Fixes
Specs: https://xd.adobe.com/spec/304c38a9-347c-453b-a679-63b324d51c79/#screen/3c497833-7011-4f6b-bdf9-b9e2785e8d95/Overview%20Layout%20Fixes

1 Textual information in Tickets tab (Total Value of Live tickets, etc) needs to be in two columns. This looks better as 4 rows is too much on top of each other. 

2 `Go To All Transactions` should be renamed to `See All →` with arrow from Source Sans Pro glyphs. The position, type size and color of it also needs fixing. Nb! This should not display at all when there are no transactions or tickets. 

3 The spacing above Recent Transactions and Staking Activity needs to be greater, check specs. 

4 Not sure on the rendering, but the title font size also looks larger than it should be (Recent Transactions and Staking Activity)

5 Also refer to point no 5 in issue #1362 about graphs and the overall size of them.

Will also follow-up one issue about the Staking and TX cards. 

screen shot 2018-05-05 at 20 25 16

alexlyp:
Closes #1363 

There are some lingering issues in #1362 that will require extra work.  Due to the usage of recharts the stacked charts aren't able to nicely have rounded tops and bottoms that lay on top of eachother.

### DIFF
--- a/app/components/charts/BalanceChart.js
+++ b/app/components/charts/BalanceChart.js
@@ -1,21 +1,18 @@
 import { BarChart, Bar, XAxis, YAxis, Tooltip } from "recharts";
 import { injectIntl } from "react-intl";
 import messages from "./messages";
-import { yAxisStyle, xAxisStyle, homeChartSize, padding, radiusTop,
-  radiusMiddle, radiusBottom } from "./Styles";
+import { yAxisStyle, xAxisStyle, homeChartSize, padding, radiusTop, radiusBottom } from "./Styles";
 import ChartTooltip from "./ChartTooltip";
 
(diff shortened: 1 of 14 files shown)


matheusd:
I'm seeing just two slight problems on this:

The date on the charts tooltips is at the bottom instead of the top. 

image

You need to use more specific css classes for the charts tab headers, otherwise the tabs in other parts of the app get the border:

image

PR #1374, opened 2018-05-09
Refactor main.development electron templates

### LINKED ISSUE #1137: Cleanup/Refactor main.development.js
This file has become quite large and bloated.  We should split out functionality into reasonable modules.

Short list of potential modules:
- Config (already exists, but could use clean up)
- Logging (already exists)
- dcrd/wallet launching/shutdown
- Ipc handling
- electron window/menutemplate creation

If you would like to work on this issue please submit a PR per module so review can proceed quickly.

vctt94:
This PR refactors electron templates on main.development.

I believe it closes #1137, what do you guys think? Any other improvements we can have about it?

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -2,14 +2,7 @@ import fs from "fs";
 import Store from "electron-store";
 import ini from "ini";
 import { stakePoolInfo } from "./middleware/stakepoolapi";
-import { app } from "electron";
 import { appDataDirectory, getGlobalCfgPath, dcrdCfg, getWalletPath, dcrwalletCfg, getDcrdRpcCert } from "./main_dev/paths";
-
-// setPath as decrediton
-// app is not set after building so we need that check
-if(app) {
-  app.setPath("userData", appDataDirectory());
-}
 
 export function getGlobalCfg() {
   const config = new Store();
--- a/app/config.js
+++ b/app/config.js
@@ -2,7 +2,13 @@ import fs from "fs";
 import Store from "electron-store";
 import ini from "ini";
 import { stakePoolInfo } from "./middleware/stakepoolapi";
+import { app } from "electron";
 import { appDataDirectory, getGlobalCfgPath, dcrdCfg, getWalletPath, dcrwalletCfg, getDcrdRpcCert } from "./main_dev/paths";
+
+// setPath as decrediton
+// app is not set after building so we need that check
+if(app)
+  app.setPath("userData", appDataDirectory());
 
 export function getGlobalCfg() {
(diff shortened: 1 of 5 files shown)


alexlyp:
Looks good overall, though first time I launched the PR it noted that I didn't have a locale set, not sure what could have been the cause of that.  2nd load seem to be fine.

vctt94:
Thanks for catching this bug.

This error happens because locale at config.json does not match any known locale. But the reason this was happening, is because when template was initiating config, decrediton config directory was not set yet as decrediton, so it used the Electron directory instead. 

I fixed moving app.setPath to config, but looks like app is not accessible after build by config, so we need a check for it.

Tested and things seems to be working now.  But I'd appreciate another test.

app/config.js
```diff
@@ -2,8 +2,14 @@ import fs from "fs";
 import Store from "electron-store";
 import ini from "ini";
 import { stakePoolInfo } from "./middleware/stakepoolapi";
+import { app } from "electron";
 import { appDataDirectory, getGlobalCfgPath, dcrdCfg, getWalletPath, dcrwalletCfg, getDcrdRpcCert } from "./main_dev/paths";
 
+// setPath as decrediton
+// app is not set after building so we need that check
+if(app)
```
matheusd: Very small nit.

Either add braces ({}) on the whole if or leave it at the same line. And add a single space after the `if`.

matheusd: Also, I'm not sure I'm following the comment ("app is not set after building").

I added a small `console.log` on both branches of the if, and tested after a `yarn dev`, `yarn build/start`, `yarn package-dev-linux` and all three go through the `true` branch (ie, app **is** set).

Can you explain a bit better what you're seeing when app is not set?

matheusd:
Ok, I see. This is not a build vs non build, but code on the node.js stack vs on the react stack.

I'm not terribly happy this code is in config.js vs on main.development, since this is supposed to be called only on the node stack (I mean the `app.setPath` call) but if @alexlyp doesn't see a problem, then it's fine by me.

Much cleaner code on main.development now, good job! :)

vctt94:
I see, thanks for clarification.

It makes sense being called only on node stack, I can try a work around on it so we do not need to call it on config.js.

alexlyp:
Ya after looking more closely at the change, I'd like a different solution than the if (app) {} in config.js

vctt94:
I moved back locale to main.development, because there is not really a reason for moving it to templates. I believe this way is better.

alexlyp:
cool looks better this way.

Issue #1243, opened 2018-03-13
SPV intergration checklist


alexlyp:
Here is all the various issues that need to be complete for SPV to be available for usage with decrediton:

- [x] Add "SPV mode" config option
- [x] Remove daemon connectivity while in "SPV mode"
- [x] Remove/replace block count request from insight with another signal from wallet for syncness check
- [x] Remove various functionality that will not be available in "SPV mode"

vctt94:
Can I work on this issue?

alexlyp:
Well let's start discussing various things we need to do here.  The changes in dcrwallet are still a WIP, but we should be able to start completing some of the easier tasks.

vctt94:
Cool, I want to help with dcrwallet checklist as well. I am thinking on start with 

- [ ]  Add "SPV mode" config option
And
- [ ]  Remove various functionality that will not be available in "SPV mode"

How does that sound?

alexlyp:
Yah I think that's a good start

PR #1389, opened 2018-05-14
Start spv mode

vctt94:
This PR starts #1243. 

It adds SPV to config and remove the following functionalities when in spv.

- purchase ticket 
- ticket auto buyer
- vote

### DIFF
--- a/app/components/shared/ShowWarning.js
+++ b/app/components/shared/ShowWarning.js
@@ -1,0 +1,9 @@
+const ShowWarning = ({ warn, className }) => (
+  <div className={"warning" + (className ? (" " + className) : "")}>{warn}</div>
+);
+
+ShowWarning.propTypes = {
+  warn: PropTypes.object.isRequired,
+};
+
+export default ShowWarning;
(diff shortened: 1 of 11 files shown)


matheusd:
AFAIK, it's not that you won't be able to purchase tickets, you just won't be able to use the ticketbuyer ("automatic buyer" in decrediton) for that. 

Also, the voting in decrediton is used to set vote choices in the pool, **not** for a solo voter, so those actions also won't be affected by the SPV work.

jrick:
Buying tickets is problematic for SPV wallets because they won't know the ticket price at the beginning of a new interval.  The ticket price (sdiff) in the block header is the stake difficulty to buy tickets in that block, not the next block.

matheusd:
Right, I don't know all details how SPV is being implemented, but doesn't the wallet have all block headers, implying it knows the current block height? If so, we can just prevent the wallet from buying tickets near the end of the sdiff change window and wait until at least one (or better yet, a couple of blocks to protect against reorg) have been received.

matheusd:
Don't really need to change dcrwallet for that. You only need to get the `currentBlockHeight`, and check if it is near the edge of the an sdiff change (basically `currentBlockHeight mod WorkDiffWindowSize`).

That might actually be something interesting to display on decrediton's purchase ticket screen (how many blocks until an sdiff change).

dajohi:
@vctt94 can you rebase?

vctt94:
rebased

PR #1397, opened 2018-05-16
[launcher] Wallet selection screen update 

### LINKED ISSUE #1369: Confusing splashscreen for restoring keys
If you do a fresh install of Decredition and want to restore your keys it is a bit confusing because you need to click "create new wallet" and then get a choice between creating a new wallet or restoring from a seed.

Suggestion: Have two buttons, create new wallet and restore from seed, and skip the step after "create new wallet".

### LINKED ISSUE #1383: Implement updated Wallet Cards design
Specs: https://xd.adobe.com/spec/ff91a336-c60d-4ad0-90c3-91780cf9082d/
Preview: https://xd.adobe.com/view/33dbd420-7478-48bc-9a36-6cfb4a230496/

Breakdown: 
- When the blockchain is still loading, the `Launch Now →` isn't displayed.
- When blockchain finishes loading, `Launch Now →`fades in (e.g. 130ms transition / quart type easing) 
- After fading in, the arrow (→) should begin an animated loop of doing a slight nudge/bounce to the right. 
(Example value: 130ms to change pos. in X +8px; 80ms return to original position w/ minimal bounce; 340 ms break; repeat). 
This should overall be something subtle like this, but is needed to have as a reminder where to launch a wallet from. 
- Clicking on the new `Edit Wallet` button displays the `Remove Wallet` and `Change Network` icons on all wallet cards as well `Cancel Changes`only on cards that are currently hovered over. Also changes Edit Wallet to a `Cancel` icon (also cancelling editing).
- Remove and Change Network open confirmation modals. 
- `Last Accessed` is only displayed on the currently selected wallet and when hovering on other wallets. 
- Also check Card States (https://xd.adobe.com/spec/ff91a336-c60d-4ad0-90c3-91780cf9082d/#screen/a7906e1f-2c29-423f-8f29-ca0c59ec8923/Card%20States) for updating the hovers a bit, currently not selected wallets have a blue outline when hover, should be grey. 

dcr_launcher_animation

Nb! Edit icon is currently a placeholder, will follow-up soon with another issue to several icons as a group.

alexlyp:
Close #1369 
Close #1383 

This now switches the daemon loading to before wallet load on advanced daemon mode, which also means moving remote/appdata into global config rather than wallet config.  

We have now also cleaned up the wallet creation flow to remove the secondary page.  From wallet selection it now goes directly to the desired creation screen.


alexlyp:
Yah open an issue with that seed word.  Was able to recreate that

matheusd:
Using existing wallets worked fine.

But when simulating an empty install (renamed `~/.config/decrediton/wallets` to `_wallets`) the app went directly to "restore wallet" without showing me the seeds for the new wallet (or giving me the choice to either create or import a wallet). Having a single pre-created wallet worked fine.

We might want to show the logs (and maybe settings) on the initial screen now, since if the blockchain loading hangs we want users to be able to see the logs. Don't need to add it to this PR though - I'll create an issue for that.

alexlyp:
OK that empty wallet issue should be fixed now.

Yah I was thinking about that.  I suppose we can immediately show the dcrd logs (if applicable) then show the dcrwallet logs once it's selected/loaded.  But yah we can do in a following PR

PR #1408, opened 2018-05-21
Add paste with middle click on seed restore

### LINKED ISSUE #1342: [launcher] Allow for middle-click seed paste from primary selection cache
Currently middle-clicking in the Restore wallet from seed does nothing on Mac/Linux.  It would be preferable to be able to paste the previously selected text by middle clicking. 

After doing some research/testing, it is simple enough to capture the middle-click event from onMouseUp/Down but it does not seem like it's possible to get the Primary Selection value.  

I believe this value should be accessible somehow, since middle-clicking in other fields does properly paste as expected.

### LINKED ISSUE #1407: [Restore Seed] Restore seed button is overlapped by seed words
The restore wallet buttons for switching between words and hex is overlapped by the place for writting the seed words.

vctt94:
closes #1407 and closes #1342.

I think would be nice having a tooltip or some kind of notification warning about the possibility of pasting with the middle click. What do you guys think?

### DIFF
--- a/app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
+++ b/app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
@@ -20,13 +20,17 @@ class ExistingSeedForm extends React.Component{
 
   handleOnPaste = (e) => {
     e.preventDefault();
-    const lowercaseSeedWords = SEED_WORDS.map(w => w.toLowerCase());
     const clipboardData = e.clipboardData.getData("text");
-    const words = clipboardData
-      .split(/\b/)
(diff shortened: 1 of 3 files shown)


alexlyp:
The middle click appears to work, though I'm having trouble clicking on the private passphrase inputs now.  Also after the middle click the first box remains open and awaiting for a selection to be made, not great UX.

I don't think it requires any warning, IMO

vctt94:
Thanks for reviewing my PR.

I am making the options from react select miss now. I tried focusing on the passphrase input, but the values from the seed were being pasted on the passphrase input and it seems a bit tricky working with blur on the react-select component, so I am just missing the options by its class.

Can you re-review and see how the UX is now?

alexlyp:
Cool seems to be better now.

PR #1419, opened 2018-05-24
Show transactions received while away

### LINKED ISSUE #911: On decrediton startup show transactions received while away
It would be nice to be able to have notifications appear about received txs/votes/revokes while decrediton was off on the next subsequent decrediton load.

Here is a possible way to solve this issue:  Record in config: the height of the last transaction confirmed in the wallet.  On reload show any more recent transaction from the one in the config, then update the height accordingly.

vctt94:
fixes #911 

I removed the cleanShutdown method from before-quit, because it calls for the second time, which gives a connection error when trying to get the block height.

 Is there a reason for cleanShutdown being called there?

I am just adding these recent transactions to the selector, we still need to show it in a component.

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -193,10 +193,11 @@ export const getTransactionsSinceLastOppened = () => async (dispatch, getState) 
   };
   const config = getGlobalCfg();
   const lastBlockHeightSeen = config.get("last_height");
-  const { currentBlockHeight, walletService } = getState().grpc;
+  const { currentBlockHeight, walletService, recentTxSinceLastOpenedCount } = getState().grpc;
 
   try {
-    const { mined, unmined } = await walletGetTransactions(walletService, lastBlockHeightSeen, currentBlockHeight);
+    const { mined, unmined } =
+      await walletGetTransactions(walletService, lastBlockHeightSeen, currentBlockHeight, recentTxSinceLastOpenedCount);
     const transactions = [ ...mined, ...unmined ];
 
     transactions.forEach( tx => {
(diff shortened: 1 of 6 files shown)


alexlyp:
Not sure why you made the changes to checkDaemon here.  Those should be done in a seperate PR.  

And it looks like the setLastHeight is in the wrong place.  It should be set by taking the currentBlockHeight from state and setting it in shutdownApp in DaemonActions.  Obviously, probably good to do some sanity checks to not set bad numbers or anything.

vctt94:
Thanks for reviewing my PR.

I made those changes here because I was getting the blockheight with getblockcount, but the way you said looks better.

I am setting the height on DaemonActions now and I am only checking if the current block height exists, before setting it into config. Do you have other sanity checks in mind that would be good having? Maybe check if `currentBlockHeight` is a number?

Also I am not removing the cleanShutdown from the before-quit anymore, but I am still seeing the connection error on the second call, even on master, when I am using the remote daemon. Should we have an issue for that?

connectionbug

vctt94:
Thanks @matheusd. 

I agree with you and now I am returning only the 10 first transactions.

app/actions/ClientActions.js
```diff
@@ -177,6 +178,48 @@ export const getWalletServiceAttempt = () => (dispatch, getState) => {
...
+    ticketsRevoked:       [],
+    totalReward:           0,
+    totalDCR:              0,
+  };
+  const config = getGlobalCfg();
+  const lastBlockHeightSeen = config.get("last_height");
+  const { currentBlockHeight, walletService, recentTxSinceLastOpenedCount } = getState().grpc;
+
+  try {
+    const { mined, unmined } = await walletGetTransactions(walletService, lastBlockHeightSeen, currentBlockHeight);
```
matheusd: You need to pass `recentTxSinceLastOpenedCount` here.

matheusd: Otherwise, you're still getting the same amount of txs from the wallet, only you're limiting on the client side.

vctt94: Sorry about that. Now I am doing it properly.

PR #1429, opened 2018-05-28
Calculate startup stats backwards

matheusd:
This switches the startup statistics calculation to happen backwards (from current balance back to -15 days) instead of forwards from the start of the wallet.

This speeds up startup time for wallets with a large number of transactions, as we now only need to obtain transactions up to about 15 days in the past.

The trickiest things to make this work is accounting for immature balances and ticket amounts (when calculating forwards, we can also count on having the ticket before the vote, so when going backwards we need to be able to grab individual ticket data given the vote tx).

I had to update the grpc api to the latest one from dcrwallet because I needed the block hash on the `GetTransaction` call.

One way to verify that the balances are correct is the following: export the normal balances csv, then change line [statisticsActions:217](https://github.com/decred/decrediton/compare/master...matheusd:backwards-stats?expand=1#diff-cc9008396c1c91c570339202d627d5d3R217) to `true` (so the report is always generated backwards) then export it again. They should be identical (modulo sorting order).

### DIFF
--- a/app/actions/StatisticsActions.js
+++ b/app/actions/StatisticsActions.js
@@ -30,8 +30,6 @@ export const getStartupStats = () => (dispatch) => {
       // changes with the previous balance, taking care to set sent/received
       // balances to 0
       dailyBalances = dailyBalances.data.slice(0, 15).reverse();
-      //dailyBalances = dailyBalances.data.slice(-15);
-      console.log(dailyBalances);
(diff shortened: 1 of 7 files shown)


alexlyp:
Can you bump the requiredVersion as well.  Should be 4.38.0 in index.js, looks like it's still at 4.27.0

PR #1430, opened 2018-05-28
Add config option to disable animations

matheusd:
This adds a config option (`ui_animations`) and implements the appropriate behavior to disable animations/transitions/motions when requested.

This is meant as an attempt to track down some recent bugs reported, that might be related to the animations being slow on older machines.

The option is currently not exported on the UI, since for the moment it is meant as a last resort alternative in older machines.


matheusd:
Sorry for that one!

PR #1436, opened 2018-05-31
[security] Update security page styling

### LINKED ISSUE #1411: Visual Update: Security Center
Preview: https://xd.adobe.com/view/865086b6-3db6-4273-bc22-779c0b15be85/
Specs: https://xd.adobe.com/spec/7c824c6d-257c-4998-7fcc-edc7e3bfaba6-0ab3/

- Re-arranged to security center layout into a single view instead of 3 subpages. 
- Sign Message and Verify Message are switched between from the tabs. 
- Buttons copy should update accordingly to each tab, displaying `Sign Message` or `Verify Message`
- Updated the info modal to common style (to be addressed in a few other views, separate issues). 
- Designed the validate address module. This would need a new field type defined, that has: 
1. Default state (during no interaction, same as others) 
2. Active state (dark grey) only triggered when active on and while the entered address isn't complete (e.g. in a case when the address is typed one letter at a time). Though when de-activated with incomplete entered address, turns to invalid state.  
3. Valid/Owned Confirmation (blue underline) – entered address is valid, owned. Triggers as an active state when the complete address is detected (e.g. pasted in) as well continues when de-activated (clicked elsewhere, address not deleted).
4. Valid/Not Owned (green underline) - entered address is valid, not owned. Interactions same as 3. 
5. Invalid - entered address is invalid. Triggers while active if a complete entered address is incorrect (e.g. pasted in), remains displayed when de-activated if field not cleared. Triggers when de-activating field if entered address is partial. 

With states 3,4,5, a relevant feedback icon note should display next to it: 
[decrediton - address feedback icons.zip](https://github.com/decred/decrediton/files/2028222/decrediton.-.address.feedback.icons.zip)

Nb. I've got a small update to the fields coming up, there's a common issue of un-designed warnings triggereing too soon (as in this view currently). Those will be addressed with a new error state for the fields as well some design related bug fixes.

alexlyp:
Close #1411


app/style/SecurityCenterMessagePage.less
```diff
@@ -126,4 +185,4 @@
     background-size: 30px;
     background-repeat: no-repeat;
   }
-}
+}
```
matheusd: Needs a new line here.

PR #1441, opened 2018-06-02
[Fix #1428] - Increment *autoHideDuration* for Snackbar

### LINKED ISSUE #1428: Error: Unable to purchase tickets- message box disappears too quickly to read
When trying to purchase a ticket, I get the error message stating "unable to purchase tickets." The message box has a lot of text but does not stay up long enough for me to read it all. It disappears after roughly three seconds. I'd like it to stay up longer.

adigunhammedolalekan:
Issue #1428 stated that message box disappears too quickly, this is due to Snackbar **autoHideDuration**  config set to 4secs, this property has now been changed to 10secs.

### DIFF
--- a/app/components/Snackbar/index.js
+++ b/app/components/Snackbar/index.js
@@ -69,7 +69,7 @@ class Snackbar extends React.Component {
         className={snackbarClasses(message || "")}
         open={!!message}
         message={message ? <Notification {...message} /> : ""}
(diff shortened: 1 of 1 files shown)


alexlyp:
This was fixed in @matheusd's recent PR #1439, now if the user is hovering the snackbar it will no longer disappeear.

PR #1458, opened 2018-06-06
Add support to watching only wallet creation

### LINKED ISSUE #941: Support watching only wallets
This is likely to require multiple modifications in Decrediton and maybe a few changes to the dcrwallet end of the grpc channel.

This will useful after the statistics work, as it will allow to more easily monitor a cold wallet.

vctt94:
This PR closes #941 

As discussed with @matheusd, I believe we should have an issue to add the getmasterpubkey command to decrediton, as it is required to create a watching only wallet and not every user is familiar with cli.


app/actions/WalletLoaderActions.js
```diff
@@ -101,6 +101,24 @@ export const createWalletRequest = (pubPass, privPass, seed, existing) =>
       .catch(error => dispatch({ error, type: CREATEWALLET_FAILED }));
   };
 
+export const CREATEWATCHONLYWALLET_ATTEMPT = "CREATEWATCHONLYWALLET_ATTEMPT";
+export const CREATEWATCHONLYWALLET_FAILED = "CREATEWATCHONLYWALLET_FAILED";
+export const CREATEWATCHONLYWALLET_SUCCESS = "CREATEWATCHONLYWALLET_SUCCESS";
+
+export const createWatchOnlyWalletRequest = (extendedPubKey, pubPass ="") => (dispatch, getState) => {
+  dispatch({ type: CREATEWATCHONLYWALLET_ATTEMPT });
+  return createWatchingOnlyWallet(getState().walletLoader.loader, extendedPubKey, pubPass)
```
matheusd: Change this to `wallet.createWatchingOnly` as it's how we're trying to standardize usage of the wallet functions.

app/components/buttons/WatchOnlyWalletSwitch.js
```diff
@@ -0,0 +1,25 @@
...
+import { InvisibleButton } from "buttons";
+import { FormattedMessage as T } from "react-intl";
+import "style/EyeFilterMenu.less";
+import "style/StakePool.less";
+
+const WatchOnlyWalletSwitch = ({ enabled, onClick, className }) => (
+  <div className={className ? className : ""}>
+    <Tooltip text={enabled ? <T id="autobuyer.enabled" m="Watch only Wallet" /> : <T id="autobuyer.disabled" m="Not Watch only Wallet" />}>
+      <EventListener target="document">
+        <div className="teste">
```
matheusd: Why do you have this EventListener here? Also, change `className="teste"` to something more meaningful or remove altogether if not needed

app/components/views/GetStartedPage/WalletSelection/index.js
```diff
@@ -102,15 +114,38 @@ class WalletSelectionBody extends React.Component {
...
   createWallet() {
-    const { newWalletName, createNewWallet } = this.state;
+    const { newWalletName, createNewWallet,
+      isWatchOnly, masterPubKeyError, walletMasterPubKey } = this.state;
     if (newWalletName == "" ) {
       this.setState({ hasFailedAttempt: true });
       return;
     }
+    if (isWatchOnly) {
+      if(masterPubKeyError || !walletMasterPubKey) {
```
matheusd: Add a space after the `if`.

app/actions/WalletLoaderActions.js
vctt94: I thought we were using the same name from the WalletLoaderServiceService and the request name in there is `createWatchingOnlyWallet`. Is that wrong?

vctt94:
I am checking now if the opened wallet is watching only or not on the grpc call, but it only works on the first load, because when refreshing we get that the wallet is already loaded:
`Error: 6 ALREADY_EXISTS: loader.OpenExistingWallet: item already exists: wallet already opened`.

I believe we can use the config for persisting this info, but I believe it would be better done in a different PR so this one won't get too big. Also I added this info to selectors, we probably need an issue for deactivating functionalities as well.

app/actions/WalletLoaderActions.js
matheusd: Yeah, sorry should have been more accurate.

You can keep the `createWatchingOnlyWallet`, it's just that instead of importing and referencing it directly (`import { createWatchingOnlyWallet, ... } from "wallet" }`, use the wallet alias that imports everything (`import * as wallet from "wallet"`). 

We should eventually change all references to "dangling" functions to use the same style of module reference for consistency.

matheusd:
I think you can do this change on this PR, as this is actually required for the wallet to work correctly.

Another alternative, instead of storing this on the config, would be to store on the node.js stack by adding an ipc call to set and another to retrieve the watchingOnly status, so we'd rely on a single source of truth and prevent the config vs wallet from getting out of sync.

vctt94:
I like the idea to use the node stack with ipc, so I used it. Thanks for the tip. 

Also, now I am calling the methods by the wallet alias.

matheusd:
Could you double check if creation is working? I'm trying to create a watching only wallet on the lastest version and it gets stuck in the screen to choose between "Create New Wallet" and "Restore Existing Wallet"

vctt94:
What are you getting? For me wallet creation is working fine. It takes a while to fetch block headers, but it loads just fine.

matheusd:
Try the following:
Rename `~/.config/decrediton/wallets` to `_wallets`. Restart decrediton. See if you can create a new wallet (both non-watch-only and watch only).

This seems broken on master for me too, so not a problem on your PR. Just confirm whether you see this behavior or if it's my build that's broken.

vctt94:
Yes, I get stucked on the "Create New Wallet" and "Restore Existing Wallet" as well. It also happens on master as you said.

matheusd:
You need to change this line: https://github.com/decred/decrediton/blob/master/app/components/views/GetStartedPage/CreateWallet/CreateForm.js#L19

from `... || createNewWallet ?` to `... || !createNewWallet ?`

Then you need to fix the flow, such that when creating the watching only wallet it doesn't go through the page to create a new wallet or import an existing seed.

PR #1461, opened 2018-06-07
WIP: Chinese decrediton GUI translation

yinyunqiao:


alexlyp:
We only accept translations through transifex and are currently using a 3rd party service to provide those translations.  So this PR won't be necessary, thanks.

PR #1479, opened 2018-06-15
[stakepools] allow adding custom stakepool

### LINKED ISSUE #1464: [stakepools] Manually add a stakepool URL for API key entry
Due to various reason, we should allow users to use a custom stakepool URL when adding a stakepool or setting one up initially.  Once used, this should be added to the stakepool list like any other and should be selectable in the drop downs as expected.

To successfully be added you should get back a valid purchaseInfo response from the stakepool's api.

matheusd:
Close #1464 

Adds the ability to add a custom stakepool by typing its address in the stakepool select drop down, and selecting and option to add that stakepool.

Before adding the stakepool, a request is made to its `/api/v1/stats` endpoint, to check if it exists and get its available api versions. Only then is the stakepool added to the supported list.

Easiest way to test it is probably to run a local webserver (eg: via `php -S` or python simplewebserver) that returns a fixed json with the required fields (eg: copy from an existing stakepool).

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -129,7 +129,7 @@ const importScriptSuccess = (importScriptResponse, votingAddress, cb, willRescan
 export const importScriptAttempt = (passphrase, script, rescan, scanFrom, votingAddress, cb) =>
   (dispatch, getState) => {
     dispatch({ type: IMPORTSCRIPT_ATTEMPT });
-    return wallet.importScript(sel.walletService(getState()), passphrase, script, rescan, scanFrom)
+    return wallet.importScript(sel.walletService(getState()), passphrase, script, false, 0)
       .then(importScriptResponse => {
         if (rescan) dispatch(rescanAttempt(0));
         dispatch(importScriptSuccess(importScriptResponse, votingAddress, cb));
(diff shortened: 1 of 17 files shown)


alexlyp:
I'm trying to think of a better way to prove validity since `api/v1/stats` can sometimes take a really long time to respond.  I think that should be getting cached by the stakepools, but that's a bigger change.

Also, it doesn't seem too clear that one can enter a custom stakepool.  I think this is mostly due to the value in the stakepool drop down never clears until the typing begins.

Also seems that the clear value button shows up, but doesn't seem to do anything.

matheusd:
Updated to set clearable as false in the default case and added a pseudo-option to indicate you can add a stakepool by adding.

Re: using `api/v1/stats`, it's pretty easy to add a new endpoint to dcrstakepool to return only immediate information. Something like `api/v2/about`, which returns the api version, pool fee, network, etc (the needed information to add the stakepool).

The only problem with that approach is that we'd need to get stakepools to update their software (or only support those that support a new api version 3 with that new endpoint). If you think that's ok, I can implement the endpoint in dcrstakepool and test it out (I'm running my own testnet stakepool for the split ticket tests, so it's pretty easy to update it if needed).

alexlyp:
Yah I think this will do for now, and I suppose this feature will only be used pretty rarely so no need to implement a new endpoint for this one

alexlyp:
I like the new UI improvements, much better.  

Getting an error when trying to add a new stakepool on testnet when I already have teststakepool.decred.org configured.  Basically seems like it doesn't like having an empty list for that dropdown/select I think.

Issue #1490, opened 2018-06-25
[watch-only]  Various functionalities to hide while isWatchingOnly mode


alexlyp:
List of functionality to hide:
- Send DCR
- Tickets (since you can't import scripts/add stakepool)
- Add Accounts
- Change Private Passphrase
- Sign/Verify message

vctt94:
Can I get this one?

alexlyp:
You can start by hiding sign/verify, add accounts and change passphrase.  We need to get some guidance from @linnutee about what to do with the tabs that need to be hidden (send and tickets etc)

linnutee:
Prototype: https://xd.adobe.com/view/f62cab0f-de1d-4385-665c-cdea96b80679-801d/
Specs: https://xd.adobe.com/spec/563119f3-5bbf-43bd-5c65-6b002287f541-2000/
Icons: [decrediton - icons 200718.zip](https://github.com/decred/decrediton/files/2215093/decrediton.-.icons.200718.zip)


Included a design example to address this with disabled states and masking rather than completely hiding. This way neither neither the layout breaks and there's a bit more information available for the user. 
- Full views that have no available functionality (send and tickets) have the content area masked and display a modal (without close option) explaining why the view is disabled. 
- Views that have partial functionality (accounts, settings and security center) either the fields or buttons (or both) in inactive state. 
- When clicking on those states, should trigger a notification with the copy "This functionality is disabled for watch-only Wallets". I didn't do a mock-up for the notification, can just use the current snackbar how it is. It's a further item in the pipeline to do an overhaul for the notifications. 
- Disabled buttons already got an existing style element. Inactive fields on the other hand I think have been mixed up in the past with filled/completed fields to some degree in some instances. Reference from https://github.com/decred/decrediton/issues/1504 and component sheet: https://xd.adobe.com/spec/1c4345df-59aa-4dca-595d-0f82e91ecbaa-6a72/screen/dd46576b-2b5b-4ad3-acc8-80873d6f8b92/UI-component-states/
- Included a disabled icon for the password button and also updated (color) icon for the info modal (should be applied through-out).

matheusd:
Small note: message verification works for watch only wallets, so don't disable that (you can verify messages from any address, not necessarily from your wallet).

PR #1513, opened 2018-06-29
[multi] Hide watching only functionalities

vctt94:
starts #1490 

This PR hides some functionalities when in watching wallet only mode. 

- sign/verify message
- new account button
- change passphrase button

We still need to figure what to do with tabs that need to be hidden.

I believe hiding the tabs would be fine, if we have a message showing that the wallet is not fully functional because of the watching only mode.

### DIFF
--- a/app/components/views/AccountsPage/Accounts/List.js
+++ b/app/components/views/AccountsPage/Accounts/List.js
@@ -67,6 +67,7 @@ AccountsList.propTypes = {
   onShowAccountDetails: PropTypes.func.isRequired,
   onHideAccountDetails: PropTypes.func.isRequired,
   accountNumDetailsShown: PropTypes.number,
+  isCreateAccountDisabled: PropTypes.bool.isRequired,
 };
 
 export default AccountsList;
--- a/app/components/views/AccountsPage/Accounts/List.js
+++ b/app/components/views/AccountsPage/Accounts/List.js
@@ -5,19 +5,21 @@ import { DecredLoading } from "indicators";
 import { InfoDocModalButton, PassphraseModalButton } from "buttons";
 import { AddAccountModal } from "modals";
 
-const AccountsListHeader = ({ onGetNextAccountAttempt }) =>
-  <StandaloneHeader
+const AccountsListHeader = ({ onGetNextAccountAttempt, isCreateAccountDisabled }) =>{
+  return  <StandaloneHeader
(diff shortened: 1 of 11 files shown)


matheusd:
I thought a little bit about this. I believe we should use individual selectors for enabling/disabling individual features, since the availability of each individual feature/page/action will possibly depend on lots of factors.

Eg: sending funds is disabled on a watch only wallet, but must be enabled on a trezor wallet when that wallet is connected (and possibly disabled when disconnected). Purchasing tickets as well, though it might take longer to support on trezor, so we may need to selectively enable sending txs but not purchasing tickets, etc.

So, instead of using `isWatchOnly` throughout the code, we should use feature selectors like `sendEnabled`, `featSendEnabled`, `isSendingEnabled`, `isFeatSendEnabled` or something along those lines.

What do you guys think? We can further discuss this on chat as well.

vctt94:
I like your idea!

I believe we can still have flags for which mode is activate and just pass an object through a selector with the keys we would enable or disable the functionalities.

For example, we would have `isWatchingOnly` and add what would be disabled like `sendDisabled`, `createAccount`, etc... 

This way I believe would be easier for supporting trezor, spv, etc. 

What do you think?

matheusd:
I'm not sure I'm following you, but the general gist I'm thinking is something to the effect of:

In selectors.js:
```
const isWatchingOnly = get(["grpc", "isWatchingOnly"]);
const isTrezor = get(["grpc", "isTrezor"]);
const trezorModelVersion = get(["grpc", "trezorModelVersion"]); // false, 1=model one, 2=model T, etc
const isSendDisabled = and(not(isTrezor), isWatchingOnly); // or something like that
```

Then pass `isSendDisabled` along a connector to the appropriate places.

vctt94:
I see, I believe what you say makes more sense. 
I thougth on something like 
```
export const isWatchingOnlyDisabled = {
  send,
  createAccount,
  ...
} 
```
But we would hit the same problem to pass variables through the code.

I'll implement the way you said, as I think it makes more sense

alexlyp:
I got 2 errors after creating a watch only wallet and going to the accounts page:
isChangePassPhraseDisabled is required but undefined.

Looks like on new watch only wallet creation those aren't set on startup.

alexlyp:
Otherwise looks good.

vctt94:
fixed

PR #1536, opened 2018-07-16
promisify stakePoolApi

vctt94:
I started promisifying the stakePoolApi.

As axios is already promisified, I am not sure the need to promisify the requests as well.
I like the way it is because the requests are centralized at the middleware,
what do you guys @matheusd and @alexlyp think?

Thanks.


alexlyp:
yah looks ok, go ahead and do the others.  Thanks!

PR #1538, opened 2018-07-17
Agrouping messages so it only shows once

### LINKED ISSUE #1534: Batch snackbar message display
When many messages are generated at once (eg: buying many tickets in one go) the UI gets slow with all individual messages being shown.

There needs to be a small batching delay before displaying messages (eg: 250ms) where all messages are aggregated such that only the last one is displayed.

vctt94:
Fixes #1534 

Right now I am Removing messages that are the same or messages that have the same `txHash` so them only show once.

### DIFF
--- a/app/components/Snackbar/index.js
+++ b/app/components/Snackbar/index.js
@@ -68,12 +68,14 @@ class Snackbar extends React.Component {
   }
 
   checkIsSameMessage(messageObj, oldMessageObj) {
-    if(messageObj === oldMessageObj)
+    if (messageObj === oldMessageObj) {
       return true;
-    if(!messageObj || !oldMessageObj)
+    }
+    if (!messageObj || !oldMessageObj) {
       return false;
+    }
     const { type, message } = messageObj;
-    if(type !== oldMessageObj.type) {
+    if (type !== oldMessageObj.type) {
       return false;
     }
     // message can be a FormattedMessage from react-intl or a transaction
--- a/app/components/Snackbar/index.js
+++ b/app/components/Snackbar/index.js
@@ -33,14 +33,10 @@ class Snackbar extends React.Component {
   constructor(props) {
     super(props);
     this.hideTimer = null;
-    this.newMessage = null;
-    this.timeoutRunning = null;
     this.state = {
       message: props.messages.length > 0
         ? props.messages[props.messages.length-1]
         : null,
-      delay: 250, // delay in ms to render component
-      isWaiting: true,
     };
   }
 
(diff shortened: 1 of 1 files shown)


app/components/Snackbar/index.js
```diff
@@ -30,17 +30,23 @@ const snackbarClasses = ({ type }) => ({
...
-        : null
+        : null,
+      delay: 250, // delay in ms to render component
+      isWaiting: true,
     };
   }
 
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({isWaiting: false}); }, this.state.delay);
```
matheusd: This doesn't work as you intend it to work.

The snackbar component only gets mounted once (since it's added to the App container). You can easily verify this by adding a console message inside the setTimeout() callback: isWaiting only gets set to false once.

app/components/Snackbar/index.js
```diff
@@ -30,17 +30,23 @@ const snackbarClasses = ({ type }) => ({
...
+  }
+
   componentWillReceiveProps(nextProps) {
```
matheusd: You need to add the delay here (a bit below actually, after the component has determined this is a new message). 

You need to store the new message in an intermediate attribute of the class (*not* the state yet, so we don't trigger a render) and start the 250ms timer here (if it's not already running - this is important). Once the timer callback is called, then you update the state with the new message (which will trigger the render).

If there are multiple messages being received within the 250ms window (which you'll detect if `componentWillReceiveProps` gets called while the timer is set), your options are:

1. Discard the new messages and display the first one (and keep the timer running)
2. Discard the old one and display the most recent (and keep the timer running)
3. Discard the old one and display the most recent (but restart the timer)

vctt94:
I see what you said. I believe now it is working properly. The repeated messages are not showing more than once, but we still show each of the ticket transaction, is that fine?

app/components/Snackbar/index.js
```diff
@@ -127,6 +137,8 @@ class Snackbar extends React.Component {
   }
 
   render() {
+    if (this.state.isWaiting)
+      return null;
```
matheusd: Small nit. Try not to use this style of if statement (statement in the following line without curly braces). It is notorious for some nasty bugs (eg: https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/)

matheusd: Use either in the same line which doesn't cause ambiguity or add curly braces.

vctt94:
fixed

app/components/Snackbar/index.js
```diff
@@ -30,14 +30,17 @@ const snackbarClasses = ({ type }) => ({
...
     this.hideTimer = null;
+    this.newMessage = null;
+    this.timeoutRunning = null;
     this.state = {
       message: props.messages.length > 0
         ? props.messages[props.messages.length-1]
-        : null
+        : null,
+      delay: 250, // delay in ms to render component
+      isWaiting: true,
```
matheusd: You can drop `isWaiting` (or try again to modify the logic).

Notice that isWaiting is only set to true in one place: the constructor. Therefore, after the first message is received it will never be true again,

vctt94:
Yeah, you were right, so I changed the logic. I noticed that the message verifying is always false as well, so it enters that verification every run, what I believe causes the many messages to show. 

I fixed how messages are compared and I believe it should work properly now, can you give me your feedback?

app/components/Snackbar/index.js
```diff
@@ -30,30 +30,63 @@ const snackbarClasses = ({ type }) => ({
...
+    if(messageObj === oldMessageObj)
+      return true;
+    if(!messageObj || !oldMessageObj)
+      return false;
+    const { type, message } = messageObj;
+    if(type !== oldMessageObj.type) {
+      return false;
+    }
+    // message can be a FormattedMessage from react-intl or a transaction
+    if (message.defaultMessage !== oldMessageObj.message.defaultMessage) {
```
matheusd: `if` with following space and brances

All within the range of 7 lines :)

app/components/Snackbar/index.js
```diff
@@ -30,30 +30,63 @@ const snackbarClasses = ({ type }) => ({
...
+    return isSame;
+  }
+
+  checkIsSameMessage(messageObj, oldMessageObj) {
+    if(messageObj === oldMessageObj)
+      return true;
+    if(!messageObj || !oldMessageObj)
+      return false;
+    const { type, message } = messageObj;
+    if(type !== oldMessageObj.type) {
```
matheusd: `if` without following space and with braces

app/components/Snackbar/index.js
```diff
@@ -30,30 +30,63 @@ const snackbarClasses = ({ type }) => ({
...
+        return;
+      }
+    });
+    return isSame;
+  }
+
+  checkIsSameMessage(messageObj, oldMessageObj) {
+    if(messageObj === oldMessageObj)
+      return true;
+    if(!messageObj || !oldMessageObj)
```
matheusd: `if` without following space and without braces

vctt94:
fixed. 

I give a look at putting the code at the selector, but I believe it will not be much of a difference.

PR #1542, opened 2018-07-19
Added passphrase keyboard entry option

### LINKED ISSUE #646: Implement keylogger resistant passphrase entry UI
In order to protect the user from keyloggers stealing their passphrase, implement a virtual keyboard in addition to the default input field.  By default, the input field should be active and the user should have the ability to select **I use an untrusted computer** to activate the virtual keyboard.

image

The virtual keyboard might be implemented using [react-screen-keyboard](https://www.npmjs.com/package/react-screen-keyboard)

Related to #639

Jujhar:
This adds a popup keyboard for when inputting a passphrase, used to prevent keylogger tracking threat.

<b>Please test Import Script  passphrase, I could not find where that was.</b>

This should show up in:
* Buy Tickets / Add a Stakepool (from Tickets)
* Add Account / Add new (from Accounts)
* Update Private Password (from Settings)
* Import Script Module

It works fine in Buy Tickets and Add account. I disabled it on Update Private Password.

Fixes #646


alexlyp:
This was discussed thoroughly, and we came to conclusion that we don't want a software keyboard.  Closing.

PR #1545, opened 2018-07-23
Prevention of third click (mouse wheel) opening new window open

### LINKED ISSUE #1537: [Sidebar] Middleclick on menu tabs open a new decrediton window
When clicking on menu tabs with middle click, decrediton opens a new window.

We should prevent this event.

Jujhar:
By default doing a third click on something on the sidebar opens a new electron window however with this fix that action is stopped.

Fixes #1537

### DIFF
--- a/app/components/views/HomePage/index.js
+++ b/app/components/views/HomePage/index.js
@@ -66,21 +66,4 @@ class Home extends React.Component{
 
 }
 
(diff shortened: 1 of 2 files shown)


Jujhar:
Updated it into that file with example pattern, should be ready to merge.

PR #1546, opened 2018-07-24
Add event listener to submit send transaction form on enter key press

### LINKED ISSUE #1221: Enter key shortcut in 'Send trx' dialog
Pressing Enter does not currently shortcut to the default Send button:

image

shweini:
This enables the Enter key shortcut on the send transaction page such that when e enter key is pressed, it defaults to the send button if the transaction is valid

This fixes #1221

### DIFF
--- a/app/components/buttons/ModalButton.js
+++ b/app/components/buttons/ModalButton.js
@@ -12,6 +12,7 @@ class ModalButton extends React.Component {
 
   showModal() {
     this.setState({ show: true });
+    this.onShow();
   }
 
   hideModal() {
@@ -22,6 +23,17 @@ class ModalButton extends React.Component {
     const { onSubmit } = this.props;
     this.hideModal();
     onSubmit && this.props.onSubmit(...args);
+  }
+
+  onShow() {
+    const { onShow } = this.props;
+    onShow && this.props.onShow();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.showModal) {
+      this.showModal();
+    }
   }
 
   render() {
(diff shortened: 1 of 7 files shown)


app/components/views/TransactionsPage/SendTab/index.js
```diff
@@ -225,6 +228,15 @@ class Send extends React.Component {
     publishUnminedTransactions && publishUnminedTransactions();
   }
 
+  onKeyDown(e) {
+    if (e.keyCode === 13 && this.getIsValid()) {
+      var b = document.getElementsByClassName("content-send");
```
matheusd: This breaks react's patterns of not directly using dom references. You need to somehow propagate the enter event up the stack until it reaches the proper place.

The recently merged PR #1540 changed how transactions are signed and published to account for future trezor integration, so take a look at that (specifically at the new SendTransactionButton). You need to be able to propagate the event of the enter key up to that button using react idioms.

shweini: fixed

app/components/buttons/ModalButton.js
```diff
@@ -24,6 +25,17 @@ class ModalButton extends React.Component {
     onSubmit && this.props.onSubmit(...args);
   }
 
+  onShow() {
+    const { onShow } = this.props;
+    onShow && this.props.onShow();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.showModal) {
```
matheusd: Probably you want to change this to `if (nextProps.showModal && !this.props.modal)` so that modal showing is triggered only when the showModal props is changing. Otherwise, any other prop changing can trigger the modal to be shown again while showModal is true.

shweini: While i wasnt able to reproduce the error, changes have been effected where errors were found.

PR #1557, opened 2018-07-27
[various] Changes to allow new dcrwallet SPV mode

alexlyp:
~This PR requires https://github.com/decred/dcrwallet/pull/1221 binary to be used.~

The new SPV mode in dcrwallet uses a simple all-inclusive startupSync method that does everything required to prepare the wallet for usage while in SPV mode.  This is accessed by the new rpc `SpvSync`.  This creates a new response stream that will stay open for the life of the dcrwallet process, is canceled or has an unrecoverable error.  

This PR also includes updates and refinements to the existing normal mode startup procedures.  A new call for `FetchMissingCFilters` has been added.  Also we now rely on `RescanPoint` requests to tell us where the last rescan had ended so we can consistently pick up from there.  There was also some juggling of startupSteps that took place to further correct and smooth out this process.

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -36,7 +36,11 @@ function getWalletServiceSuccess(walletService) {
 
 export function startWalletServices() {
   return (dispatch, getState) => {
-    setTimeout(() => { dispatch(loadActiveDataFiltersAttempt()); }, 1000);
+    const { walletCreateExisting, walletCreateResponse, rescanPointResponse, spvSynced } = getState().walletLoader;
+    if (!spvSynced) {
+      dispatch(loadActiveDataFiltersAttempt());
+      setTimeout(() => { dispatch(getTicketBuyerServiceAttempt()); }, 1000);
+    }
     setTimeout(() => { dispatch(getNextAddressAttempt(0)); }, 1000);
     setTimeout(() => { dispatch(getTicketPriceAttempt()); }, 1000);
     setTimeout(() => { dispatch(getPingAttempt()); }, 1000);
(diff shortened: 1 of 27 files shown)


matheusd:
Testing it out now. I'm going to dump all problems here at first, though there might be bugs that are actually from the dcrwallet PR.

- I got a RESOURCE_EXHAUSTED when restoring a wallet for the first time (see below for full message). This might be the origin of the next bug. This didn't happen when I created a new wallet (from seed)
- Wallet tx notification wasn't working right after opening the restored wallet (tried sending from the faucet and transfering between accounts). Had to close and reopen to get notifications.
- I got flashed the "discover accounts - enter passphrase" page when opening the restored wallet for the second time
- Dcrd is still getting started. I'm guessing adding the option to start/not start it will be a future PR?
- Autobuyer is losing configuration after changing pages 
- Migrated wallet worked without a hitch. Yay! :)

*****
Transactions ntfns error received: Error: 8 RESOURCE_EXHAUSTED: Received message larger than max (17055314 vs. 4194304)
I can provide the seed to the wallet generating this if needed.

matheusd:
spvconnect doesn't seem to be actually filled or exposed anywhere, but I'm still only grabbing headers/filters from the local dcrd node

alexlyp:
Strange I'm unable to replicate the RESOURCE_EXHAUSTED error.  I tried to upgrade a wallet from 4.40.0 to both normal mode and spv on 5.1.0 and they both successfully completed.  Are you able to reliably reproduce it?  If so, then ya that would be great to get that seed for testing.

I'll be addressing the discover address form screen flash.

matheusd:
Just tested with the newest versions of the wallet and decrediton branches and the resource exhausted doesn't seem to be happening anymore. I'll keep an eye on it if it starts again.

alexlyp:
@matheusd do you have the dcrwallet logs for that RESOURCE_EXHAUSTED error?

matheusd:
Sorry, I had already deleted the wallet/log before looking at this. I'll make a point to record it next time I see it.

PR #1563, opened 2018-07-30
[watch-only] Add watching only warning modal

vctt94:
This PR is a continuation of #1490. It adds a tab modal, which still shows the tabs, bluring only the tabbed page body. 

The react-motion is making the modal control harder, because the component only unmounts after the effect is done, which is causing the modal to stay on the screen, even after a route change. 

I solved that by adding a `tabModalHidden` when changing routes on tab. But I still have two problems with this solution: 

- When switch between tabs and goes back to a tab which should show a modal quickly enough, as the component does not have time to unmount, it does not mount again. Therefore, it does not call the `tabModalShown` and do not show the modal. 
- When going from a disabled tab to another disabled tab, the modal misses after showing for a while, because the older modal gets unmounted.

I believe using the mount and unmount lifecycle to control this tab modal, is going to get too complicated. 
So I am thinking on maybe have a different component for that kind of modal, what do you guys think?


matheusd:
What about *not* using a modal, and just adding a properly styled component in the purchase tickets page when the feature is disabled? You'd just add a div with a classname as the overlay blur after the page (but inside it in the DOM hierarchy) and then the message "pseudo-modal" after that.

vctt94:
Thanks @matheusd for your suggestion.
It became much simpler than add one more reducer state the way you said and it is working fine now, I believe.

There is still some functionalities to add to finish #1490, but I believe this PR will get too big if I do everything on it. What do you think?

vctt94:
rebased

PR #1565, opened 2018-08-01
Updated Decred input field

### LINKED ISSUE #1511: [inputs] Implement amount design as specified in the docs
Specs: https://xd.adobe.com/spec/1c4345df-59aa-4dca-595d-0f82e91ecbaa-6a72/

The Input B specifications need to be implemented for the send page amount input (and possibly other places as well.)  

This is a bit more than just a styling change due to needing to parse and change the input actively while the user is entering in the information.  And I'm not 100% that it can be done.  I'm not aware of a way to split apart portions of an input to which some will get differing styling.

Jujhar:
On Transactions -> Send DCR the input field is formatted in such a way that that everything before the first 3 main digits after the decimal point appear bigger than rest of digits.

Validation should work. (eg your amount is too low)

Rounding is not perfect it can go up or down one minor decimal point but some code has been implemented to fix a deliberate attempt to through it off.

Fixes #1511

### DIFF
--- a/app/components/inputs/InputComponent.js
+++ b/app/components/inputs/InputComponent.js
@@ -53,7 +53,6 @@ class Input extends React.Component {
     document.getElementById("parsedInputDisplay").innerHTML = number;
     document.getElementById("parsedInputDisplay").style.display = "inherit";
     document.getElementById("viewInputDisplay").style.display = "none";
-    document.getElementById("parsedInputDisplay").focus();
   };
   onChange2 = () => {
     var number = document.getElementById("parsedInputDisplay").innerHTML;
@@ -150,11 +149,11 @@ class Input extends React.Component {
               placeholder={placeholder}
               value={isNullOrUndefined(value) ? "" : value}
               onChange={this.sendToParser}
-              onFocus={this.onChange}
+              onMouseOver={this.onChange}
               onBlur={this.onChange2}
               onKeyDown={this.onKeyDown2}
             >
-              <span id="parsedInputDisplay" style={{ display: "none" }}
+              <span id="parsedInputDisplay" style={{ display: "none", marginTop: "2px;" }}
                 onKeyDown={this.onKeyDown2}
               ></span>
               <div id="viewInputDisplay">
--- a/app/components/inputs/InputComponent.js
+++ b/app/components/inputs/InputComponent.js
@@ -43,9 +43,18 @@ class Input extends React.Component {
     onKeyDown && !e.defaultPrevented && onKeyDown(e);
   }
   onKeyDown2 = (e) => {
-    if (e.keyCode === 13) {
+
+    // filter non numeric keys, below allows arrow keys and decimal
+    if (e.keyCode == 8 || (e.keyCode >= 37 && e.keyCode <= 40) ||
+        e.keyCode == 46 || e.keyCode == 190 || e.keyCode == 110) {
+
+      // continue with flow
+    }
+
+    else if (e.keyCode === 13 || e.keyCode < 48 || e.keyCode > 57) {
       e.preventDefault();
     }
+
     const { onKeyDownSubmit, onKeyDown } = this.props;
     (e.keyCode === 13) && onKeyDownSubmit && onKeyDownSubmit(e);
     onKeyDown && !e.defaultPrevented && onKeyDown(e);
@@ -55,6 +64,7 @@ class Input extends React.Component {
     var number = this.state.inputMain + this.state.inputSub;
     this.setState({ inputVal: number });
   };
+
   onChange2 = (e) => {
     this.setState({ inputPhase: 0 });
     var number = e.target.textContent;
@@ -146,6 +156,7 @@ class Input extends React.Component {
               onFocus={this.onChange}
               onBlur={this.onChange2}
               onKeyDown={this.onKeyDown2}
+              suppressContentEditableWarning = {true}
             >
               <span id="parsedInputDisplay" style={{ marginTop: "2px" }}
                 className={"parsedInput-" + this.state.inputPhase}
--- a/app/components/inputs/InputComponent.js
+++ b/app/components/inputs/InputComponent.js
@@ -11,7 +11,11 @@ class Input extends React.Component {
   getInitialState() {
     return {
       inputUnitDiv: null,
-      divClassName: "input-and-unit " + (this.props.className || "") + (this.props.disabled ? " disabled " : "")
+      divClassName: "input-and-unit " + (this.props.className || "") + (this.props.disabled ? " disabled " : ""),
+      inputPhase: 0,
+      inputVal: "0.003000000",
+      inputMain: "0.00",
+      inputSub: "0000000"
     };
   }
   componentDidMount() {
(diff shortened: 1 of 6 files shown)


alexlyp:
Overall works pretty well.

Couple things I noticed:
- There is a slight (1-2px) shift up when clicking on the field when entering an amount.
- The field that you actually change for the amount should be given focus on click.  Right now seems like you have to click twice to start changing the amount.

Jujhar:
Fixed the things you mentioned I changed it to change wben mouseover rather than mouse click. By the way it is an span field so it can't be focused.

app/components/inputs/InputComponent.js
```diff
@@ -0,0 +1,194 @@
...
+    // Send to other older previous unstyled component and trigger event
+    document.getElementById("parsedInput").value = number_f;
+    var event = new Event("input", { bubbles: true });
+    var x = document.getElementById("parsedInput");
+    var tracker = x._valueTracker;
+    tracker.setValue("'" + x.value + "'");
+    x.dispatchEvent(event);
+
+    document.getElementById("parsedInputDisplay").style.display = "none";
+    document.getElementById("viewInputDisplay").style.display = "inherit";
```
matheusd: NAK on this. As a general rule in react projects, if you're writing code that uses `getElementById`, you're doing it wrong. 

You need to pass a semantic state or prop value to the component, then on the render() method you set the style (or how we usually do it, a className) to the components being rendered.

app/components/inputs/InputComponent.js
```diff
@@ -0,0 +1,194 @@
...
+            </div>
+          ) : null}
+        </Aux>
+    );
+  }
+}
+
+function pad(str, max) {
+  str = str.toString();
+  return str.length < max ? pad("0" + str, max) : str;
```
matheusd: Using recursion for a simple pad function seems like overkill to me. You can also just use `str.padStart(max, "0")` instead of having to write this pad function.

app/components/inputs/InputComponent.js
Jujhar: Not picturing it if its not too much work can you give an example by doing it, would help out understanding a lot by exactly knowing how to do it in future.

matheusd: Here's one of the simplest examples I could find in our codebase: 

https://github.com/decred/decrediton/blob/master/app/components/shared/CopyToClipboard.js

It uses both props and state to change the css classes applied to html elements.

Jujhar: Ah I see thank you, updated it.
Note I had to leave two instances to mirror the changes for validation and submit event.

app/components/inputs/InputComponent.js
Jujhar: Updated

app/components/inputs/InputComponent.js
```diff
@@ -0,0 +1,188 @@
...
+    if (isNaN(left_part) == true || isNaN(right_part) == true) {
+      left_part = "0.00";
+      right_part = "000000";
+    }
+
+    this.setState({ inputMain: left_part });
+    this.setState({ inputSub: right_part });
+
+    // Send to other older previous unstyled component and trigger event
+    document.getElementById("parsedInput").value = number_f;
```
matheusd: Still using getElementById...

Jujhar: That one section is to mirror the changes for validation and submit event. The thing I made is a different element that links to this original input and this part sends the new element changes to the old element. I am unsure how to change this because of the 'value' variable of the original input (<input value={value}) not sure where exactly that 'value' variable comes from and how it is changed and how the validation works with this 'value'.

matheusd: The `value` attribute being passed to the original `<input>` element comes from whoever is using this input component: you can see that it's coming from the props being passed to this component (I'm going to mark it with a comment so it can be obvious).

(apologies if you already know what I'm about to show; it's just that it seems you're not so experienced in React patterns)

It's common in React applications to have an input lifecycle that roughly looks like this:

```javascript
class ComponentInput extends React.Component {
  render() {
    return <input value={this.props.value} onChange={e => this.props.onChange(e.target.value)} />
  }
}

class ComponentThatUsesInput extends React.Component {
  onChage(newValue) {
    this.setState({ value: newValue });
  }

  render() {
    return <ComponentInput value={this.state.value} onchanve={this.onChange} />
  }
}

```
Notice that the content of `ComponentInput` (the contents of the `value` prop) is actually controlled by the _user_ of `ComponentInput` (instances of `ComponentThatUsesInput` class). 

Another way of looking at this is that `value` is a _state_ of `ComponentThatUsesInput` while it's a _prop_ of `ComponentInput`.

Yet another way of looking is that the _owner_ of whatever is in `value` is `ComponentThatUsesInput`, while `ComponentInput` merely displays its content and signals intentions to modify it (the `onChange` event) up the stack.

If you look at Decrediton's original `Input` component (and any place where it's used, such as the [message text input](https://github.com/decred/decrediton/blob/master/app/components/views/SecurityPage/VerifyMessage/Form.js#L77)) you'll see that it's following this pattern.

When you have a situation where you need to do some intermediate processing in a component, what you need to do is modify this "content propagation loop" such that the lower component processes the content then upstreams the modified value. For example, let's say that you want to always prefix the value with spaces, you'd modify the above code in the following way:

```javascript
class ComponentInput extends React.Component {
  onChange(newValue) {
    this.props.onChange("    "+newValue.trim())
  }
  render() {
    return <input value={this.props.value} onChange={e => this.onChange(e.target.value)} />
  }
}
```

Notice the following points:
- We *only* needed to change `ComponentInput`
- In `render()`, instead of directing the `onChange` event to whatever is passed via props, we hooked a local function to it
- In that function we preprocess the new value before signaling the change up the component stack

So that's roughly what you need to do in order to create a new component that responds in the appropriate way: the component receives a value via its props, reacts locally to events generating a new value which is propagated up via an event prop.

More concretely for the task at hand: in the `render()` method of your new component, you process `this.props.value` in whatever way you want (split it or whatever) and display its various parts. You then hook up individual `onChange` (or `onKeyDown` or whatever events are needed) to generate a new "candidate" value  that is passed up via `props.onChange`.

app/components/inputs/InputComponent.js
```diff
@@ -0,0 +1,188 @@
...
+  };
+
+  render() {
+    const {
+      showErrors,
+      invalidMessage,
+      requiredMessage,
+      required,
+      invalid,
+      value,
```
matheusd: `value` local variable

app/components/inputs/InputComponent.js
```diff
@@ -0,0 +1,188 @@
...
+      placeholder,
+      onChange,
+      disabled,
+      readOnly,
+      unit,
+      hidden,
+      type,
+    } = this.props;
```
matheusd: coming from `this.props` (ie, the props injected into this component by whoever is using it).

app/components/inputs/InputComponent.js
Jujhar: Great thank you for the guide.

matheusd:
Ok, I'm now doing a functional test and there seem to be a lot of regressions on this input:

- I get a warning in the console about "a console is content editable"
- I can type non-number characters in the cash input (test out how it currently works in master)
- There's a bad "popping" effect on input sizes when moving the mouse over

Jujhar:
* If you look at the specs the main point of this task is to have an input key that has two different sizes and this is not possible without using div content editable. As far as I know there are no drawbacks to using content edible as this is only a frontend component and works smoothly

* If you type a non number and click away it will be reverted to the previous number input.

* The mouse over popping effect was an alternative to clicking to initiate the change which I preferred and have reverted back. It somewhat requires the user to click twice but I think it is intuitive and since this is not something that user does over and over again it isn't really a hold up and doubt is actually noticeable.

matheusd:
Only changing the style on click is an improvement but I'm still getting errors in the console and the input is still allowing invalid (non-numeric) characters to be typed.

Jujhar:
Fixed those two things, I supressed the content editable children error as it needs children to replace each other.

Jujhar:
Rebased

vctt94:
Need fix lint

alexlyp:
Closed due inactivity and debate about usefulness.

PR #1567, opened 2018-08-02
Update to proper api.proto and bump required version

alexlyp:
Missed an update to use `repeated string spv_connect`

### DIFF
--- a/app/actions/WalletLoaderActions.js
+++ b/app/actions/WalletLoaderActions.js
@@ -415,7 +415,10 @@ export const spvSyncAttempt = (privPass) => (dispatch, getState) => {
   const { discoverAccountsComplete, spvConnect } = getState().walletLoader;
   var request = new SpvSyncRequest();
   if (spvConnect) {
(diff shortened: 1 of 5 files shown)


app/actions/WalletLoaderActions.js
```diff
@@ -415,7 +415,10 @@ export const spvSyncAttempt = (privPass) => (dispatch, getState) => {
   const { discoverAccountsComplete, spvConnect } = getState().walletLoader;
   var request = new SpvSyncRequest();
   if (spvConnect) {
-    request.setSpvConnect(spvConnect);
+    var splitSpvConnect = spvConnect.split(",");
```
matheusd: hmmm, could the incoming `spvConnect` here be an array? I believe it's coming from an exported config for the moment, so if feasible, let's use an array as it avoids this kinds of hacky things

alexlyp: that's true.  I'll make that change with the config

PR #1571, opened 2018-08-03
[launcher] Fix start up sequencing of normal mode

alexlyp:
We were previously starting the transaction notifications prior to beginning the initial rescan.  This had a possibility of a race that would cause RESOURCE_EXHAUSTED errors in dcrwallet due to too large of message sizes being sent.

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -2,7 +2,7 @@
 import * as wallet from "wallet";
 import * as sel from "selectors";
 import eq from "lodash/fp/eq";
-import { getNextAddressAttempt, loadActiveDataFiltersAttempt, rescanAttempt,
+import { getNextAddressAttempt, loadActiveDataFiltersAttempt,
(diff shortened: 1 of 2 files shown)


alexlyp:
Still working out an issue with sequencing, WIP

alexlyp:
OK ready for testing

matheusd:
Still testing, but seems to have skipped over the import stakepool scripts step (which happens before rescan) when restoring a wallet

matheusd:
Tested a few combinations of restore/open wallet in regular/spv/advanced/non-advanced mode and didn't hit the resource_exhausted error.

Only problem I see is that it's flashing the import stakepools scripts during wallet restore but automatically starting the rescan (and skipping the import).

alexlyp:
Gotcha addressing now

alexlyp:
Ok fixed issues with stakepool entry screen

PR #1572, opened 2018-08-04
Add options for selecting preferred timezone 

### LINKED ISSUE #1099: Use correct timezone for transaction date/time
Eg: [this testnet tx](https://testnet.dcrdata.org/explorer/tx/44466fda195cf91b3931c83c1ad418a995125c235d201e837318f922bf74e569) happened at  (UTC) 2017-10-16 18:56:27 but is shown on my running decrediton instance as "3:56:27 PM", when the correct time (due daylight savings being effective on this date at my current locale) is 4:56:27 PM. 

So we need a better way of detecting/selecting the user timezone and/or a better way to convert from the int timestamp to a string date/time.

shweini:
This adds a radio button for users to select their preferred timezone ("local" or "UTC") with which they want transaction times to be displayed in. 

In local, Daylight savings time is taken into consideration. 

This fixes #1099

### DIFF
--- a/app/helpers/dateFormat.js
+++ b/app/helpers/dateFormat.js
@@ -1,9 +1,45 @@
 import { format } from "util";
+import { getGlobalCfg } from "config";
+
+function dateToLocal(d) {
+  Date.prototype.stdTimezoneOffset = function () {
+    var jan = new Date(this.getFullYear(), 0, 1);
+    var jul = new Date(this.getFullYear(), 6, 1);
+    return Math.max(jan.getTimezoneOffset(), jul.getTimezoneOffset());
+  };
+
+  Date.prototype.dst = function () {
+    return this.getTimezoneOffset() < this.stdTimezoneOffset();
+  };
+
+  var date = new Date(d);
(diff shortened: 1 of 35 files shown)


app/helpers/dateFormat.js
```diff
@@ -1,9 +1,45 @@
 import { format } from "util";
+import { getGlobalCfg } from "config";
```
matheusd: I'd try to avoid using `getGlobalCfg` here.

The helpers module is intended to be as self contained as possible, so introducing the getGlobalCfg here is not great.

One way to remove this dependency would be the following: 

- Introduce the `dateToLocal`, `dateToUTC` as you did
- Create a `tsToDate` function in the `selectors.js` file. That would get from the currently saved config state and return the correct function (either dateToLocal or dateToUTC)
- Refactor the uses of `tsToDate` (they're currently either in components or on actions) to get from the selector.

@alexlyp thoughts on this? The above will be slighly more refactoring work but would drop the dependency to config here. Do you think it's worth doing?

PR #1573, opened 2018-08-07
Finish watching only view

### LINKED ISSUE #1490: [watch-only]  Various functionalities to hide while isWatchingOnly mode

vctt94:
This PR closes #1490 and it is built on top of #1563.

I added a component to wrap other components and trigger the watching only notifications when clicked. 
I added a POC on the security page view, so you guys can see it. The idea is to wrap each component which will be disabled because of the watch only state. 

What do you guys think?


matheusd:
Seems like a reasonable approach.

vctt94:
This PR is ready for a review

vctt94:
fixed, I also found this class .modal-section, which is not being used anymore and the PurchaseTicketsInfo.less and removed it from the the code.

PR #1575, opened 2018-08-08
[win32] fix win32ipc.node location

matheusd:
`require()` doesn't accept a run time constant. The location of the module needs to be a static string.


alexlyp:
tACK on windows

PR #1576, opened 2018-08-08
Add min and max amount filter for transaction

### LINKED ISSUE #785: Feature Request: Additonal filter type in "History" page.
Would like to request an extension to the current filters, namely "Regular" transaction. This filter should allow user to choose a range of value of the said transactions. For example, I want to be able to see only transactions between 100-200 DCR. 

This is important because as it is, the regular transactions includes a lot of micro transactions involving very small fees and is really not very useful to user end most of the time.

Alternatively, create an extra option to "hide" these micro transactions instead.

vctt94:
closes #785.

This PR is still a wip, because right now I am using the dcrInput to write the filters and I think would be good having a slide ranger for that.

I looked into some libraries for that and found this one: https://github.com/whoisandy/react-rangeslider 

Any suggestions?


matheusd:
The lib seems reasonable.

vctt94:
This lib has no support to react 16. I'll try to find another one.

vctt94:
This PR is ready for a review.

I created our own slide ranger component, because I did not find any that was worth using but it has no style so far.

vctt94:
Thanks for your feedback.

You are right, the idea was to start with the options non visible. I changed it to make some changes and did not change back. Also I added a timeout to the method, now it is making less calls.

app/components/views/TransactionsPage/HistoryTab/index.js
```diff
@@ -100,6 +100,17 @@ class History extends React.Component {
   }
 
   onChangeFilter(value) {
+    const { isChangingFilter } = this.state;
+    if (isChangingFilter) {
```
matheusd: Not quite there yet.

Notice in this file that `isChangingFilter` is not set to true anywhere - meaning that this code path is never exercised. Also, you're scheduling the `changeFilter` in two places.

Roughly what needs to happen is the following:

```javascript
OnChangeFilter(value) {
  const { isChangingFilterTimer } = this.state;
  if (isChangingFilterTimer) {
    clearTimeout(isChangingFilterTimer);
  }
  this.setState({ isChangingFilterTimer: setTimeout(() => this.changeFilter(value), 2000) });
}
````

Notice the following:

- `isChangingFilterTimer` is the handle to the most recently started timeout
- If no changes happen until the timeout (2000ms in this example), then the `changeFilter` is triggered and an update is requested
- If new changes happen, then the previous timeout is canceled and a new one started from zero

The point here is that, within the 2000ms interval many changes to the values can take place but only a single `changeFilter` must be triggered at the end of this.

You'll also need to clear the `isChangingFilterTimer` in the `changeFilter` (omitted here for brevity).

**Comment by alexlyp** [comment]
Looks like there are still some issues @matheusd pointed out above.

Also please take a look at #1706 for design directions

**Comment by alexlyp** [comment]
it's working really well!

Couple things I noticed that we could probably open additional issues:
1) Sort by tooltip on hover of sort selections.
2) Since the bar is pretty small we should have "clicks" in the bar be set at 0.5 or 1 DCR, so if you slide the bar 1 px (or whatever) to the left or right it changes by that minimum amount.  
3) We'll need to do some  final styling of the bar.

Great job on this.  I'll try and review the full code today or tomorrow.

**Comment by alexlyp** [comment]
Just noticed that you did something wacky on the yarn.lock.

I'd make sure you have yarn 1.10 installed, revert the lock back to master, run yarn again then commit that updated lock.

File: app/components/buttons/EyeFilterMenuWithSlider/index.js [hunk 2]
```diff
@@ -0,0 +1,152 @@
...
+      minAmount: 0,
+      maxAmount: 100,
+      min: 0,
+      max: 100,
+      sliderShower: true,
+      expandedSliderInfo: false,
+    };
+  }
+
+  getSliderWhenOpeningMenu() {
```
matheusd: Rename to `createSliderRangeInElement(range)`

app/components/buttons/EyeFilterMenuWithSlider/index.js
```diff
@@ -0,0 +1,152 @@
...
+    setTimeout(() => {
+      const range = document.getElementById("min-max-slider");
```
matheusd: Don't use `getElementById` as it breaks componentization (you wouldn't be able to add 2 instances of `EyeFilterMenuWithSlider` at the same time due to them having the same id).

Parametrize `getSliderWhenOpeningMenu` (renaming it to `createSliderRangeInElement`) with what you used as `range` (that's the `range` parameter), then add a `ref` prop to `#min-max-slider` (I'll not where below).

app/components/buttons/EyeFilterMenuWithSlider/index.js
```diff
@@ -0,0 +1,152 @@
...
+  getSliderWhenOpenedMenu() {
+    const { currencyDisplay } = this.props;
+
+    const { sliderShower, expandedSliderInfo,
+      min, max, maxAmount, minAmount } = this.state;
+
+    return (
+      <div className="history-slider-wrapper">
+        <div className="history-amount-range-label"><T id="history.amount.range" m="Amount Range" /></div>
+        <div id="min-max-slider" className="min-max-slider"></div>
```
matheusd: Here, you'd use:

```suggestion
        <div id="min-max-slider" className="min-max-slider" ref={r => this.createSliderRangeInElement(r)}></div>
```

So when the ref is established, the `noUiSlider` lib will be invoked to create the slider in this particular ref.

vctt94:
Thanks for the review guys. 

I added the @matheusd suggestions and added the step on the slider. 

The tooltip I think it would be good having an issue for that, cause I believe it will be necessary to tweak a bit on the `EyeFilterMenu`, cause right now we are only wrapping it with the tooltip, and I believe it will be necessary to pass an insider tooltip or something similar for that tooltip to get out.

app/components/buttons/EyeFilterMenuWithSlider/index.js
vctt94: good catch

matheusd:
Code looks ok now. Only hit another small usability issue: If you click exactly on top of one of the edges (the round thingies) the transaction list starts to update even though the actual filter didn't change.

And if you click either inside or outside the track *without* dragging the mouse, the values update but a new listing isn't triggered

vctt94:
Thanks for the catch @matheusd ,  I fixed it.

I also found a bug on the min max slider value and fixed as well.

matheusd:
Changes are working, but I found another small usability issue.

If you change the filter, go to a different page, then come back, the filter resets to the default 0-100 but the transaction list itself remains unchanged.

I think you should read the range filter from the tx filter state when entering the component and setup the initial view of slider based on that, so that users can choose a filter, look at a particular transaction, then go back and further change/refine the filter without losing the previous filter.

vctt94:
You are right @matheusd , fixed that as well :)

vctt94:
Actually, I think it is better to remove this part of code: 

```
if(value === minAmount || value === maxAmount) {
  return;
}
```
Because the `minAmount` and `maxAmount` value are the ones when the slider is created, and if the value is changed, this check stops working, calling the method again to update transactions, and if called to the values where they were before, it does not call the method. 

I tried changing that to `this.state.minAmount` and `this.state.maxAmount`, but as the  setState is called on the update method, it is the same when the `set` is called.

PR #1584, opened 2018-08-12
Create wallet page layout fix

### LINKED ISSUE #1578: [create-wallet] create wallet page messes up when width is too large
When creating a wallet, on the step to add the seed's words, if the width is too large the page get messed up

pagebug

Jujhar:
The create wallet confirm seed page should now be proper and always usable no matter how large the width is. 

Fixes #1578


alexlyp:
already resovled

Issue #1582, opened 2018-08-12
Visual Design: v1.3.0 RC1 grouped minor bug fixes


linnutee:
Overall RC1 is looking good, mapped out some smaller design fixes that have a worthwhile effect of visual hygiene for the release: 
Preview: https://xd.adobe.com/view/b3e28600-8bfb-4c5a-441d-aeafa10d5859-c5d3/ (scroll down)
Specs: https://xd.adobe.com/spec/5a5726c5-e6ea-4273-6262-82582819feec-2c66/

1 Launcher cards should be left-aligned alongside rest of the elements. 

2 Different icon states are not displaying correctly (e.g. for example, a Purchased ticket displays an odd green ticket icon, whilst it should display Unmined, so on.) @alexlyp can consider making this a separate issue, if it requires further design documentation. 

3 Fine details around alignment and some styling of the ticket and tx cards is a little bit off. Measured from screenshot:
- The white background rows should be +1px taller in height
- Card titles (voted, amount) should be either +2 or +3px positioned up in X (+3 if rows are made higher). 
- Card subtitles, should also be positioned either +2 or +3px upwards in X. Basically these need to look centred with enough spacing towards the bottom edge.
- Not certain if we removed addresses from recent tx altogether, not showing up at least. (If done so, not a bug and ignore this).
- Staking card subtitles should be with color `#8997A6` (atm in RC1: `#C4CBD2`, that was the color for Dx.…… addressed subtext, being one step lighter than staking subtext). 
- Staking card titles text: `Source Sans Pro; Regular; 15px`

4 Some misalignment of the fields and search styling in Export View. 

5 History empty states icon and text overlap. 

6 Won't go much further with My Tickets view, as this should get further design work in the future imo. However this oversized kebab button has been an eye sore for a while. Should be 10px in height. 

7 Expanded accounts background should be `#F9FAFA`

8 Ticket fee icon is not displaying correctly under tx details view. Instead displays the previously used wallet icon.

PR #1592, opened 2018-08-15
Low contrast label fix

### LINKED ISSUE #1591: Too Low Contrast in History of Transactions View When No Transactions
I can barely see the text "No transactions found".
See 
poor contrast

Jujhar:
Darkened the grey to a close black for the _No Transactions Found_ label.

image

Fixes #1591

### DIFF
--- a/app/components/indicators/NoTransactions.js
+++ b/app/components/indicators/NoTransactions.js
@@ -3,7 +3,6 @@ import { FormattedMessage as T } from "react-intl";
 export default () => (
   <div className="no-transactions-indicator">
     <T id="noTransactions.description" m="No Transactions Found" />
-    <br />
     <div className="no-transactions-indicator-img">
     </div>
   </div>
--- a/app/components/indicators/NoTransactions.js
+++ b/app/components/indicators/NoTransactions.js
@@ -3,5 +3,8 @@ import { FormattedMessage as T } from "react-intl";
 export default () => (
   <div className="no-transactions-indicator">
     <T id="noTransactions.description" m="No Transactions Found" />
+    <br />
+    <div className="no-transactions-indicator-img">
+    </div>
   </div>
 );
--- a/app/components/views/TransactionsPage/HistoryTab/Page.js
+++ b/app/components/views/TransactionsPage/HistoryTab/Page.js
@@ -36,15 +36,13 @@ const Page = ({
     useWindow={false}
     threshold={loadMoreThreshold}
   >
-    <div className="history-content-title" style={transactions.length == 0
-      ? { display: "none" }
-      : { display: "initial" }} >
+    <div className="history-content-title"
+      style = {{ display: transactions.length == 0 ? "none" : "initial" }} >
       <div className="history-content-title-text">
         <T id="history.title" m="Transaction History" />
       </div>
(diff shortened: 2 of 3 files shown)


alexlyp:
Please use 091440 instead of the one you chose.

Jujhar:
Updated, should be ready to go now.

Previously it did not overlap if window was at a certain width now it will not overlap at all.

matheusd:
Please look at the latest specs as pointed out in #1582 

https://xd.adobe.com/spec/5a5726c5-e6ea-4273-6262-82582819feec-2c66/screen/f18f4868-bc65-4642-ae12-a7611037d606/Tickets-My-Tickets-Empty-state-3/

Jujhar:
Updated. I changed the colour as well as according to the specs.

app/components/indicators/NoTransactions.js
```diff
@@ -3,5 +3,8 @@ import { FormattedMessage as T } from "react-intl";
 export default () => (
   <div className="no-transactions-indicator">
     <T id="noTransactions.description" m="No Transactions Found" />
+    <br />
```
matheusd: Please don't use `<br>` for layout, use a proper css style such as `margin`.

matheusd:
Also notice that in the specs require that the title and filters not be shown when there are no transactions.

app/components/views/TransactionsPage/HistoryTab/Page.js
```diff
@@ -40,7 +40,9 @@ const Page = ({
       <div className="history-content-title-text">
         <T id="history.title" m="Transaction History" />
       </div>
-      <div className="history-select-tx-types-area">
+      <div className="history-select-tx-types-area" style={transactions.length == 0
+        ? { display: "none" }
+        : { display: "initial" } } >
```
alexlyp: We don't use inline style

Jujhar:
Updated those please have another look.

alexlyp:
You're still using an inline style.

app/components/views/TransactionsPage/HistoryTab/Page.js
alexlyp: Well I suppose we do use inline styles in certain applications.  But I'd rather this be 

`style{{display: transactions.length == 0 ? "none" : "initial"}}`

Jujhar: OK great think its ready now.

Jujhar:
I ran lint-fix and it passed

app/style/MiscComponents.less
```diff
@@ -647,7 +652,15 @@
 }
 
 .no-transactions-indicator {
+  color: #617386;
```
alexlyp: Where did this color come from?  You need to use the design specs to choose the proper color.

Jujhar: That is from the specs, it is the color of the text.
https://xd.adobe.com/spec/5a5726c5-e6ea-4273-6262-82582819feec-2c66/screen/f18f4868-bc65-4642-ae12-a7611037d606/Tickets-My-Tickets-Empty-state-3/

alexlyp: Gotcha, just didn't recognize it.

PR #1593, opened 2018-08-15
[stats] account for unmined txs in backwards stats

matheusd:
Unmined transactions need to be accounted for when generating stats in backwards mode, otherwise the daily balances charts get broken when purchasing tickets.

This bug only happens after the split transaction is mined but before the ticket is mined, so we need to account for unmined tickets to produce data for the chart.


matheusd:
To reproduce the bug you need some special conditions:

- A wallet with a voted ticket on a previous day (so you had a locked balance on one of the last days of the charts, but today that balance is spendable)
- Just enough balance to purchase a ticket with very little change

You then purchase a ticket and wait for the split transaction to be mined. In the interval between the split being mined and the ticket itself, the chart will be bugged (after the ticket is mined the chart is fixed).

Issue #1614, opened 2018-08-23
[Test] Add functional tests


vctt94:
Right now we started chatting about add tests on decrediton.

@alexlyp suggested to start adding functional tests on critical parts first, and we believe this is a good start.

We can start by adding e2e tests on main's decrediton functionalities:

- [ ] Send Transactions
- [ ] Receive Transactions
- [ ] Purchase Tickets 
- [ ] Balance Check
- [ ] Seed Restore
- [ ] Revoke Tickets
- [ ] Add New Account
- [ ] Sign Message
- [ ] Verify Message
- [ ] Validate Address

After that we can see how things are going and decide what to test next.

We are still discussing about possible libs and framework.

vctt94:
Closing this for now as we agreed it will be better making tests at dcrwallet's end

PR #1616, opened 2018-08-27
Add first integration test

vctt94:
This PR starts #1614 

This is still a wip so you guys can see the direction I am going with the integration tests.

So far I am using Jest, webdriverIO, Selenium and chromedriver to be able to test.

To test it is necessary to: 

- copy decrediton's bin folder to `node_modules/electron/dist/resources/` 
- start the chromedriver server `./node_modules/.bin/chromedriver --url-base=wd/hub --port=9515`.
- run `yarn test`

Right now Decrediton can sync when testing in non advanced mode. The goal after that is to test the main functionalities which is described at #1614.

As the test's wait time are hard coded, if the blockchain needs to sync, it will cause the tests to fail as decrediton will not sync in the given time. 

I believe the proper way to make this tests would be with the simnet network, as @matheusd suggested. But I am not sure the effort it will be necessary to add simnet and right now it is possible to test with the testnet network. We can discuss that further.

After the tests are finished, decrediton is not closing, because I did not discover a proper way to close the daemons from the tests.

References: 

- https://electronjs.org/docs/tutorial/using-selenium-and-webdriver
- http://webdriver.io
- https://jestjs.io/


test/e2e/app.test.js
```diff
@@ -0,0 +1,57 @@
...
+describe('Application launch', async () => {
+  beforeAll(async () => {
+    await launch();
+  })
+
+  // afterAll(() => app.end())
+
+  it('Sync Decrediton', async () => {
+    await sleep(5000);
+    await app.click('//*[@id="root"]/div[1]/div/div/div/div/div[5]/div/div/div[1]/div[5]/span')
```
matheusd: This is one of the main challenges I had when working with selenium, though it was somewhat less prominent pre-react because we used classes and IDs more extensively: how do you locate an element to perform an action or check a state.

Don't know what the best practices are with react (post some links if you find anything about it). But this style will _definitely_ be very brittle: a single move of a component in the hierarchy is going to break this (and possibly many other) tests.

Maybe we should start adding a `testID` attribute to components which will be used in a test, such that this id is generated when building a development version of decrediton but not on production. But this will increase the verbosity of the code.

vctt94:
It was supposed to pass, does it fail with a message?

test/e2e/app.test.js
vctt94: Right now I am using xpath, but yeah, you are right, this can be a problem when testing.

Here http://codefuse.io/ui-test-object-identification-with-id-css-xpath/ there is a text explain the pro's and con's in using ids, css selectors and xpath.

I believe the proper order would be: Ids, then css selectors, and then in last case xpaths, when can not use both, because this is the more suggestive to changes. But beyond that I believe there is nothing we can do beyond fixing the test when doing changes on decrediton. 

I do not think would be good changing all the code to add testID all over the code, just for that. Maybe adding ids?

matheusd:
This was the error in the terminal:

https://0bin.net/paste/ZvLJiq-tC0zxTunu#dpHxiz8PqJDV4CMe76AlGS9NOzK2AoYKeXZG6LNhKfu

vctt94:
Your decrediton is not opening or does it open and closes? If it open and closes I believe the problem is with the symlink on the folder not being reachable which makes decrediton closes. If that is the case, can you try copying and pasting it?

matheusd:
Copying and using spv mode went further along, but now I'm getting 
``An element could not be located on the page using the given search parameters (".overview-header")```

vctt94:
that's because decrediton did not finish syncing on the given time. Did it spend time syncing? Maybe I need to increase the time, 10 seconds may not be enough.

vctt94:

I added a send transaction test, but in order to the tests pass, it will be necessary a testnet wallet with passphrase 12345678 and it needs to be the last one opened, beyond the steps described on the description. 

Also Travis is failing, but I am not sure on how to integrate these tests with Travis, it will be necessary some more research on that.

test/e2e/app.test.js
```diff
@@ -0,0 +1,86 @@
+const assert = require('assert')
```
brunobraga95: I did not really understand why this folder is called e2e, we are adding integration tests right? if so, this is misleading.

vctt94: Thanks for reviewing this PR.

You are probably right, but we discussed on slack about adding gRPC tests, and we got to the conclusion it would be better to add them at dcrwallet. Therefore, this tests were moved to https://github.com/decred/dcrwallet/pull/1298.

So this PR is abandoned, and I will close it.

PR #1617, opened 2018-08-27
[launcher] Implement steps for new SpvSync stream API

alexlyp:
~This requires usage of https://github.com/decred/dcrwallet/pull/1244~

This is the first step in properly refactoring and cleaning up the launch procedure to instead use the SyncResponse API.  This will allow is us remove a lot of the cruft surrounding the various step for starting and syncing dcrwallet behind the scenes.   

I have also simplified the loader progress for SPV fetching headers. Now it simply shows the time back instead of trying to give an estimate of how much longer it should run.  This seems reasonable now since fetching headers with a reason set up should be a matter of minutes so giving an estimate wouldn't be necessary.


alexlyp:
Hrm I believe I've encountered this periodically on master.  Can you try to replicate it on master as well?

app/actions/WalletLoaderActions.js
```diff
@@ -430,15 +446,86 @@ export const spvSyncAttempt = (privPass) => (dispatch, getState) => {
...
+        break;
+      }
+      case SyncNotificationType.FETCHED_HEADERS_STARTED: {
+        const fetchTimeStart = new Date();
+        dispatch({ fetchTimeStart, type: SYNC_FETCHED_HEADERS_STARTED });
+        break;
+      }
+      case SyncNotificationType.FETCHED_HEADERS_PROGRESS: {
+        const lastFetchedHeaderTime = new Date(response.getFetchHeaders().getLastHeaderTime()*1000);
+        console.log(lastFetchedHeaderTime);
```
matheusd: remove

vctt94:
Actually on master I am getting stuck when  restoring a wallet, but if I refresh decrediton, the bug does happen.

alexlyp:
OK dcrwallet master has been updated with the SpvSync stream, this PR should be ready for final review.

vctt94:
Here is an image of what I mentioned: 
loopbug

vctt94:
I think maybe decrediton is opening many grpc connections, because when I asked to close decrediton I received many error messages. Here is the log https://pastebin.com/qtXiuXAR

alexlyp:
Hrm strange, I think I've noticed this on master as well.  Can you confirm that it's a master issue?

PR #1625, opened 2018-08-28
Fix Watch only logo

vctt94:
This fixes the logo being undefined when opening a watching only wallet. I will also open an issue to refactor these selectors so use only one in both case using an `or` method.


alexlyp:
Seems OK, though I'm thinking this is kind of showing an issue with our ongoing naming system.  Worth thinking about refactor selectors to make naming and usage more clear.

PR #1631, opened 2018-09-04
[multi] fix tx details and home icons

### LINKED ISSUE #1622: [Overview] Duplicated information at stake transactions
On the overview stake transactions, there are some duplicated information about tickets, because they were supposed to show as purchased and voted, but when a ticket is voted, both transactions show as voted.

Image Showing the same information twice: 
decreditonbuggy

matheusd:
Fix #1622 

This fixes the icons related to ticket transactions in the tx details page and shows the correct status for recent stake txs in the overview.

The fix is slightly hackish, but I'll improve it later as part of a refactoring of the GetTickets call.

### DIFF
--- a/app/components/views/TxDetails.js
+++ b/app/components/views/TxDetails.js
@@ -17,13 +17,13 @@ const messages = defineMessages({
 });
 
 const headerIcons = {
-  in:         "plusBig",
-  out:        "minusBig",
-  Coinbase:   "plusBig",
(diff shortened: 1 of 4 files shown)


matheusd:
Screenshots of the results

Unmined, Immature and voted:
screenshot_2018-09-04_10-39-05

Simple wallet with only a single ticket being continuously re-purchased
screenshot_2018-09-04_10-40-59

PR #1636, opened 2018-09-05
[actions] Improve close wallet action

### LINKED ISSUE #1551: Add close wallet action
Add an action (maybe in the main sidebar?) to close the current wallet, return to the wallet selection screen and allow the user to open or create a different wallet.

This might be slightly tricky as we need to close the dcrwallet process and all connections, and all the UI code assumes that once the wallet is loaded it won't change.

Might wanna hold off implementing this only after spv hits, since this will further change things.

matheusd:
Fix #1551 

This improves the close wallet action by cleaning up wallet state from redux, cancelling all outstanding streams and stopping the wallet, then returning to the wallet selection page.

Tested on spv and regular sync modes, but only in linux.


vctt94:
Tested and seems to work fine. Good job

alexlyp:
Seems to work pretty well.  Noticing the Open Wallet page coming up on subsequent loads briefly.  Otherwise gtg

matheusd:
@alexlyp Which page exactly?

alexlyp:
After I close the wallet and then select a new wallet to load, the Open Wallet page (asking for the public encryption passphrase) is shown briefly while the button for that is appears to be "loading/working".

matheusd:
Weird, I never hit that. Did the previously open wallet require the public encryption passphrase?

alexlyp:
No it was a default password.

Definitely not a huge issue and shouldn't block this PR, I can open an issue to address it since it's sort of out of scope of these changes.

alexlyp:
Looks good.  I'll keep testing and take care of the open wallet flash if I keep seeing it.

PR #1644, opened 2018-09-12
[launcher] Add new RpcSync stream for loading

alexlyp:
This requires usage of https://github.com/decred/dcrwallet/pull/1284.

This allows the entire startup procedure to be cleaned and refined.  Instead of using a series of linked calls, we are able to now just call the rpcSync once and the rest is handled by dcrwallet.  

Now since we are using the same notifications as the previously added SpvSync, we are able to reuse much of that existing infrastructure.  

NOTE: There is still an outstanding issue with the importing of stakepool keys during startup.  Unclear if the fix for that will end up being included here or if we will update that at a future date.


matheusd:
Ok, next set of tests. I was able to create a new wallet and load existing wallets, but seems that wallet restore with an existing seed gets stuck in the "Syncing Wallet".

alexlyp:
Ok dcrwallet has been updated to include the rpcsync method

alexlyp:
I will be trying to replicate the issues with restore from seed

alexlyp:
Ok looks like it's good to go for final testing

app/index.js
```diff
@@ -203,7 +203,7 @@ var initialState = {
   },
   walletLoader: {
     spvConnect: globalCfg.get("spv_connect"),
-    spvInput: false,
+    syncInput: false,
```
matheusd: Need to add the remaining new state vars. I believe they're:

- syncError
- syncAttemptRequest 
- syncCall
- peerCount

app/components/views/GetStartedPage/SpvSync/index.js
```diff
@@ -72,6 +72,9 @@ class SpvSync extends React.Component {
   }
 
   onSetPassPhrase(passPhrase) {
+    if (passPhrase != "")
```
matheusd: Surround block with `{}`

app/components/views/GetStartedPage/DaemonLoading/index.js
```diff
@@ -45,11 +67,50 @@ class DaemonLoading extends React.Component {
...
+
+    const { onRetryStartRPC, onSetWalletPrivatePassphrase } = this.props;
+
+    onSetWalletPrivatePassphrase && onSetWalletPrivatePassphrase(passPhrase);
+    onRetryStartRPC(false, passPhrase);
+    this.resetState();
+  }
+
+  onKeyDown(e) {
+    if(e.keyCode == 13) {   // Enter key
```
matheusd: Add space after `if`

PR #1663, opened 2018-09-25
Build with node 10.11.0

dajohi:
- Fix linter warnings
- Upgrade dependencies


alexlyp:
Everything running clean for me.  Linux dev, Linux pkg and Windows pkg.

PR #1674, opened 2018-09-28
[tickets] Fix SPV ticket purchasing

alexlyp:
Currently ticket purchasing for SPV is only not allowed on the second to last block of a window.  This is due to the upcoming changing price will ultimately cause the ticket itself to be pruned from the mempool.  Now we are just blocking ticket purchasing on that 1 block.

### DIFF
--- a/app/connectors/spv.js
+++ b/app/connectors/spv.js
@@ -5,7 +5,6 @@ import * as sel from "../selectors";
 const mapStateToProps = selectorMap({
   spvMode: sel.spvMode,
   blocksNumberToNextTicket: sel.blocksNumberToNextTicket,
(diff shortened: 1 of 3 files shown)


app/connectors/spv.js
```diff
@@ -4,6 +4,7 @@ import * as sel from "../selectors";
 
 const mapStateToProps = selectorMap({
   spvMode: sel.spvMode,
+  blocksNumberToNextTicket: sel.blocksNumberToNextTicket,
```
matheusd: You can drop the next item (`blocksPassedOnTicketInterval`) since it's not used in the component anymore

PR #1679, opened 2018-10-01
[win32] Explicitly close rx pipe when closing wallet

### LINKED ISSUE #1645: [main] Close wallet on windows
Currently the newly added close wallet functionality does not properly shutdown the wallet on windows.  This is due to the current way windows is shutdown with pipes on windows.  It is shutdown when decrediton closes, which closes the pipe which causes the process to close.  We'll need to re-rig this to allow for the pipes to close on request instead of only on decrediton shutdown

### LINKED ISSUE #1653: 1.3 RC2 issue: Error creating wallet
When you want to cancel creating a new wallet on 1.3 RC2 (while you are already shown the seed), clicking the `cancel` button or the top right `X` gives you an "Error creating wallet" message.

After closing the wallet and restarting it, the name of the failed wallet shows up in the welcome screen with the message "Setup Incomplete" as shown below: 
image

When you click "Launch Wallet" here, Decrediton shuts down without warning.

This issue was found and replicated on a Windows 64-bit device.

### LINKED ISSUE #1675: v1.3.0 - Cannot remove partially setup wallets
Tested on Windows 10 Insider Preview

Steps to repro:
* Create a New Wallet OR Restore Existing Wallet
* Cancel
* Try and Remove

See attached videos. I **was** able to remove after exiting Decrediton and re-launching.

https://1drv.ms/v/s!AktSiyBbVlbrgZiCQ_pFIz4Y3cHl3N0
https://1drv.ms/v/s!AktSiyBbVlbrgZiCRJ0WPyRosRqgGU8

matheusd:
Close #1675 
Close #1653 
Fix #1645

### DIFF
--- a/app/main_dev/launch.js
+++ b/app/main_dev/launch.js
@@ -14,7 +14,10 @@ const logger = createLogger(debug);
 
 let dcrdPID;
 let dcrwPID;
+
+// windows-only stuff
 let dcrwPipeRx;
+let dcrdPipeRx;
 
 let dcrwPort;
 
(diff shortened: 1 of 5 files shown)


matheusd:
@jrick Updated the module so that `ClosePipe` works with the same pipe object produced by `CreatePipe` as you suggested.

See if you prefer this version.

PR #1685, opened 2018-10-03
[i18n] Prevent duplicate IDs in translation strings

### LINKED ISSUE #1677: Remove duplicated translation strings
There are a few duplicated translation strings throughout the code.

Most of these have the same value, but occasionally a string will slip through with the same id but different value, which breaks the translation system (eg: the logs title fixed in #1676).

We need to consolidate all duplicated strings to prevent any future problems, then ideally add a CI rule to detect and prevent them from being merged.

matheusd:
Close #1677 

~WIP for the moment to test CI integration in our workflow.~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ install:
   - yarn
 
 script:
-  - yarn lint
-  - yarn test
+  - yarn lint || travis_terminate 1
+  - yarn test || travis_terminate 1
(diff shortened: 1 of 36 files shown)


matheusd:
Ready for review. Might be slightly easier to review the changes by looking at each commit individually (first one is the new script, the others are fixing the dupes).

PR #1690, opened 2018-10-05
[settings] Clean teardown upon certain setting changes

alexlyp:
~This still requires pipe work to be merged for it to properly work on Windows.~

Currently there are 2 settings changes that will cause a reset to occur: SPV mode and Advanced Daemon.  We will be adding Network (mainnet/testnet) toggle once this PR has settled and has been shown to work.

Currently I am testing these possible paths:
- [x] Normal  to Advanced
- [x] Advanced to Normal
- [x] Normal to SPV
- [x] SPV to Normal
- [x] Advanced to SPV
- [x] SPV to Advanced

I'll also be exposing the SPV Connect configuration option, but the above paths will be tested with this option turned off.  I will be probably adding SPV Connect setting to the options that cause reset.  But like network change we'll leave it for another PR.

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -219,6 +219,22 @@ export const createWallet = (createNewWallet, selectedWallet) => (dispatch, getS
     });
 };
 
+export const CLOSEDAEMON_ATTEMPT = "CLOSEDAEMON_ATTEMPT";
+export const CLOSEDAEMON_FAILED = "CLOSEDAEMON_FAILED";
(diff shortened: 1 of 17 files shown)


matheusd:
I've updated #1679 to also close the dcrd process via the new win32ipc module `closePipe`. Given that the pipetx refactor can't be put in until after electron get's updated with node 10.10+, we can merge only the piperx end and leave pipetx for the future.

Let me know in which order you want to merge stuff (either this one or #1679 first) so I can property test in windows...

alexlyp:
OK updated with new closePipe, ready for final testing

PR #1705, opened 2018-10-10
[launcher] Delay getBestBlock slightly to avoid issues on reload

### LINKED ISSUE #1701: [launcher] GetBestBlock failing sometimes
I'm seeing periodic GetBestBlock request failures from time to time.  Usually on reloads, but I believe I've seen it on normal start up as well.

alexlyp:
Close #1701 

I can regularly hit the getBestBlock failure error without this change.  I was able to complete 10 reloads in a row with no issues.  This was the case in both SPV and normal mode.


matheusd:
New version is definitely better!

PR #1727, opened 2018-10-17
[stats] Improve startup stats in large wallets

matheusd:
This improves the startup time and usability for large wallets by:

- Using local (js) code to decode transactions instead of the wallet service
- Only starting to listen for tx/account notifications after full startup
- Preventing repeated stats calculation if the initial calc time was big

Testing on a large testnet wallet (96k+ votes, 3k+ live tickets) startup time reduced from ~10min to ~2min.

Currently a WIP as there's some instrumentation left to facilitate testing.

The initial code for transaction decoding was done by @alexlyp


alexlyp:
Tested great on a large testnet premine wallet that has been consistently voting.  Definitely can remove the instrumentation now so we can do final reviews.

matheusd:
Rebased and cleaned up. Ready for final review and tests.

PR #1728, opened 2018-10-18
More allusive passphrase update failed message

### LINKED ISSUE #1694: Error shown if passphrase reset is attempted with wrong passphrase
**Steps to recreate:**
- Open the settings tab
- Click "Update Private Passphrase"
- Enter an incorrect value for current passphrase, and two matching values for the new passphrase.
- Continue

**Expected result**
- A meaningful error informing the user their passphrase was incorrect

**Actual result:**
- An error saying: `TypeError: Cannot read property 'PRIVATE' of undefined`

Jujhar:
Fixes #1694

Now it says "Entered incorrect private passphrase reset blocked."

### DIFF
--- a/app/reducers/snackbar.js
+++ b/app/reducers/snackbar.js
@@ -156,7 +156,7 @@ const messages = defineMessages({
   },
   CHANGEPASSPHRASE_FAILED: {
     id: "settings.errors.changePassphraseFailed",
-    defaultMessage: "{originalError}"
+    defaultMessage: "Entered incorrect private passphrase reset blocked."
   },
   DECODERAWTXS_FAILED: {
     id: "decodeRawTx.errors.decodeFailed",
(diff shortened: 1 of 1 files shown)


app/reducers/snackbar.js
```diff
@@ -156,7 +156,7 @@ const messages = defineMessages({
   },
   CHANGEPASSPHRASE_FAILED: {
     id: "settings.errors.changePassphraseFailed",
-    defaultMessage: "{originalError}"
+    defaultMessage: "Entered incorrect private passphrase reset blocked."
```
alexlyp: This message doesn't make any sense

Jujhar: When you reset the password you are asked for three options first being your old private password and if that is wrong this message comes up.

Maybe its better if it is "Password reset failed entered incorrect previous passphrase"?

alexlyp: I think you're being too verbose.  

"Update password failed.  Incorrect private passphrase, please try again."

PR #1744, opened 2018-10-25
[ticketbuyer] Implement TicketBuyerV2

alexlyp:
~This requires https://github.com/decred/dcrwallet/pull/1307 (rpc v5.4.0)~

This PR implements the new ticketbuyer v2 that was added by @jrick in his dcrwallet SPV PR.

Because of the stake difficulty being extremely stable these days, there is now no need for all of the extra options that were included.  This makes the ticketbuyer much more streamlined and easy to use.  Now the user can just supply their preferred balance to maintain, account and stakepool to use.  

This also will now allow the user to running multiple ticket buyers (one for each account).


alexlyp:
OK this is ready for testing and final reviews.

Issue #1736, opened 2018-10-19
[feature request] Close passphrase modal when user hits escape


jholdstock:
Kind of self explanatory. It would be nice to quit out of the passphrase modal without having to use the mouse - for me the most intuitive way would be hitting the escape key.

PR #1745, opened 2018-10-25
Close passphrase modal on ESC keypress

szpasztor:
Resolves https://github.com/decred/decrediton/issues/1736

### DIFF
--- a/app/components/modals/PassphraseModal/PassphraseModal.js
+++ b/app/components/modals/PassphraseModal/PassphraseModal.js
@@ -11,6 +11,10 @@ class PassphraseModal extends React.Component {
     this.state = this.getInitialState();
   }
 
+  componentDidMount() {
+    document.addEventListener("keydown", this.onKeyPress, false);
+  }
+
   onCancelModal() {
     this.resetState();
     this.props.onCancelModal && this.props.onCancelModal();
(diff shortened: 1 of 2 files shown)


app/components/modals/PassphraseModal/PassphraseModal.js
```diff
@@ -11,6 +11,10 @@ class PassphraseModal extends React.Component {
     this.state = this.getInitialState();
   }
 
+  componentDidMount() {
+    document.addEventListener("keydown", this.onKeyPress, false);
```
matheusd: Don't use globals if at all possible. This will cause this event handler to be called whenever any key anywhere in the app is pressed after the first modal is shown, will cause a stack of handlers being added (one for every time a modal is shown), is likely to cause problems given it's never removed (and will be called even after this component has been unmounted) etc.

Could this handler be added only on the top-level component of the modal?

szpasztor: A quick search shows that `PassphraseModal` is referenced in many different places - managing the handler in multiple contexts seems error prone. What if the handler stayed on the modal but was removed when the modal is closed (in `onCancelModal`)? That would address the stack problem and be more DRY. Let me know what you think or if I misunderstood anything.

matheusd: By "top level component" I meant the top level component _inside_ the PassphraseModal (the `<Modal ...` instance).

So `Modal` would have a new optional prop (`onKeyDown`) propagated down its own stack, or maybe the cancelation via ESC could be implemented directly in `Modal`, so that *all* modals could be closed with that.

szpasztor:
Moved the handler to `Modal`, also verified in the dev tools that the `eventListener` is not present after the modal has been closed.

PR #1747, opened 2018-10-29
Update pkgs

dajohi:


alexlyp:
Hrm looks like the stylelint bump is angry about some stuff.  Either revert that bump or find a way to quiet the linting errors.

PR #1749, opened 2018-10-29
[design] Clean up and add rigging for color themes

alexlyp:
This is the first step to introduce alternative color themes for decrediton.

There are no more direct references to hex color codes in the CSS, though there may be some remaining rbga() refs that will need to be addressed as well.  I will continue to tweak various elements as I find oddities with colors.

For now, I've stubbed in a very rough pass at a "dark theme" that can be used until the @linnutee and the others at Eeter deliver an official dark theme.


app/style/main.less
```diff
@@ -94,9 +70,76 @@
...
+  --error-text-color: #fd714b;
+  --error-red: red;
+  --vote-yes-color: #2ed8a3;
+  --vote-no-color: #f2734f;
+  --vote-abstain-color: #596d81;
+  --white-border: #d1d5db;
+}
+
+.theme-light {
+  --background-back-color: #fff; // all gone
```
matheusd: What do these "all gone" mean here?

alexlyp: Ah apologies on these, some left over comments that I was using to be organized

PR #1754, opened 2018-10-30
[launcher] Use getblockchaininfo to properly display progress

### LINKED ISSUE #1295: User was able to get into main UI prior to sync
Unknown what may have caused this, but I'm guessing that they had a bad response for current block count so it didn't force a check and let them through.

This led to some odd issues when in the full app with rescan etc.  

Not sure on the potential solutions, but we should brainstorm

### LINKED ISSUE #1753: [launcher] Use newly implemented getblockchaininfo for progress
Now we can use the new dcrd rpc command for "getblockchaininfo" to get the "verificationprogress" to report back how far along the user is for syncing the blockchain.

This will remove the need to check blockchain height from dcrdata and will be able to consistently display the progress for normal (non-SPV) dcrd syncing.

alexlyp:
Close #1753 
Close #1295 

I haven't yet cleaned up the privacy settings to allow contact to dcrdata since I believe we'll soon be adding a new call for the dev subsidy amount.


davecgh:
So, I noticed this is manually calculating the verification progress based on the current height and the sync height.  That is fine for now, because it's the same thing dcrd is doing, but I would suggest using the `verificationprogress` field, because the plan in the future is to significantly improve the reporting by using the overall number of transactions instead of blocks, since blocks alone is not really a great measure when the block sizes grow.

Issue #1743, opened 2018-10-25
Settings optimizations


MariaPleshkova:
Preview: https://xd.adobe.com/view/d8af575b-6f2d-478b-4a9a-6b889c6ce444-ae9e/
Specs: https://xd.adobe.com/spec/47abb9cc-efef-42e3-45ae-9c70b38b8f77-319a/

Add the button "Close Wallet" in the header and add the pop up with "Are you sure you want to close Decrediton?"

jholdstock:
Hi @MariaPleshkova, just to clarify, the "Close Wallet" button does not actually close Decrediton. It only closes the current wallet, resulting in the user landing back on the wallet selection menu.

linnutee:
Ah right, thought also this was closing the software. I think the modal safety roadbump could remain here though. Can update the copy to: "Are you sure you want to close `Walletname` and return to launcher?" or something in lines of that. And one could be added to the interaction for closing Decrediton to avoid any accidental quits.

vctt94:
I can work on this

alexlyp:
Great go for it.

PR #1762, opened 2018-10-31
Add close wallet button style

vctt94:
closes https://github.com/decred/decrediton/issues/1743

As I needed to rebase, I changed the toggle theme button place, this is how it looks right now: 

new


matheusd:
The only issue I see is that the text of the close button is slightly off center vs the specs.

vctt94:
Fixed

Issue #1665, opened 2018-09-27
Add network fund info under governance tab


jzbz:
I've been asked by several people recently where they can see the balance of the project treasury. Including the balance of the network fund and a link to it on dcrdata would be a useful feature in my view.

xaur:
Agreed, financial info like this helps stakeholders with their decisions making. Could be useful in CLI tools too.

alexlyp:
`curl -s https://explorer.dcrdata.org/api/address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx/totals -o - | jq '.dcr_unspent'`

This will return the current dev subsidy balance

jzbz:
This has actually been addressed and is on the bottom of the page now so I should have closed the issue.

alexlyp:
@jz afaik this has not been implemented in decrediton yet.

jzbz:
My bad, I thought I was commenting on dcrdata.

szpasztor:
@alexlyp I could take this issue - can you please add extra guidance as it's one of my first issues? Basically any info that's not in the decred docs already or steps to complete.

alexlyp:
yup go for it.  

So first I'd add a call in politeiapi to GET the subsidy from `https://explorer.dcrdata.org/api/address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx/totals`

Then once you have that call returning what you'd like you can add it to the getStartupWalletInfo() in ClientActions.js so it can be available to add to state (which will then let you add a selector and use in the component of your choice).

So start with the new restful call, and add it to that startup function. Open a PR with that completed (label it WIP) and then we can go from there to add to the Governance header.

matheusd:
You can also leave it for the second round of changes, but given this is the first access we'll add to dcrdata, we'll also need a new privacy option to enable access to it.

PR #1764, opened 2018-10-31
Show treasury balance under governance tab

szpasztor:
Resolves https://github.com/decred/decrediton/issues/1665

Subtasks
- [ ] get fund info from dcrdata
- [ ] new action in ClientActions.js
- [ ] add reducer & selector
- [ ] show under governance tab
- [ ] privacy option for dcrdata

Let me know if this is the right direction, and if yes, a bit more details about the remaining steps.


app/middleware/politeiaapi.js
```diff
@@ -45,5 +47,7 @@ export const getVoteStatus = (piURL) => GET(piURL, "/v1/proposals/votestatus");
 export const getProposal = (piURL, token) => GET(piURL, "/v1/proposals/" + token);
 export const getVoteResults = (piURL, token) => GET(piURL, "/v1/proposals/" + token + "/votes");
 
+export const getTreasuryInfo = (dcrdataURL) => GET(dcrdataURL, "/address/Dcur2mcGjmENx4DhNqDctW5wJCVyT3Qeqkx/totals");
```
szpasztor: Hardcoded the mainnet address for now, what's the testnet address?

matheusd:
It's going in the right direction. A few tips on the next steps:

> add reducer & selector

You need to modify the `reducers/grpc` file to use the new type of action you've added (`SET_TREASURY_BALANCE` in your code) so that the `treasuryBalance` is replicated in the main state. You then add a selector to grab that state var.

> show under governance tab

You need to modify the appropriate component to display this info somewhere, then modify the corresponding connector to pass the selector you created in the previous section.

>  privacy option for dcrdata

You'll need to modify the `main_dev/externalRequests` file to add a new constant for a new type of third party connection (you can modify it to be enabled by default in the standard privacy option). Then modify the settings page to include this as a new option.

szpasztor:
@matheusd The PR is ready for review. 

_Note:_ if the display currency is set to `Atoms`, it is possible that the balance shown is not an integer (due to the limitations of the IEEE 754 values that JavaScript uses for numbers). However, if this is a concern, it should be handled in another issue IMO.

vctt94:
needs rebase

alexlyp:
make sure to run `yarn lint-fix` as well.

alexlyp:
Hrm where is the fund amount shown?  Not seeing it currently.

alexlyp:
Ah nvm I needed to allow the new privacy setting.

app/actions/ClientActions.js
```diff
@@ -100,6 +104,9 @@ export const getStartupWalletInfo = () => (dispatch) => {
         await dispatch(findImmatureTransactions());
         await dispatch(getAccountsAttempt(true));
         await dispatch(getStartupStats());
+        if (dcrdataEnabled) {
+          await dispatch(getTreasuryBalance());
```
matheusd: You can drop the `await` from here, given that the treasury balance data is not critical to the wallet operation and if it fails (eg: due to dcrdata being offline or unreachable or if there's a change in the api that breaks compatibility) we should still be able to continue with the wallet startup.

szpasztor: Done

Issue #1651, opened 2018-09-14
Disable "About Decrediton" modal duplicate


linnutee:
<img width="1015" alt="screen shot 2018-09-14 at 23 36 36" src="https://user-images.githubusercontent.com/17774057/45573994-10f65c00-b877-11e8-896d-8949415cbd4f.png">

When opening "About Decrediton" from the op system navbar, it should display the same modal instead of the system default one.

alexlyp:
Hrm OK I'll look into this one, hopefully it's just a setting for mac builds or something.

linnutee:
Should be, I've checked some mac apps – e.g. telegram, spotify got it the same way. And in Skype the same action opens a view within a Settings modal even.

alexlyp:
This may be relevant to solving this issue: https://github.com/electron-userland/electron-builder/issues/2468#issuecomment-428641768

PR #1765, opened 2018-11-01
Disable "About Decrediton" modal duplicate

szpasztor:
Resolves https://github.com/decred/decrediton/issues/1651


szpasztor:
General method: use electron's `icpRenderer` to send events from main to render process and handle the click in the latter. 

Need help on how to proceed though. `AboutModal` should get its `show` prop from the newly added `aboutModalVisible` in state. How can I determine the connections between connectors and components? In this example, which connector should I use? Already tried searching for a given connector's usages but since their names are short, there are thousands of string matches.

matheusd:
This one is going to be a bit tricky.

I think you'll need to create the `<AboutModal` component directly in the `render()` method of the App container, based on whether the state you created is set or not. And you'll need to hook the `onCancelModal` so that the state is reset and the modal hidden.

alexlyp:
Any updates on this @szpasztor ?

szpasztor:
Rendering the `AboutModal` directly in the App container led to problems related to `ReactDOM.createPortal` receiving `null` as its domNode for some reason (might be related to the target domNode having the same parent dom element as the component); didn't have success after hours of debugging. I'll continue with a partial rewrite of AboutModal without using a portal.

szpasztor:
Update: found a solution by rendering two modals: one for the macOS navbar click and one for all other types of invocations (ie. macOS AboutModal button press and everything on Windows and Linux). The app verifies that at most one of the two kinds is rendered at any given time so there's no overlap. 

Tested on macOS 10.14.1. Not tested on Linux nor Windows. 

Ready for review @matheusd

szpasztor:
Here's the behavior I verified to be working: 
- Clicking navbar/about  (wallet select page or loaded app) => about modal shown
- Clicking decrediton's existing about modal button (wallet select page) => about modal shown
- Clicking navbar/about again without closing about modal first => nothing happens
- Clicking navbar/about with any modal open => nothing happens
- About modal can be exited in any of the above

Can we get someone to test for regressions on Windows?

PR #1766, opened 2018-11-01
[launcher] Add SPV Choice page

### LINKED ISSUE #1755: SPV enabling launch page
Preview: https://xd.adobe.com/view/7e93d6bd-7c03-4d24-6bc0-b6fca10a9a45-08a2/
Specs: https://xd.adobe.com/spec/76e88518-6127-4a05-5662-666c01547674-00d9/

Add SPV enabling launch page ( #1713 ). 
On the layout you can see hover effect for "Enable SVP" card. The choice gets applied when clicking on the card.

It uses the similar pattern as the "Privacy options" page does:
image

Also need to write a correct copy for it, as the current one is a placeholder (showed on the screenshot below texts that have to be changed)
image

alexlyp:
Close #1755 

Fully functional, but still trying to finalize some of the copy on the explanation of SPV vs Regular.


alexlyp:
Agree on that.  I'll add an issue to address this in the future.

Issue #1775, opened 2018-11-05
Issues with timezones in Decrediton 1.3.1


rocknet:
I've seen multiple issues with timezone processing today in Decrediton 1.3.1.  The hours are not adjusted properly from UTC for me to US ET (-0500 currently but all my txns are during -0400).  When trying to troubleshoot this I've found a few issues, I'll use this txn as an example: [https://explorer.dcrdata.org/tx/380b4ed5726c00f87893c158fa0c8ec2c3cc6ef9ad8cf616a02255e797ed8826](https://explorer.dcrdata.org/tx/380b4ed5726c00f87893c158fa0c8ec2c3cc6ef9ad8cf616a02255e797ed8826)

Environment:
Win7 64Bit SP1
Today's date: November 4, 2018
US ET (we changed from UTC-4 to UTC-5 early this morning) https://www.timeanddate.com/time/zone/usa/new-york

Steps to reproduce
With Timezone in Settings set to "Local", open the Transactions -> History tab
Find transaction with the has specified above in my transaction history (though the problem is consistent with seemingly all transactions)
Note: For convenience, this is block 232673, with a timestamp of: 1524540145
https://explorer.dcrdata.org/api/block/232673/verbose?indent=true

Expected:
Considering the explorer shows this occurred at 3:22:25 AM UTC on April 24, 2018, I expect this to show 11:22:25 PM on April 23, 2018 because as of that date ET was -0400.

Actual:
I see the transaction with the hash above with a datetime of April 24, 2018 12:22:25 AM.  This is a difference of three hours, but US ET is never 3 hours from UTC.

Further step:
In Decrediton, click the transaction to see the details

Expected
The field next to "Purchased On" has the date and time

Actual
![image](https://user-images.githubusercontent.com/1266404/47972370-a5a45b80-e069-11e8-84ce-8a1a27505b3b.png)
Note: This appear to only be an issue with transaction type: "Voted"

Further step
Change the timezone to UTC
Export the transactions from the Export tab

Expected
Times are in UTC with no UTC offset specified

Actual
Times are the actual UTC time values, but have a local offset (-0400 for all transactions I have), which makes the times incorrect once the offset is added.

rocknet:
Thanks.  I'll see if I can build master and try it out, that'd be later tonight for me.

[...]

Issue #1421, opened 2018-05-25
Fix Buffer() deprecated node 10 warnings


matheusd:
Biggest offender (in line count) is the generated grpc code.

There are some libraries also using this code, so it will be quite a job to track down each one, open issues upstream, wait for them to be fixed, merged and released then update the dependency.

As a start, we should list the offending libraries here to track down exactly each one with a problem.

alexlyp:
Thanks for opening this issue.  Hopefully most libs will be making the updates soon.

matheusd:
See https://github.com/grpc/grpc-node/issues/670 for grpc.

vctt94:
fixed

PR #1777, opened 2018-11-06
[deps] Update to Electron 3

matheusd:
Updates to the latest electron (3.0.6) and grpc (1.16.0).

It also adds a config option to disable hardware acceleration, since I found that on some OS versions the new electron version fails to properly render when hardware acceleration is enabled.

Tested on linux and windows.

Part of #1775.


matheusd:
Yeah, there's an open issue for that (#1421).

PR #1778, opened 2018-11-06
[launcher] Add get info

### LINKED ISSUE #1770: [launcher] Add getinfo call to dcrd to check network
There is a chance for a user to get stuck if they have `testnet=1` set in their dcrd.conf but decrediton is set to use mainnet.

To avoid this we could use a dcrctl `getinfo` call to make sure the network settings match.  And if they don't send the user to a new Error page describing what they have to do to fix (remove `testnet=1` from their dcrd.conf)

vctt94:
closes #1770

### DIFF
--- a/app/components/views/FatalErrorPage.js
+++ b/app/components/views/FatalErrorPage.js
@@ -1,11 +1,10 @@
 import { FormattedMessage as T } from "react-intl";
 import { fatalErrorPage } from "connectors";
 import { KeyBlueButton, RemoveDaemonButton } from "buttons";
+import { DIFF_CONNECTION_ERROR } from "main_dev/constants";
 import "style/Layout.less";
 
 const resourcesUnavailableError = "resource temporarily unavailable";
-
-const diffConnection = "daemon and decrediton does not have the same network";
 
 const corruptedError = "corrupted";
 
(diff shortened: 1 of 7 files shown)


app/components/views/FatalErrorPage.js
```diff
@@ -28,19 +29,21 @@ class FatalErrorPage extends React.Component {
...
-                  <RemoveDaemonButton
-                    className="fatal-remove-button"
-                    modalTitle={<T id="fatal.removeConfirmModal.title" m="Remove daemon data"/>}
-                    modalContent={<T id="fatal.removeConfirmModal.content" m="Warning this action is permanent! Please make sure you want to remove your blockchain data before proceeding. Decrediton will automatically shutdown after deleting the folder. Please manually restart it afterwards."/>}
-                    onSubmit={deleteDaemonData}
-                    buttonLabel={ <T id="fatal.button.delete" m="Delete and Shutdown"/>}/>
-                </Aux> :
-                <T id="fatal.suggestion.fallthrough" m="Please note the error above and go to the support channel on slack/matrix/rockchat for help resolving the issue." />
+              daemonError && daemonError.indexOf(DIFF_CONNECTION_ERROR) !== -1 ?
+                <T id="fatal.suggestion.diffConnection" m="This error typically means you have the testnet flag on your dcrd.conf file. You should check your dcrd.conf file and remove the testnet=1." /> :
```
alexlyp: "in your dcrd.conf"

app/main_dev/ipc.js
```diff
@@ -137,6 +137,55 @@ export const stopWallet = () => {
...
+
+  const spawn = require("child_process").spawn;
+  const dcrctl = spawn(dcrctlExe, args, { detached: false, stdio: [ "ignore", "pipe", "pipe", "pipe" ] });
+
+  dcrctl.stdout.on("data", (data) => {
+    const parsedData = JSON.parse(data);
+    const isTestNet = parsedData.testnet;
+    logger.log("info", "isTestnet: " + isTestNet);
+
+    mainWindow.webContents.send("check-getinfo-response", { isTestNet });
```
alexlyp: I'd probably just return all the data in case we want to use it later.  Then you can pull the isTestNet downstream.

vctt94:
fixed

PR #1785, opened 2018-11-08
Update deprecated react code

vctt94:
This PR updates the code that are being deprecated in react 17. 

The methods `componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate` are going to be deprecated at react 17: 
https://reactjs.org/docs/react-component.html 

Where was needed to do more code refactor then simple changing the props, I committed separated, to make code review easier. 

Also this PR still a work in progress because I could not test `app/components/views/TicketsPage/PurchaseTab/StakePools/index.js` because I am not being able to reimport an API key into my wallet. 

Additionally our snackbar uses componentWillReceiveProps, and I think it would be better to use memoization, one of the methods that is suggested to replace the  componentWillReceiveProps method. Here more about it:
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

I think it would be better to discuss that before start coding those changes into the snackbar.

### DIFF
--- a/app/components/SideBar/LastBlockTime.js
+++ b/app/components/SideBar/LastBlockTime.js
@@ -8,10 +8,10 @@ class LastBlockTime extends React.Component {
     this.state = this.getBlockDate(props.lastBlockTimestamp);
   }
 
-  componentWillReceiveProps(nextProps) {
-    const { lastBlockTimestamp } = this.props;
-    if (lastBlockTimestamp !== nextProps.lastBlockTimestamp) {
-      this.setState(this.getBlockDate(nextProps.lastBlockTimestamp));
+  componentDidUpdate(prevProps) {
+    const { lastBlockTimestamp } = prevProps;
+    if (lastBlockTimestamp !== this.props.lastBlockTimestamp) {
+      this.setState(this.getBlockDate(this.props.lastBlockTimestamp));
     }
   }
 
--- a/app/components/buttons/ModalButton.js
+++ b/app/components/buttons/ModalButton.js
@@ -30,8 +30,8 @@ class ModalButton extends React.Component {
     onShow && this.props.onShow();
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.showModal && !this.props.modal) {
+  componentDidUpdate(prevProps) {
+    if (this.props.showModal && !prevProps.modal) {
       this.showModal();
     }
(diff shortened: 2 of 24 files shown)


matheusd:
Do you have a link with the proposed changes for react 17 or a doc describing the rationale for those deprecated functions?

vctt94:
hey @matheusd 

I believe the discussion is happening at react's github and I believe it is because of the wrong use and side effect of them. 

Here is a text which explains about async rendering: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html  And the misusing of those methods. From the text:
 
`These lifecycle methods have often been misunderstood and subtly misused; furthermore, we anticipate that their potential misuse may be more problematic with async rendering. Because of this, we will be adding an “UNSAFE_” prefix to these lifecycles in an upcoming release. (Here, “unsafe” refers not to security but instead conveys that code using these lifecycles will be more likely to have bugs in future versions of React, especially once async rendering is enabled.)`

But I am not following close this discussion, so it is hard to me to say.

matheusd:
Thanks for the references, that was exactly what I needed to see to understand the changes.

So, what's happening is that they are deprecating these methods to reduce possible bugs and allow async rendering.

We should use this opportunity to not just rename componentWillUpdate to componentDidUpdate, but to actually review the underlying usage pattern in each component and update it accordingly (eg: to use getDerivedStateFromProps). [This](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html) is another important read on this topic.

So, my next question is: are the changes in this WIP all instances where the deprecated methods are being used (apart from the snackbar you already mentioned)?

vctt94:
I am not sure if I understand your question

matheusd:
What I mean by that is if you have replaced **all** locations in the code where `componentWillReceiveProps` etc are being used.

vctt94:
Yes, it does.
But https://github.com/decred/decrediton/pull/1793 also tickle on the componentWillReceiveProps.

matheusd:
Can you rebase please? Now that I've understood the reasoning behind the change, I'll test it out.

vctt94:
Rebased

vctt94:

It was a little more difficult than I first thought, because when children's props were updating the component was not re-rendering, but I fixed that as well.

PR #1797, opened 2018-11-14
[Governance] Use tabbedPage at proposals

vctt94:
This PR is a clean up at governance page to also use tabbedpage


vctt94:
fixed

PR #1806, opened 2018-11-16
[logs] Show decrediton logs at logs tab

### LINKED ISSUE #1050: View Decrediton Logs
Besides the dcrd/dcrwallet logs, we should also be able to see Decrediton logs on production as that may help diagnose problems.

vctt94:
This PR closes #1050 .

I changed the timeout which makes the logs auto update themselves, to update when clicking to show the log.

The way it was before, when opening the decrediton logs, it would spam the `getLogs` message on the decrediton log every 2 seconds.

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -12,6 +12,7 @@ import { setupProxy } from "./main_dev/proxy";
 import { cleanShutdown, GetDcrdPID, GetDcrwPID } from "./main_dev/launch";
 import { getAvailableWallets, startDaemon, createWallet, removeWallet, stopDaemon, stopWallet, startWallet, checkDaemon, deleteDaemon, setWatchingOnlyWallet, getWatchingOnlyWallet, getDaemonInfo } from "./main_dev/ipc";
 import { initTemplate, getVersionWin, setGrpcVersions, getGrpcVersions, inputMenu, selectionMenu } from "./main_dev/templates";
+import { readFileBackward } from "./helpers/byteActions";
 
 // setPath as decrediton
 app.setPath("userData", appDataDirectory());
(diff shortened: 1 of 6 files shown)


alexlyp:
Good start.

1) So we intentionally did that interval so the logs would update when you are looking at them. Otherwise you have to click away to get any log update.  I'm OK to remove if others agree that the UX without it is proper.
2) You need to somehow limit the amount of logs shown/returned to the user for decrediton.  Right now for me it's loading a log file that was started Feb 18, so it basically freezes the application trying to display it.  I believe you can look at the logging stuff for dcrwallet dcrd for ideas about to limit.  2000 lines or so should be sufficient.

vctt94:
Thanks. 
1.  Maybe we can have the interval for the other logs, but I believe it can be a problem for the decrediton logs, as it will keep spamming: 
```
2018-11-16 18:16:04.707 [INF] DCTN: Get Dcrd Logs 
2018-11-16 18:16:04.708 [INF] DCTN: Get Dcrwallet Logs 
2018-11-16 18:16:04.709 [INF] DCTN: Get Decrediton Logs 
```  
Making the log file bigger just letting the logs tab opened. 


  2. Sure, I'll do that.

alexlyp:
Ah I see what you're getting at.  Hmmm maybe we can just silence those requests from being logged into decrediton.log

vctt94:
I think that work too :)

app/main.development.js
```diff
@@ -195,7 +195,21 @@ ipcMain.on("get-dcrwallet-logs", (event) => {
 });
 
 ipcMain.on("get-decrediton-logs", (event) => {
-  event.returnValue = "decrediton logs!";
+  const logFileName = logger.transports.file.dirname + "/" +logger.transports.file.filename;
+  fs.readFile(logFileName, (err, data) => {
```
matheusd: Yeah, I don't really like this approach. This is reading the entire decrediton log file into memory every time this ipc endpoint is requested (ie, every 2 seconds once the user accesses the log page).

An improvement over this would be to use `fs.Open` and a function to copy data backwards from the end of the file (I don't know if there's anything ready for that, so you'd need to do a bit of research or implement it manually using the fs module).

Another option would be to hook the `main-log` IPC receiver, and add to a memory buffer that will hold the last log lines (like what is done for dcrd/dcrwallet).

vctt94: fixed. This PR is ready for another review

PR #1816, opened 2018-11-20
[icons] Update menu with hovering animations

### LINKED ISSUE #1809: Visual/Motion Design: Update Navigation icons and microanimations
Alrighty, thanks to @kyleFirethought we've got a set of microanimations to cover almost entire icon and symbol set. Will be creating more issues to get these implemented in Decrediton. 

When working on implementing the microanimations, first check through this readme for better understanding: https://medium.com/@firethought/base-iconset18-motion-pack-readme-a96f96e868 

Preview and instructions: https://xd.adobe.com/view/fe4da41a-f315-40d0-6ee7-059f0fb842f5-887c/ 
Assets: https://www.notion.so/eeter/Decrediton-Navigation-microanimations-11cfd35653fd4a84bfb861a63ebddeed

Brief insight: most of the animations have either a `loop` or `open` or both variants available. Loop means the animation starts from a complete icon and end in the same form (looping playback). Open is a type of transition where the animation starts from an empty or partial state and ends into a complete icon (should only play back one time). 

For the navigation all microanimations used are **only loops**. Meaning when the user hovers on a nav icon, the loop will start and play until: 
a) user no longer hovers on the icon 
b) user clicks on the item, loop plays till end and displays the still active state  

I've included both gif, png sequence and lottie json with only `loop` variants as well the `static` versions of both default and active icon states (/png folder). Implement which works best.  

Also from the visual design front:
1  Updated Accounts icons metaphor as this has now for a while been in conflict with the Wallet icon. New one refers to a card and a generative DCR "D…" address, adding up to the directly usable accounts within wallets. 

2 Updated the Total balances icon to a more relevant abacus metaphor 
screenshot 2018-11-19 at 16 39 47

Ref to /dcrdesign issue: https://github.com/decred/dcrdesign/issues/55

alexlyp:
Close #1809


alexlyp:
Agree with @matheusd and really this comes down to not being about to implement the lottie versions of the animations, imo.  With lottie the icons would immediately start to react on hover instead of what's happening currently (which I believe is that the gifs "play" in the background while not being shown, so sometimes the gifs will be through their mid-animations rather than at the beginning, or ideally from the place where the user last unhovered.)

PR #1821, opened 2018-11-21
[overview] Update legend for overview charts

### LINKED ISSUE #1818: Visual Design: Overview improvements
Preview: https://xd.adobe.com/view/147f12c2-e81b-43cb-429a-08f74a33290b-624e/
Specs: https://xd.adobe.com/spec/ebb53c77-fcde-44db-5236-3bf9ca1ed0e5-9d84/

Small improvements to make the top area a bit more coherent: 
- Added color-coded legends in front of chart value summaries. 
- Switched Sent/Received to match positions displayed on the chart.
- Created a new and simplified summary for the Tickets tab. Only displaying the amount of tickets as larger summarised numbers. The count for "active tickets" stands for sum of "purchased" and "live" tickets (hence the split circle). Total DCR values replaced into the info text and applied the blue highlight pattern representing changing values. 

Legend svg's: [overviewLegends.zip](https://github.com/decred/decrediton/files/2605312/overviewLegends.zip)

### LINKED ISSUE #1819: Navigation: Hierarchy update
Preview: https://xd.adobe.com/view/6174e5ff-32bb-4c2c-6903-f3612061bd89-1745/

Since Governance is a new feature having it further up should affect the users positively towards exploring the proposals. Governance similarly to Transactions is a more interactable view right of out of box, unlike Tickets has no (or less) requirements for use. Tickets in a sense is more similar to accounts, having higher requirements for use, but once setup is more actively used. Thus should stay right above it. (Additionally: In the future I would also consider playing around with combining Accounts and Tickets to reduce the nr. of nav items, but only if the "staking account" or some similar solution can be implemented.)

Settings should be the last item in the navigation instead of help, as it commonly tends to be in either the same or otherwise similarly isolated and clear position in software and websites. 

Proposed update:
- Switch positions of Governance and Tickets, leaving Governance as the third item. 
- Switch positions of Settings and Help, leaving Settings as the last item.

alexlyp:
Close #1818 
Close #1819 

Also includes an ordering change to the sidebar menu.

### DIFF
--- a/app/components/SideBar/MenuLinks/index.js
+++ b/app/components/SideBar/MenuLinks/index.js
@@ -7,12 +7,12 @@ import theme from "theme";
 const linkList = [
   { path: "/home",          link: <T id="sidebar.link.home" m="Overview" />,             icon:"overview" },
   { path: "/transactions",  link: <T id="sidebar.link.transactions" m="Transactions" />, icon:"transactions" },
-  { path: "/tickets",       link: <T id="sidebar.link.tickets" m="Tickets" /> ,          icon:"tickets" },
   { path: "/governance",    link: <T id="sidebar.link.governance" m="Governance" />,     icon:"governance" },
+  { path: "/tickets",       link: <T id="sidebar.link.tickets" m="Tickets" /> ,          icon:"tickets" },
   { path: "/accounts",      link: <T id="sidebar.link.accounts" m="Accounts" />,         icon:"accounts" },
   { path: "/security",      link: <T id="activesidebar.link.security" m="Security" />,   icon:"securitycntr" },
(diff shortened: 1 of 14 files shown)


vctt94:
Wait, looking at it again, is it right to the locked part of a ticket be negative on the chart?

alexlyp:
Yah that looks like a bug that I thought was patched.  Can you open an issue for that?

Yah I believe there is an open issue about the graphs.  This would require some hacking on the graphing package itself.  It needs to deteremine that it is the "bottom" portion of the stack and then round that corner.  Don't think it'll be a super easy fix, but haven't really gotten in to the regraphs code much.

PR #1830, opened 2018-11-24
Add memoization to snackbar

vctt94:
This PR finishes to change the code that is being deprecated in react 17.

I added memoization because I followed the suggestion described here: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

### DIFF
--- a/app/components/Snackbar/index.js
+++ b/app/components/Snackbar/index.js
@@ -5,6 +5,7 @@ import ReactTimeout from "react-timeout";
 import EventListener from "react-event-listener";
 import Notification from "./Notification";
 import theme from "theme";
+import memoize from "memoize-one";
 import { eventOutsideComponent } from "helpers";
 import { spring, TransitionMotion } from "react-motion";
 import { TRANSACTION_DIR_SENT, TRANSACTION_DIR_RECEIVED,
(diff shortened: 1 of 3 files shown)


app/components/Snackbar/index.js
```diff
@@ -40,14 +41,22 @@ class Snackbar extends React.Component {
...
+      return;
+    }
+    if (this.props.messages.length > 0) {
       this.enableHideTimer();
     }
 
     const messagesByKey = keyBy(this.state.messages, "key");
-    const newMessages = nextProps.messages.map(m => messagesByKey[m.key] ? messagesByKey[m.key] : m);
-    this.setState({ messages: newMessages });
+    const messages = this.newMessages(this.props.messages, messagesByKey);
```
matheusd: This particular memoization application isn't really necessary.

Notice: You've added an `if (prevProps.message === this.props.messages)` guard at the start of the function, which means that the `this.props.messages` in this line will *always* be a different object (otherwise the call would have exited early).

Also notice that the `keyBy` invocation will generate a new object every time it's called, which means that `messagesByKey` in this line will also be different in every invocation.

This means that the memoized value will *never* be used.

The point of memoization is so that you don't have to care whether the parameters have changed, the memoization function taking care to "remember" the values in according to whatever cache strategy it's using (eg: `memoization-one` will store the last set of parameters and the last returned value).

If you want to specifically add memoization, then `componentDidUpdate` is not the best place to do this: you probably want to add it to the `render()` method directly, passing `this.props.messages` and `this.state.messages` directly, then calculate both the `messagesByKey` and `newMessages` inside the memoization function. But then you'll need to update the state based on the result of this call (whether it changed the final `messages` map or not) so I don't even think memoization is the right thing to do here (without a larger overall structure change to snackbar).

By the way, the way you implemented `componentDidUpdate` (even with the recalculation of `newMessages`) is already performing the correct caching of the state (due to the initial `if` you added). You can test this by adding a couple of log lines here, and in the `render()` method, then add a `this.setState({ dummy: Math.random() })` inside `clearHideTimer()` so that a re-render can be triggered by moving the mouse inside/outside the messages. If you do this, you'll notice that the end of `componentDidUpdate` is only reached if the list of messages changes (either due to a new message coming in or an old message going out of the list).

vctt94: Thanks for the clarification, I removed the memoization

Issue #1808, opened 2018-11-19
Notify user of new Politeia votes


jzbz:
When a user opens their wallet they should ideally be notified if there are new proposals that have entered the voting stage. To get really fancy, they would only get pinged if they have tickets that are eligible to vote.

vctt94:
Notification has been implemented. Not sure about the eligible tickets part. If so, can we close this one?

PR #1835, opened 2018-11-26
[politeia] Notifications for new proposals and votes

### LINKED ISSUE #1832: Interaction design: Governance notifications
Preview: https://xd.adobe.com/view/ecc8fa7c-65fe-456e-5cf4-4524881ff4fd-9af9/
Specs: https://xd.adobe.com/spec/53c7c738-eb51-47d9-4e7f-e33ca991fceb-25f6/
Icon: [bellNotification.svg.zip](https://github.com/decred/decrediton/files/2612486/bellNotification.svg.zip)

A three-level notification to inform users of Pi related activities. 

**Bell-icon in the navigation:** 
Since this is a first time introduction, the bell-icon should only trigger if a new proposal enters voting state. The bell-icon is removed when user opens up the new proposal (regardless of voting or not)

For this release, I would not trigger the bell-icon in navigation when a new proposal is added "under-discussion" as this may lead to it being ever-present. However this may be changed once it's been in use for a while. 

**Counter-circles** are present for both "under-discussion" and "under vote" whenever a new item is added which has not yet been opened. Count is a sum of new proposals not viewed.

**Row-color** is also present for both "under-discussion" and "under vote" proposals whenever a new item is added which has not yet been opened/viewed.

matheusd:
Close #1832 
Part of #1808 

This
- Changes to fetch politeia proposals after wallet startup without having to enter the governance page
- Adds two new config entries to store the latest timestamp and block height when politeia data was successfully fetched
- Uses those two config items to determine whether a vetted or active vote proposal are _new_ ones (either newly vetted or newly voting)
- Shows the appropriate notification ui

There's an outstanding issue due to https://github.com/decred/politeia/issues/618: Given we only have the timestamp for proposal creation (vs proposal vetting for discussion) this version has a bug in that proposals that have been vetted after the user has fetched the latest batch of proposals will not be marked as `modifiedSinceLastAccess` (because the user has already passed through the block height of the proposal _creation_).

Alternatives for this are:

- Wait for https://github.com/decred/politeia/issues/618 to be solved, and than change which timestamp is used to determine new proposals (preferred method)
- Locally store the proposals that have been seen and check after loading the proposal list whether this has been seen


alexlyp:
I think it's reasonable to just wait until 618 is solved to determine the new proposals

PR #1838, opened 2018-11-27
[stakepools] Refactor actions and apikey entry ux

### LINKED ISSUE #1773: [1.3.1] Dont reset stakepool choice if API key is incorrect
Steps to recreate:
- Create a new wallet
- Open the tickets tab
- Select a stakepool which is not the default already selected
- Enter an invalid API key (any old string)
- Press continue
- Enter valid passphrase 
- Press continue

Expected result:
- Error is shown 
- Stakepool choice is remembered

Actual result:
- Error is shown 
- Stakepool choice is reset back to the default

Note: This same behaviour is also observed if the API key is correct, but the passphrase is wrong

matheusd:
Close #1773 

This changes the interaction of stakepool api key entry to preserve the api key and host selection if an error (such as wrong passphrase or wrong api key) is triggered. 

It also refactors the stakepool actions in the following ways:

- Use async/await calls to reduce code/indentation depth
- The importScriptAttempt() function now no longer triggers success/error snackbars (which removes the need for many conditionals and parameters throughout stakepool import/ticket purchasing actions)
- A manualImportScriptAttempt() function was added for manual script import (which does trigger snackbar messages and rescans)
- the setStakepoolInfo() call now does some of the checks that were done deeper in the call stack (removing the need to propagate arguments)
- The snackbar has been changed to intercept the error signature of wrong passphrase errors and show a standard message on that case (instead of requiring parsing all over the place)

There are some outstanding snackbar issues and improvements that I can see, but I'll leave them for a second PR.

### DIFF
--- a/app/reducers/snackbar.js
+++ b/app/reducers/snackbar.js
@@ -12,7 +12,7 @@ import {
   STARTTICKETBUYERV2_SUCCESS, STARTTICKETBUYERV2_FAILED,
   STOPTICKETBUYERV2_SUCCESS,
   REVOKETICKETS_SUCCESS, REVOKETICKETS_FAILED,
-  IMPORTSCRIPT_SUCCESS, IMPORTSCRIPT_FAILED,
+  IMPORTSCRIPT_MANUAL_SUCCESS, IMPORTSCRIPT_MANUAL_FAILED,
   RENAMEACCOUNT_SUCCESS, RENAMEACCOUNT_FAILED,
   GETNEXTACCOUNT_SUCCESS, GETNEXTACCOUNT_FAILED,
   CHANGEPASSPHRASE_SUCCESS, CHANGEPASSPHRASE_FAILED,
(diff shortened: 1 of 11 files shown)


app/reducers/snackbar.js
```diff
@@ -214,6 +221,10 @@ const messages = defineMessages({
   STOPTICKETBUYERV2_SUCCESS: {
     id: "stopTicketBuyer.Success",
     defaultMessage: "Ticket Buyer successfully stopped."
+  },
+  WRONG_PASSPHRASE_MSG: {
+    id: "errors.wrongPassphrase",
+    defaultMessage: "Wrong private passphrase entered. Please verify you have typed the correct private passphrase for the wallet."
```
alexlyp: Everything looks good, but I'm seeing this message get cut in the middle of a "the".  So it show t on the top row and he on the bottom.  Maybe needs to have a different word wrap on the snackbar div?

matheusd: Yeah, that's one of the things I'm planning to address on a separate snackbar PR (fix the word wrap).

This is likely to need changes in the individual message components of the snackbar, so I wanted to leave it for another PR, but if you don't have objections I can add another commit here for it.

alexlyp: Ah gotcha. Cool

PR #1839, opened 2018-11-27
Fix negative locked

### LINKED ISSUE #1823: [overview - charts] Locked value shows as negative
On the overview charts, the locked value are showing as negative, but they are supposed to be as positive.

Here is an image:

needchange

vctt94:
This PR closes #1823 .

It also clean up a bit the `StatisticsActions`.

I am starting the clean up by separating the methods from `balancesStats`, making them more independent, and separating the backwards logic from the straight forward one.

The objective behind that is to make it easier to understand and maintain the statistics logic. 

This still a WIP because I still have to check if all the exports methods are right, and there are some more clean up to do, but I am opening this PR now so you guys can see it.

The immature and locked values on daily balance looks wrong on master and this PR also fixes it

### DIFF
--- a/app/actions/StatisticsActions.js
+++ b/app/actions/StatisticsActions.js
@@ -114,7 +114,7 @@ export const EXPORT_COMPLETED = "EXPORT_COMPLETED";
 export const EXPORT_ERROR = "EXPORT_ERROR";
 
 export const exportStatToCSV = (opts) => (dispatch, getState) => {
-  console.log(opts)
+  console.log(opts);
   const { calcFunction, csvFilename } = opts;
 
   var fd;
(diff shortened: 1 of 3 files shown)


matheusd:
I'm not sure moving the internal methods out of the balance calc stats is so important, but if you feel strongly enough about it, then sure.

You might wanna test/check with https://github.com/decred/dcrwallet/pull/1330 given that it modifies how the lockedByTickets/total balances work (do notice that PR performs a database migration on the wallet, so you might wanna backup your wallet dirs first).

vctt94:
I think this way was easier because I can use the same methods on the backwards and the forward logic. If you have suggestions on ways to make the code cleaner, they can be helpful, as well. 

I will test with this PR, thanks for informing.

PR #1840, opened 2018-11-29
[snackbar] Small improvements

matheusd:
A few small code and ui improvements for the snackbar.

I was going to solve the text overflow issue in one way which I thought would be a larger bit of code, but ended up finding a simpler solution.


matheusd:
Messages after fixing the text overflow issue:

screenshot_2018-11-29_09-33-48

PR #1844, opened 2018-11-29
[governance] Various updates to proposal page styling

### LINKED ISSUE #1831: Visual Fixes/Notices: Governance/Proposals subpage
Not sure if these might be in-progress still, but will mention a few bugs noticed from latest test-build: 

Preview: https://xd.adobe.com/view/c79b437f-befe-447c-640a-21d942fc7a3e-b1aa/

1. There is a 1 px glitch between the bottom of white header bg and the content area grey bg.

2. Second level of grey (isolating the proposal states) is missing. Note the difference, that "Proposals" subpage has three steps of grey in the content area background, whilst "Consensus changes" has two (this is correct currently).  

3. "Under Discussion" "Under Vote" and "Finished Voting" uses the subpage nav. pattern. This should have its own pattern. 

4. Right edge of the first grey step ends unevenly on the right side. 

5. Under "Finished Voting" the row color codings have incorrect green value (using cyan currently) and have some triangular glitches in the bottom. 

6. All subviews have currently #00000 black text. Can check specific values from :https://github.com/decred/decrediton/issues/1759

alexlyp:
Close #1831

### DIFF
--- a/app/components/layout/TabbedPage.js
+++ b/app/components/layout/TabbedPage.js
@@ -127,7 +127,7 @@ class TabbedPage extends React.Component {
         <div className="tabbed-page-header">
           {header}
           <Switch>{headers}</Switch>
-          <RoutedTabsHeader tabs={tabHeaders} />
+          <RoutedTabsHeader tabs={tabHeaders} noCaret={this.props.noCaret}/>
         </div>
 
         <div className={"tabbed-page-body" + (this.props.governancePage ? " governance" : "")}>
--- a/app/components/layout/TabbedPage.js
+++ b/app/components/layout/TabbedPage.js
@@ -122,7 +122,6 @@ class TabbedPage extends React.Component {
     );
 
     const tabContents = this.props.uiAnimations ? this.animatedStyles() : this.staticStyles();
-
     return (
       <div className="tabbed-page">
         <div className="tabbed-page-header">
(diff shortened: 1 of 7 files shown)


app/style/Themes.less
```diff
@@ -71,8 +71,8 @@
     --error-text-color: #fd714b;
     --error-red: red;
     --error-red-hover: red;
-    --vote-yes-color: #2ed8a3;
-    --vote-no-color: #f2734f;
+    --vote-yes-color: #41be53;
+    --vote-no-color:#ed6d47;
```
matheusd: missing space

app/components/layout/TabbedPage.js
```diff
@@ -122,16 +122,15 @@ class TabbedPage extends React.Component {
...
       <div className="tabbed-page">
         <div className="tabbed-page-header">
           {header}
           <Switch>{headers}</Switch>
-          <RoutedTabsHeader tabs={tabHeaders} />
+          <RoutedTabsHeader tabs={tabHeaders} noCaret={this.props.noCaret}/>
         </div>
 
-        <div className="tabbed-page-body">
+        <div className={"tabbed-page-body" + (this.props.governancePage ? " governance" : "")}>
```
matheusd: Hmm, I'd say that instead of a "governancePage" props this should either be a generic `className` prop appended to this divs classes or a "AltCaret" (or "HightlightCaret", "FullCaret", "NoCaret" or something like that) so that this isn't just a governance page thing.

alexlyp: Ah good point, will make the adjustments

PR #1864, opened 2018-12-04
[watch only] Watch only transaction

### LINKED ISSUE #1682: Allow watch only wallets to construct transactions
Maybe this could be done by changing the send button in the send page to a "construct" button that just shows the most recently constructed tx.

This would allow decrediton to produce a tx that could be saved and moved to a separate computer for signing.

vctt94:
closes #1682 

I need to rebase, so if can test in trezor mode, that would be great.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -5,7 +5,7 @@ import { isValidAddress, isValidMasterPubKey } from "helpers";
 import { getStakeInfoAttempt, startWalletServices,
   getStartupWalletInfo } from "./ClientActions";
 import { RescanRequest, ConstructTransactionRequest, RunTicketBuyerRequest } from "../middleware/walletrpc/api_pb";
-import { reverseRawHash } from "../helpers/byteActions";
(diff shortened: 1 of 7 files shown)


matheusd:
Only small issue is that in trezor-enabled wallets, you shouldn't show the unsigned transaction, but rather continue with the signing flow (which should send the sign request to trezor).

You can manually set a wallet as using trezor (to test out the behavior, even if you can't sign it) by modifying the wallet's `config.json` (_not_ the global one) to have `trezor: true`.

vctt94:
fixed

matheusd:
I'm still seeing the raw constructed transaction on the trezor wallet, which is unnecessary.

vctt94:
My bad. Removed it now.

One thing I noticed, is that when in trezor mode, the wallet seems to not turn it off, when switching between wallets.

matheusd:
Yeah, I had noticed that bug earlier in regards to the sidebar. Fixed in #1876.

One last nit: transactions can be big and overflow the white area (eg):

screenshot_2018-12-07_11-59-42

vctt94:
fixed

PR #1875, opened 2018-12-07
[i18n] Translations for 1.4.0rc

matheusd:
(leaving this here so we don't forget to merge before kicking off the 1.4.0rc builds)

Updated translations for
  - pt
  - zh (still missing trezor strings)


matheusd:
good to go

PR #1879, opened 2018-12-07
[governance] Add abandoned proposal list

### LINKED ISSUE #1872: Abandoned proposals still listed as Under Discussion
screenshot at dec 06 21-24-41

alexlyp:
Close #1872

### DIFF
--- a/app/actions/GovernanceActions.js
+++ b/app/actions/GovernanceActions.js
@@ -9,6 +9,7 @@ import { hexReversedHashToArray, reverseRawHash } from "helpers";
 // enum values from politeiawww's v1.PropVoteStatusT
 export const VOTESTATUS_ACTIVEVOTE = 3;
 export const VOTESTATUS_VOTED = 4;
+export const VOTESTATUS_ABANDONED = 0;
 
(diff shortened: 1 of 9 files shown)


app/actions/GovernanceActions.js
```diff
@@ -9,6 +9,7 @@ import { hexReversedHashToArray, reverseRawHash } from "helpers";
 // enum values from politeiawww's v1.PropVoteStatusT
 export const VOTESTATUS_ACTIVEVOTE = 3;
 export const VOTESTATUS_VOTED = 4;
+export const VOTESTATUS_ABANDONED = 0;
```
matheusd: Is this right? On politeiawww "abandoned" is defined as having status === 6:

https://github.com/decred/politeia/blob/master/politeiawww/api/v1/v1.go#L198

alexlyp: Hrm yah looks like it should be 6, but for whatever reason I am getting 0 back from the API.  Let me confirm.

alexlyp: So getVetted is giving me them properly with 6, but then they aren't included in voteStatus response which results in a 0 instead.

PR #1893, opened 2018-12-10
[launcher] Add log check for 1.4.0 dcrd reindexing

### LINKED ISSUE #1880: [launcher] Notify users of long wait for upgrade from <1.3
Due to the dcrd reindexing that has been included in 1.4.0 users upgrading from 1.3 will encounter a long wait until dcrd is prepared for usage.  

We should capture specific lines in the logs that will let us know that they are currently reindexing.  This isn't ideal, but unfortunately since the dcrd's rpc client isn't available until after the upgrade is complete.

alexlyp:
Close #1880

### DIFF
--- a/app/components/views/GetStartedPage/DaemonLoading/Form.js
+++ b/app/components/views/GetStartedPage/DaemonLoading/Form.js
@@ -109,7 +109,7 @@ const DaemonLoadingBody = ({
             </div>
             <div className="advanced-page-form">
               <div className="advanced-daemon-row">
-                <T id="getStarted.longWaitWarning" m="You are currently upgrading to dcrd version 1.4.0.  Typically, this one-time reindexing take 30-45mins on an average machine."/>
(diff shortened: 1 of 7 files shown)


app/components/views/GetStartedPage/DaemonLoading/Form.js
```diff
@@ -101,6 +102,19 @@ const DaemonLoadingBody = ({
...
           }
         </div>
+        {daemonWarning && getCurrentBlockCount <= 0 ?
+          <Aux>
+            <div className="get-started-last-log-lines">
+              <div className="last-dcrwallet-log-line">{daemonWarning}</div>
+            </div>
+            <div className="advanced-page-form">
+              <div className="advanced-daemon-row">
+                <T id="getStarted.longWaitWarning" m="You are currently upgrading to dcrd version 1.4.0.  Typically, this one-time reindexing take 30-45mins on an average machine."/>
```
matheusd: I'd remove the reference to a specific version, given that this might happen in newer versions as well:
```suggestion
                <T id="getStarted.longWaitWarning" m="You are currently upgrading to a new dcrd version.  Typically, this one-time reindexing takes 30-45 minutes on an average machine."/>
```

PR #1895, opened 2018-12-12
[sidebar] Switch to lottie animations

matheusd:
This switches the sidebar icons/animations to using lottie instead of using gifs.

The main advantages of lottie are:

- Better behavior when interrupting/resuming animations
- Scalable (due to being svg)

This uses lottie_light instead of the default lottie component, due to the `eval()` present in the full blown lottie renderer.

I don't think this needs to be merged for 1.4.0, given we'll probably need to work out some kinks and we're close to building the rc.

Tested development and packaged versions on linux and windows.


matheusd:
I'm somewhat concerned about the performance pernalty of lottie animations. 

Given that they're javascript animations and that js is not multithreaded, whenever we get a flurry of grpc activity (eg: incoming block, missing cfilters in spv mode, etc), the animations seem to slightly hang. 

Also, this is a small profiling I've done when holding the mouse over the "help" menu link (causing the animation to loop).

Gif (current master) Version:
image

Lottie version:
image

I can provide the profiling json data (20MB+ for each session) on request.

I'll also prepare binaries for the designers to test out this version.

matheusd:
Builds are up: https://github.com/matheusd/decred-weekly-builds/releases/tag/v20181212125445

alexlyp:
Going ahead and closing this PR due to it's age and I believe we've moved on from lottie being a possibility for the time being.  If they update and performance improves maybe we can revisit.

PR #1904, opened 2018-12-14
[multi] Improve native package compilation and docs

matheusd:
Main goals for this commit are:

- Enable compilation of required modules in non-standard architectures (in particular: raspberry pis)
- Allow ignoring compilation of non-required modules (grpc-tools & win32ipc in !windows platforms)
- Improve documentation for native module building

This is basically the result of trying to run the latest version on raspberry pis, while testing and maintaining compatibility to the other platforms.

With SPV suport in the wallet, it's now become possible (if not yet terribly practical) to run decrediton on the latest Raspberry PIs. The main challenge was getting the native modules to compile on the arm7l arch using electron runtime (while ensuring other platforms weren't broken in the process).

In the end, the following changes were required:

- grpc-tools was moved from `dependencies` to `optionalDepencies`. It's pretty much hopelessly broken in electron@arm, so it's install is simply ignored. This package isn't terribly needed anyway, given that the only point of this is to install protoc, and any dev that needs to run protoc to update the grpc in decrediton is likely to have it installed in the system anyway. I'd even vote to drop it completely as a dependency.
- Added a conditional definition in the win32ipc module such that on non-windows platforms, it will not use the source code files referencing `<windows.h>` and therefore won't fail to compile (in effect, generating an empty module).
- Hard coded an update to nan@2.11.1. I had to do this manually, by deleting the lines in yarn.lock and forcing recreation of the lock file. No amount of juggling requirements, updating, removing and re-adding direct and transitive dependencies generated the correct version requirement. This is the important bit in allowing compilation of grpc in platforms other than {windows,linux,macOS}. The current lockfile specifies version 2.8.0 for nan, which is incompatible with newer versions of the grpc library.
- Added the `rebuild-natives` script, to be called on non-standard platforms so that the native modules are rebuilt targeting electron. This is was an added difficulty for trying to run on PI, because the module would usually target the node runtime instead electron, also causing all sorts of problems. This script causes the native modules in the app dir (which is the package the actually requires grpc - trying to rebuild the modules from the root dir doesn't work) to be correctly rebuilt in the current platform.
- Rewritten large parts of the readme to improve the instructions for setting up a dev environment and calling out various tricks and pitfalls in specific platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,39 +67,68 @@ To pass additional arguments to dcrwallet (such as to increase the logging level
 decrediton --extrawalletargs='-d=debug'
 ```
 
-## Developing
+## Development Instructions and Tips
 
-Due to potential compatibility issues, for now, all work should be
-done with electron 2.0.0.
+### Node and Electron Versions
 
-You need to install dcrd, dcrwallet and dcrctl.
+You need to use compatible node and electron versions to build native modules
+and in general to ensure the code you're working on will behave correctly in
+production.
+
+The current recommended versions for the main tools are:
+
+  - Node: 10.2+
+  - Npm: 6.4+
+  - Yarn: 1.12+
+  - Electron: 3.0.6
+
+To ease node version management, install all top-level tools (node/npm/yarn) using nvm ([linux/mac](https://github.com/creationix/nvm), [windows](https://github.com/creationix/nvm)).
+
+
+### Decred binaries
+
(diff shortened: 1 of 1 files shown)


alexlyp:
Right so really the only reason I added grpc-tools was to have some consistency when rebuilding grpc bindings.  I found that if it was left to the dev, the version would shift and cause huge amounts of change in the built files.

PR #1935, opened 2018-12-19
[export] Fix export transaction's title

### LINKED ISSUE #1851: [export] Some tabs titles diverge from the what is explained
On the export tab, the explanations on daily balances and balances sheet diverge from the title that actually are exported.

vctt94:
fixes #1851. 

I followed the explanation on export tab to the sheet's titles, but we can add more information if needed.


vctt94:
Do you also think we should keep the information about stake on the daily balance stats?

matheusd:
Yeah, we've already calculated that, so just leave it...

vctt94:
Now I am adding the explanation on the export page. I believe this way looks better

app/actions/StatisticsActions.js
```diff
@@ -633,18 +629,23 @@ export const dailyBalancesStats = (opts) => {
...
-      { name: "revoked", type: VALUE_TYPE_ATOMAMOUNT },
-      { name: "ticket", type: VALUE_TYPE_ATOMAMOUNT },
-      { name: "stakeRewards", type: VALUE_TYPE_ATOMAMOUNT },
-      { name: "stakeFees", type: VALUE_TYPE_ATOMAMOUNT },
-      { name: "totalStake", type: VALUE_TYPE_ATOMAMOUNT },
-    ];
-    startFunction(opts);
+  const aggStartFunction = () => {
+    startFunction({
+      series : [
```
matheusd: You removed the [ ...opt.series ] so now it's missing the immature and immatureNonWallet columns

matheusd:
Sorry for nitpicking, but seems that now the stakeRewards, stakeFees and totalStake isn't being exported anymore.

You basically need to export from the same wallet on master and your branch and get the exact same result. And the changes in your diff need to be just on the explanations (the exportTab/messages file).

vctt94:
In balance or daily balances? Here when I export it works just fine. They are exported

matheusd:
daily balances

matheusd:
Yeah, changing the order is an issue because anyone relying on the CSV to perform further processing will have to adjust their own scripts to match the new order, and they may only find out the order changed after serious reporting errors have occurred.

Example:

If I have a spreadsheet that uses as input the generated CSV and uses a given column to calculate my outstanding tax liability, if we change the order of columns the spreadsheet will break and may either completely fail (best case scenario) or generate an incorrect payment amount (over or under paying my taxes) which can have serious legal issues.

vctt94:
I see what you said and I agree with that, but as I said, right now on master the stakeRewards, stakeFees and totalStake are duplicated. I just removed the ones in the end and the first one are in the same column.

PR #1937, opened 2018-12-19
[launcher] Show settings on launcher

alexlyp:
This will show settings on the launcher all the time now.  It will hide any wallet related settings if there is no wallet yet loaded.


matheusd:
Seems that accessing the settings after the wallet has loaded isn't showing the wallet-related options (gap limit and displayed units)

vctt94:
I believe it still misses some settings information. `walletReady` stills undefined on general settings and I believe that is the reason

PR #1960, opened 2018-12-26
[mytickets] Allow cancelation when listing tickets

### LINKED ISSUE #1958: [Tickets] Unable to load more tickets
Under Tickets - my tickets, when filters set to all tickets i can scroll down and view older tickets. 

When filter set to voted and live, text showing "Loading more tickets..." with bouncing stakey but nothing gets loaded for very long time.

matheusd:
This allows cancelation of the whole getTickets process mid-way through,
when problems or delays are noticed.

Decrediton's getTickets current scans the whole transaction set for
tickets of the specified filter status, so some combinations may be slow
to fulfill.

close #1958


matheusd:
Updated to deal with the duplicate key error.

matheusd:
Updated to deal with the large-width display issues.

PR #1968, opened 2019-01-02
small translation fixes

Guang168:


alexlyp:
I'll merge this now, but make sure you've made the same changes on transifex to avoid these changes being overwritten later on.

PR #1971, opened 2019-01-02
[myTickets] Fix my tickets card not clickable in some areas

### LINKED ISSUE #1962: [1.4.0-rc] Mytickets card is not all clickable
Despite the whole card show the cursor as a pointer, there are places that clicking on them, does not triggers the action:

image

vctt94:
closes #1962

### DIFF
--- a/app/style/MyTickets.less
+++ b/app/style/MyTickets.less
@@ -32,7 +32,7 @@
 }
 
 .ticket-card.is-expanded {
-  height: 100%;
+  height: 165px;
 }
 
 .ticket-overview-header {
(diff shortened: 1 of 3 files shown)


app/style/MyTickets.less
```diff
@@ -32,7 +32,7 @@
 }
 
 .ticket-card.is-expanded {
-  height: 100%;
+  height: 165px;
```
matheusd: Returning this to 165px is breaking the height on higher-width displays (eg: when the wallet's window width is 1240px).

vctt94:
fixed

Issue #1646, opened 2018-09-13
[config] Allow CLI config option setting


alexlyp:
As pointed out by @davecgh, currently there is no easy way to access the config or much less know what options are available for setting. 

We should at the very least be providing a document that let's users know what settings are available in `config.json`

Ideally, we should let users specify options on the command line, using dcrwallet or dcrd as an example.  So something like `./decrediton --daemon_start_advanced`  would start up decrediton while in remote daemon mode.    When we looked at this previously, I believe there was some issue with option consumption due to Electron.  

Here is a basic list of options I'd like to provide for CLI setting:
1) Confirm `--testnet` setting works.
2) `--advanced` would start it in advanced mode (short hand for `daemon_start_advanced`
3) Possibly individual remote credentials (`--rpcuser=USER`, `--rpcpass=PASS` etc)
4) `--spv` would start in SPV Mode (and `--spv_connect` to specify direct peer)

PR #1889, opened 2018-12-10
Add separate switch for Trezor toggle

szpasztor:
Resolves https://github.com/decred/decrediton/issues/1882

PR #1975, opened 2019-01-07
Allow CLI config option setting

szpasztor:
Resolves https://github.com/decred/decrediton/issues/1646

Subtasks
- [x] audit old options
- [x] test and add aliases
- [x] add testnet/mainnet options
- [x] add SPV option
- [x] SPV connect
- [x] individual remote credentials (--rpcuser=USER, --rpcpass=PASS etc)
- [x] indicate CLI options used on settings screen
- [x] update CLI help message
- [x] PR to update end user docs on [SPV CLI usage](https://docs.decred.org/wallets/spv/#what-is-spv). See: https://github.com/decred/dcrdocs/pull/780
- [x] update readme (advanced daemon mode & how to run decrediton.app from CLI on macOS)


szpasztor:
As for verifying that spvConnect works: I tried launching a local daemon with `--rpclisten=127.0.0.1:19109` and passing '127.0.0.1:19109' as the spv connect peer and some more test runs but not sure what should happen. Also, can both SPV and advanced daemon mode be on at the same time?

app/main.development.js
```diff
@@ -13,13 +13,15 @@ import { cleanShutdown, GetDcrdPID, GetDcrwPID } from "./main_dev/launch";
 import { getAvailableWallets, startDaemon, createWallet, removeWallet, stopDaemon, stopWallet, startWallet, checkDaemon, deleteDaemon, setWatchingOnlyWallet, getWatchingOnlyWallet, getDaemonInfo } from "./main_dev/ipc";
 import { initTemplate, getVersionWin, setGrpcVersions, getGrpcVersions, inputMenu, selectionMenu } from "./main_dev/templates";
 import { readFileBackward } from "./helpers/byteActions";
+import { isString } from "lodash";
```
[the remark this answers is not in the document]
szpasztor: That's how it's used in launch.js too, but I'll refactor both then.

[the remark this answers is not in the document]
matheusd: Good catch!

szpasztor:
Thanks for the feedback. I agree with your UX suggestion, it could be done by adding extra boolean flags to fields that can be set from both the global config and the CLI. 

I'm not sure why changing the lower level calls instead of the settings would be better. 
- As far as I see it, the settings represent the configuration the app instance is currently running with, which happens to be same as the global conf now. The former is evidenced by many selectors (eg. spvMode/isSpv) which are used throughout the app in components, forms etc - many of these would break and inconsistencies would appear if the settings just copied the config as-is. 
- Also, it'd be confusing UX that eg. the app GUI shows SPV as set to disabled (with the dropdown disabled) but it's actually turned on. Applies to all other options too. 
- However, if cli options are stored in currentSettings, it's important not to override the config's value if the user changes a non-cli-settable option. Eg. global config has mainnet set but they start decrediton with `--testnet` and then change the proxy type on the settings screen - in this case, `network` shouldn't be overriden when saving. This can be dealt with with the boolean flags so should be fine.

matheusd:
Ok, point taken. 

I guess the ideal way forward then is to keep using `currentSettings`, but add the indication and UX that a CLI-provided setting is unchangeable.

szpasztor:
Yes, will add. 


> I think spvConnect would be better being spvconnect to matches dcrwallet.

@vctt94 Sounds reasonable but then all options should follow the same convention. The new options would be `extrawalletargs`, `custombinpath` and `spvconnect`. I'm making this change unless there are objections.

app/actions/DaemonActions.js
```diff
@@ -314,27 +314,50 @@ export const startWallet = (selectedWallet) => (dispatch, getState) => {
 };
 
 export const prepStartDaemon = () => (dispatch, getState) => {
+  console.log("prepstartdaemon");
   const { daemon: { daemonAdvanced, openForm, walletName } } = getState();
+  console.log(daemonAdvanced);
+  const cliOptions = ipcRenderer.sendSync("get-cli-options");
```
vctt94: Instead of having many `sendSync("get-cli-options")`, you can do it only once and set the cli-options into a selector.

szpasztor: Yeah, was going to do that during final refactoring.

szpasztor:
@matheusd ready for review. 

Thorough testing is required as the PR changes the startup logic. 

@vctt94 I briefly explored using Redux for storing CLI options instead of invoking `get-cli-options` (see `8c0412a`). What I found is that
- Only one occurence in components would benefit from the refactoring (GetStartedPage)
- Two occurences in actions would need to have selectors available to get the object (DaemonActions, which will probably be refactored soon anyway) 
- `app/index.js` would need to be made compatible with dispatching actions. 
I'm not sure if these are among the intended uses of Redux and if refactoring 3 calls to "get-cli-options" is worth the extra workarounds. I would leave it as-is unless there are objections.

app/main.development.js
```diff
@@ -55,12 +59,66 @@ if (argv.version) {
...
+} else if (argv.mainnet) {
+  cliOptions.network = "mainnet";
+}
+if (argv.advanced) {
+  cliOptions.daemonStartAdvanced = true;
+}
+if (argv.spv) {
+  cliOptions.spvMode = true;
+}
+if (argv.spvconnect !== undefined && isString(argv.spvconnect)) {
```
matheusd: Isn't just `isString()` sufficient for these (given that `isString(undefined) === false`)?

app/main.development.js
```diff
@@ -55,12 +59,66 @@ if (argv.version) {
...
+  const parts = argv.rpcconnect.split(":");
+  // Allowed formats: "127.0.0.1" or "127.0.0.1:19109"
+  if (parts.length !== 1 && parts.length !== 2) {
+    logger.log("error", RPCCONNECT_INVALID_FORMAT);
+    app.quit();
+  }
+  cliOptions.rpcHost = parts[0];
+  cliOptions.rpcPort = parts[1];
+}
+cliOptions.rpcPresent = rpcOptionsCount > 0 ? true : false;
```
matheusd: Maybe this should be `rpcOptionsCount >= 4` instead? So that rpcPresent will only be true if _all_ options are set? Would that work?

app/i18n/translations/original.json
```diff
@@ -555,6 +555,7 @@
   "sendToAccount.to": "To",
   "settings.SPV": "SPV",
   "settings.SPVConnect": "SPV Connect",
+  "settings.alreadySetFromCli": "This was set as a command-line option when launching decrediton.",
```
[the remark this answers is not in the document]
szpasztor: OK, so I should only specify the default message on components and the ID, and can assume they'll be added to the translation file and possibly translated later?

Also, similar changes [here](https://github.com/decred/decrediton/pull/1889/files#diff-bc607d3a9d840a15d37e84c46d1a8b27R147) have already been merged, perhaps by accident then?

app/main.development.js
szpasztor: True, the old code had both checks and I just used that.

app/main.development.js
szpasztor: Due to the validation of RPC options above, this should never execute with `rpcOptionsCount` having a value other than 0 or 4, so in terms of outside behavior the two checks should be identical. I'm fine with it though as it might be more self-explanatory for future devs reading the code. Better yet, `==` could be even better for signaling intent.

szpasztor:
@matheusd Use the following format (the same as when specifying `custombinpath` or `extrawalletargs`): 

```
decrediton -d --advanced --rpcuser=USER --rpcpass=PASSWORD --rpccert="./path/to/rpc.cert" --rpcconnect=127.0.0.1:19109
```

Can you please send the lint error you're referring to? Running `yarn lint-fix` locally shows no issues for me. 

I added the remaining changes. Once the code is final, I'll test every option again to make sure nothing broke silently.

app/i18n/translations/original.json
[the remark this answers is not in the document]
matheusd: Yeah, I hadn't noticed the change in the translation file.

matheusd:
After the latest changes, it's now failing the tests from `yarn test` (look at the results of the travis run).

Using the CLI options as you sent didn't trigger the error.

szpasztor:
Yes, just noticed. Strange because the error is shows (undefined ipcRenderer [here](https://github.com/decred/decrediton/pull/1975/files#diff-2d49e62597fd788c1ed99675bf323ddcR22)) is not an actual error when running the app, the RPC options seem to work fine. Might be a CI config issue, I'm debugging it.

szpasztor:
> When trying to use rpc flags in advanced, with a missing port, I get the following error

Good catch. I think the app could resolve this error instead of raising it. I added a default port for now and opened an issue for handling these cases in a better way. 

> Also I do not see a reason for not letting the app be in advanced and spv mode, despite it jumps the daemon connection.

Not entirely sure what you mean by 'jumps the daemon connection'. In any case, I think it's better to inform the user about this early on than ignore the option. For me it's more intuitive that if I specify `advanced`, the app will either run in advanced mode or show a message if that's not possible for some reason.

szpasztor:
All good on macOS. Would be optimal to test on other platforms too, but ready from my side.

matheusd:
I tested on linux.

matheusd:
It will be, don't worry. :)

There are a couple of parallel works going on, and a pending release to be done, so don't worry about merging something just yet.

Issue #1761, opened 2018-10-31
[macOS] Cmd+Q does not trigger exit animation


szpasztor:
- When quitting the app via the native red exit button, the app displays the exit animation as shown below. 
- However, when quitting by pressing Cmd+Q, the app successfully exits after a short delay, but no animation is shown
![image](https://user-images.githubusercontent.com/8398290/47813713-26064c00-dd4c-11e8-9566-3be3a87a8968.png)

szpasztor:
Taking this.

PR #1979, opened 2019-01-12
Properly shutdown on macOS Cmd-Q

szpasztor:
Resolves https://github.com/decred/decrediton/issues/1761


app/main_dev/templates.js
```diff
@@ -1,5 +1,5 @@
 import { app } from "electron";
```
matheusd: This is now unused and needs to be removed to fix the lint error.

PR #1981, opened 2019-01-13
Remove extra setSeconds call

szpasztor:
It seems like the first setSeconds call has no use, because it's called again with `-1` a bit later.

### DIFF
--- a/app/helpers/dateFormat.js
+++ b/app/helpers/dateFormat.js
@@ -29,10 +29,9 @@ export function dateToUTC(timestamp) {
 export function endOfDay(dt) {
   const res = new Date(dt);
   res.setMilliseconds(0);
-  res.setSeconds(0);
+  res.setSeconds(-1);
   res.setMinutes(0);
   res.setHours(0);
-  res.setSeconds(-1);
   res.setDate(res.getDate()+1);
   return res;
 }
--- a/app/helpers/dateFormat.js
+++ b/app/helpers/dateFormat.js
@@ -29,9 +29,10 @@ export function dateToUTC(timestamp) {
 export function endOfDay(dt) {
   const res = new Date(dt);
   res.setMilliseconds(0);
-  res.setSeconds(-1);
+  res.setSeconds(0);
   res.setMinutes(0);
   res.setHours(0);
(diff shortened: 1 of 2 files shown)


matheusd:
Try running your version and the original version on a javascript console with `endOfDay(new Date())`.

Are the results the same for you?

szpasztor:
No, they're not. So if this is a necessary workaround, it'd make sense to add some explanation at least to avoid confusion?

matheusd:
Sure.

szpasztor:
Do you remember why it works/any links perhaps? 

Alternatively, here's an alternative [solution](https://stackoverflow.com/questions/8636617/how-to-get-start-and-end-of-day-in-javascript/8636674#8636674) which seems to be the accepted solution and behave the same (besides using last microsecond vs second, but that can be changed if it's a problem):
```js
function endOfDay2(dt) {
  dt.setHours(23,59,59,999);
  return dt;
};
```
Running on Chrome 73.0.3671.0: 
image

matheusd:
It works because setSeconds() works even for negative or >59 seconds: in that case it subtracts/adds the "excess" into the rest of the date.

Replacing for a setHours call is fine as well.

szpasztor:
Added the setHours version.

app/helpers/dateFormat.js
```diff
@@ -24,16 +24,11 @@ export function dateToUTC(timestamp) {
...
-// of the day stored in the given date.
-export function endOfDay(dt) {
-  const res = new Date(dt);
-  res.setMilliseconds(0);
-  res.setSeconds(-1);
-  res.setMinutes(0);
-  res.setHours(0);
-  res.setDate(res.getDate()+1);
-  return res;
+// Returns a new date pointing to the last microsecond of the day of the given date
```
matheusd: Small nitpick: it's pointing to the last *milli*second instead of the last *micro*second.

szpasztor: Good catch

szpasztor: Fixed

szpasztor:
Added tests, ready for review.

Issue #1983, opened 2019-01-15
[config] Always use appropriate default RPC port if it's unspecified


szpasztor:
In the upcoming CLI options [PR](https://github.com/decred/decrediton/pull/1975) and in some cases (eg. [here](https://github.com/decred/decrediton/blob/f07d7d1f1af6e38748299e6a8dc124cc18bbc016/app/actions/WalletLoaderActions.js#L226) and [here](https://github.com/decred/decrediton/blob/f07d7d1f1af6e38748299e6a8dc124cc18bbc016/app/actions/WalletLoaderActions.js#L231)) the mainnet port is used as the default value. It would be better if decrediton always set the default depending on the current network (ie. `9109 ` for mainnet and `19109` for testnet).

PR #1988, opened 2019-01-16
Host & Port not hardcoded anymore

laszlolm:
Solving #1983


alexlyp:
Sort of bouncing off of @vctt94's comment, but yah I believe these changes are essentially a noop.  But it does make everything more consistent so I'm willing to change it.

Overall as @vctt94 mentioned the function should probably be refactored at some point.

PR #1989, opened 2019-01-16
WIP Exit alert ticketbuyer running

### LINKED ISSUE #1923: Alert when closing wallet/shutting down that ticket buyer is running
While we alert that the wallet needs to remain open when turning on the ticket buyer, it's been suggested in chat that we should also alert when closing the wallet or shutting down the app that the ticket buyer will be closed and no tickets will be further purchased.

We could also possible add warnings when going into/out of suspend mode or on app shutdown.

https://github.com/electron/electron/blob/master/docs/api/power-monitor.md

szpasztor:
Resolves #1923 

Tasks
- [x] alert on app shutdown (clicking red close button in title bar)
- [x] alert on app shutdown (Cmd+Q)
- [x] alert on wallet close (in settings menu)
- [ ] ~~prevent & alert when entering suspend mode~~
- [x] prevent & alert when on system shutdown
- [x] log warning


app/containers/App.js
```diff
@@ -51,23 +56,13 @@ class App extends React.Component {
         this.props.showAboutModalMacOS();
       }
     });
-  }
-
-  beforeWindowUnload(event) {
```
szpasztor: Removed the whole block as it seemed to have no use. Also, according to [this thread](https://github.com/electron/electron/issues/2301), it's not recommended to override this event.

No full review is needed for the PR yet.

szpasztor:
Seems like there's an issue with powerSaveBlocker on macOS, I opened an [bug report](https://github.com/electron/electron/issues/16529) about it.

szpasztor:
Ready for review.

Issue #1784, opened 2018-11-08
Auto dark theme


vanjan:
This is a feature request. 

The UI should have another, dark, theme. By default, the UI should change theme in accordance to the OS, eg macOS Mojave dark mode.

matheusd:
It's in the works. Don't know if it's possible to detect the current OS theme via electron (I'm going to guess: no) but we'll try and research that.

vanjan:
This looks promising: https://stackoverflow.com/questions/25207077/how-to-detect-if-os-x-is-in-dark-mode#25214873

laszlolm:
I'm gonna start on this issue since Electron now supports getting dark mode natively:
```
const {systemPreferences} = require('electron')
console.log(systemPreferences.isDarkMode())
```

PR #1993, opened 2019-01-19
Change dark mode automatically based on OS preferences

laszlolm:
Solving #1784

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -108,7 +108,27 @@ const installExtensions = async () => {
   }
 };
 
-const { ipcMain } = require("electron");
+const { ipcMain, systemPreferences } = require("electron");
+
+const changeDarkMode = (newAppearance) => {
+  if (newAppearance === "light" || newAppearance === "dark") {
+    globalCfg.set("theme", newAppearance === "light" ? "theme-light" : "theme-dark");
+  }
+};
+
+// Set dark / light mode on app launch
+// isDarkMode() return true or false (true when dark, false when light) so we need to convert it to 'light' or 'dark'
+if (process.platform === "darwin") {
+  changeDarkMode(systemPreferences.isDarkMode() ? "dark" : "light");
+
+  // listen to dark / light mode theme changes
+  systemPreferences.subscribeNotification(
+    "AppleInterfaceThemeChangedNotification",
+    () => {
+      mainWindow.webContents.send("toggle-theme", systemPreferences.isDarkMode() === true ? "theme-dark" : "theme-light");
+    }
+  );
+}
 
 ipcMain.on("reload-allowed-external-request", (event) => {
   reloadAllowedExternalRequests();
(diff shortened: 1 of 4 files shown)


app/main.development.js
```diff
@@ -108,7 +108,27 @@ const installExtensions = async () => {
...
+const changeDarkMode = (newAppearance) => {
+  if (newAppearance === "light" || newAppearance === "dark") {
+    globalCfg.set("theme", newAppearance === "light" ? "theme-light" : "theme-dark");
+  }
+};
+
+// Set dark / light mode on app launch
+// isDarkMode() return true or false (true when dark, false when light) so we need to convert it to 'light' or 'dark'
+if (process.platform === "darwin") {
+  changeDarkMode(systemPreferences.isDarkMode() ? "dark" : "light");
```
matheusd: Don't have a mac to test, but you should probably only do this if the theme is undefined, otherwise this will get reset every time the app is started, which might be weird if the user actually wants to run a theme other than the one for the OS.

app/main.development.js
```diff
@@ -108,7 +108,27 @@ const installExtensions = async () => {
   }
 };
 
-const { ipcMain } = require("electron");
+const { ipcMain, systemPreferences } = require("electron");
+
+const changeDarkMode = (newAppearance) => {
```
matheusd: Small naming nit: this should be `changeTheme`, `selectTheme`, `selectAppearance` or something like that, given that it's not changing the dark mode in any way, but rather selecting whether to use dark or light mode.

You could also just pass the `theme-*` value directly instead of performing another test.

app/main.development.js
laszlolm: Good catch, will fix this

app/main.development.js
laszlolm: I will remove this part since it's not used anywhere else and integrate the key line into the code handling the startup mode below

app/main.development.js
matheusd: What I meant was, this will get executed at *every* startup instead of only when the theme in decrediton is undefined (something like, `globalCfg.get("theme") === undefined`).

Can you verify whether the following happens:
- Use "systemPreferences.isDarkMode() === false" (ie, OS is configured for light mode)
- Start Decrediton
- Switch theme to dark
- Close Decrediton
- Start up again; theme reverts to light

This is what I'm questioning if it happens.

laszlolm: It happens, but the theme is never undefined on startup from the first time the user starts the app

matheusd: Hmm, probably best to move this code to config initialization then. See if moving the detection and selection of the theme here works:

https://github.com/decred/decrediton/blob/master/app/config.js#L80

This should only execute when the `theme` config is not defined on the file, which will then detect and change the theme during startup only on the first run.

vanjan:
Ideally this should work similarly to how other apps, such as Safari or FireFox, handle OS theme, ie there are 2 theme settings:

1. default: changes according to OS theme
2. theme explicitly selected by user: doesn't change with OS theme

Given 1, app theme should change whenever OS theme changes, ie not only on startup. Consider what happens when OS theme changes while the app is running -- it should update.

vctt94:
needs rebase

alexlyp:
Can you rebase this please?  Otherwise I'll be closing as it's been stagnant for some time.

alexlyp:
Closing based on stagnation.  Feel free to open another one with rebased code.

PR #2017, opened 2019-02-06
[launcher] Add pipetx to wallet on win32 platform

### LINKED ISSUE #1712: Make pipetx work on windows
WIP:
https://github.com/matheusd/decrediton/tree/pipetx-win

The basic requirement is that the module needs to use libuv's [uv_open_osfhandle](http://docs.libuv.org/en/v1.x/fs.html#c.uv_open_osfhandle). However that has only been [added to libuv 1.23.0](https://github.com/libuv/libuv/pull/1930), which has only been imported in [node 10.10](https://github.com/nodejs/node/pull/22365).

However, electron 3.x uses node 10.2 (and a [custom version](https://github.com/electron/node#branching-strategy) at that) so it's not easily upgradable.

We need to wait for electron 4 to to be released, which will use node 10.11, to then come back, update and recheck this work.

matheusd:
~**Requires #2009**~

This adds the pipetx flag to wallet initialization when running on the windows platform, so that the app can perform correct fetching of the grpc port during wallet startup and allows us to get rid of port detection via logs, which is brittle.

This requires the app to be compiled with an electron version 4 or higher, due to the correct function only being available on libuv 1.23+.

When testing this you might need to remove the entire `node_modules` and `app/node_modules` for the correct version of the module to be compiled. The win32ipc module does **not** work when using `yarn build && yarn start`, only with either `yarn dev` or `yarn package`.

Close #1712.

### DIFF
--- a/app/main_dev/launch.js
+++ b/app/main_dev/launch.js
@@ -6,6 +6,9 @@ import parseArgs from "minimist";
 import { OPTIONS } from "./constants";
 import os from "os";
 import fs from "fs-extra";
+import util from "util";
+import { spawn } from "child_process";
+import isRunning from "is-running";
 import stringArgv from "string-argv";
 import { concat, isString } from "../fp";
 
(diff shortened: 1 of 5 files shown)


matheusd:
Rebased and re-tested after #2009 merge.

Issue #2011, opened 2019-02-05
Paste is not working on hex seed restore


vctt94:
When recovering a seed, if trying to paste a hexadecimal seed, paste does not work.

It should work.

PR #2019, opened 2019-02-06
[Restore Seed] Remove error message

### LINKED ISSUE #1980: The "Restore Existing wallet" prompt Unfriendly
I tested "Restore Existing wallet"under decrediton 1.4.0 rc2.

- When I type the first word, The Decrediton prompt "**Error: 3 INVALID_ARGUMENT: walletseed.DecodeUserInput: invalid encoding: encoding/hex: invalid byte: U+0072 'r'**"

- When I type the second word, The Decrediton prompt "**Error: 3 INVALID_ARGUMENT: walletseed.DecodeUserInput: invalid encoding:: pgpwordlist.DecodeMnemonics: word absurd is not valid at position 1, check for missing words**"

I think those notices will make people misunderstand, Could we optimize it?

vctt94:
fixes #1980

I removed the logic from the `ExistingSeed` form, to keep only `JSX`.
I also did some refactor because we have some repeated code.

When testing, I noticed the hex seed was not working, as described at #2011. I will work on this next.

### DIFF
--- a/app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
+++ b/app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
@@ -63,6 +63,6 @@ const ExistingSeedForm = ({
       </div>
     </Aux>
   );
-}
+};
 
 export default ExistingSeedForm;
--- a/app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
+++ b/app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
@@ -3,132 +3,65 @@ import SeedHexEntry from "./SeedHexEntry";
 import { TextToggle } from "buttons";
 import { FormattedMessage as T } from "react-intl";
 import "style/CreateWalletForm.less";
-import { SEED_LENGTH, SEED_WORDS } from "wallet/seed";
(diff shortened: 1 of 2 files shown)


app/components/views/GetStartedPage/CreateWallet/CreateWalletForm/ExistingSeed/Form.js
```diff
@@ -3,132 +3,66 @@ import SeedHexEntry from "./SeedHexEntry";
...
+                  seedWord={seedWord}
+                  value={{ name: seedWord.word }}
+                  key={seedWord.index}
+                  onPasteFromClipboard={pasteFromClipboard}
+                />);
+            })}
+          </div> :
+          <div className="seedArea hex">
+            <SeedHexEntry
+            // needs implementing
```
matheusd: Is this still applicable?

vctt94: Yeah, that is expected.

Right now on master, the method used to handle hex seed paste is `setSeedWords`, which is a copy of `onChangeSeedWord`. But the problem is pasting is not working on master, I believe because of handlePaste method, so I removed this method and I will work on this next.

But if you prefer, I can add into this PR. The reason I did not add into it, it is only to not have a big diff and fix 1 issue with 1 PR.

matheusd: Ah, ok. I guess it's fine if it's on another PR.

PR #2036, opened 2019-02-16
sync deps

dajohi:


vctt94:
lint fails

alexlyp:
nACK on this.  Typically deps touching babel/webpack and others need to be done a bit more sensitively.

PR #2068, opened 2019-03-04
[i18n] small zh translation fixes

Guang168:
Fixes based on user feedbacks. 
Already updated on Transifex.


matheusd:
Closing since this was updated on transifex. We'll incorporate the changes in a single step once we're near the next release. Changing the .json files directly is discouraged, unless we're very close to a release and can't go through the usual transifex workflow, since these changes will be overriden by whatever is on transifex.

PR #2071, opened 2019-03-11
Add daemon timeout

### LINKED ISSUE #2037: [advanced daemon] Add timeout when starting with wrong parameters
When starting a remote daemon in advanced mode, if the parameters are wrong, decrediton will keep trying to connect without failing.

We should have a timeout after a time trying to connect and throw an error.

How to reproduce: 

- Start decrediton in advanced mode
- Choose advanced daemon 
- Add a wrong parameter, like the port

On this point, decrediton will stuck on the launching daemon page.

vctt94:
closes #2037 

I left the time slow so you guys can test it.

In order to test, just access the remote daemon with wrong parameters and wait till exceed the timeout and decrediton will return to the form page. 

I thought on a 2 minutes timeout, what do you guys think?


matheusd:
Working fine. I think 2 minutes is too long: you're either connecting pretty much right away or not at all (in the case of advanced daemon; regular mode is different, since we expect the daemon the be starting up, which can take a while). I'd say 30 seconds is sufficient.

Also, you should probably add an error notification saying something to the effect of "timeout while waiting for daemon" so that the user isn't just redirected to the daemon config page without any feedback for what happened.

vctt94:
Thanks for your review Matheus.

I agree with you about the time, I believe 30 seconds is fine. 

I added an error notification, but I believe maybe it needs to be clearer as you didn't see it. Maybe adding it below the progress bar would be better? 

image

matheusd:
Yeah, I didn't see the message. Seems that it shows up below the window when using the default window size (1178x790). 

Personally, I'd say it's better to show as a snackbar error message, since it more clearly shows that it's a transient error.

vctt94:
Removed the old error message and added to snackbar, as you suggested and switched time to 30 sec

matheusd:
> Do you think it would be good to still show some error message at the form page?

Fair enough, I think.

vctt94:
Added the daemon timeout error down the loading indicator.

PR #2087, opened 2019-03-31
Fix advanced mode with empty appdata directory not starting

### LINKED ISSUE #1228: Different appdata not working on advanced daemon start
When trying to start daemon with a different appdata directory on advanced start daemon it does not trigger any action.

vctt94:
This PR closes #1228.

It stills a work in progress, because I noticed dcrd logs are not showing when starting the app with a different appdata, and I believe it can be ok to fix on this PR. 

If I notice it is getting too big, I'll open a new one and let you all know.


vctt94:
Ready for review now. 

I tested on regular mode and advanced, as remote and different appdata daemon.

Issue #2082, opened 2019-03-29
[discussion] Add new RPC connection to dcrd


vctt94:
@matheusd Once suggested we could open a new rpc connection to dcrd so we can make some calls directly, instead of using dcrctl, as we do right now. Here it is: https://github.com/decred/decrediton/issues/1753

And discussing on matrix about adding a way to verify anchored transactions by dcrtime at decrediton, with this rpc connection it would be possible to verify it directly by the anchored file hash.

Discussion: https://matrix.to/#/!gltiHJRZiSJTzvjOEu:decred.org/$1553875230499wyjcF:decred.org?via=decred.org&via=matrix.org

So I think it would be good to have this new rpc connection

davecgh:
I'm surprised it doesn't already have that to be honest.  It's far more reliable and efficient than shelling out to a separate utility.

vctt94:
closing as #2088 is already merged

PR #2088, opened 2019-04-03
[POC] Start new rpc connection to dcrd

vctt94:
This is a POC regarding #2082.

It only works on regular mode right now. I'll still play with advanced mode. 

I am starting a new websocket connection. Receiving notifications when a block is connected and decoding its header. 

The code still needs improvements and better modularization but I am pushing this so you guys can see and comment if have some suggestion/objection.

The goal now is to refactor syncDaemon, switching the getBlockCount method to its own call. Also it will possible to refactor getDaemonInfo, to check if the network matches. After that we can see what direction to follow.


vctt94:

I tested it with regular and advanced mode (remote and appdata) and everything seems to be working.

I moved the `decodeConnectedBlockHeader` to service, together with decoderawtransaction but I am not decoding new blocks right now but we can move the `NEWBLOCKCONNECTED` to the socket, also I think would be good refactoring walletstartup, as now, all the daemon and wallet, actions and props are at the same file. 

I could make these changes at this PR, but right now it is already too long. I can make issues for that if we have no objections.

app/main_dev/launch.js
```diff
@@ -359,3 +377,70 @@ export const readExesVersion = (app, grpcVersions) => {
...
+      const dataResults = parsedData.result || {};
+      const blockCount = dataResults.blocks;
+      const syncHeight = dataResults.syncheight;
+      mainWindow.webContents.send("check-daemon-response", { blockCount, syncHeight });
+      break;
+    }
+    }
+  });
+  dcrdSocket.on("close", () => {
+    console.log("DISCONNECTED");
```
matheusd: Same. Ideally this could send a signal to the GUI so that it could alert that the daemon has closed (if the gui is not performing a clean shutdown yet).

app/reducers/snackbar.js
```diff
@@ -448,6 +447,11 @@ export default function snackbar(state = {}, action) {
     message = messages["TRZ_TOGGLEPINPROTECTION_SUCCESS_" + (action.clearProtection ? "DISABLED" : "ENABLED")];
     values = { label: action.deviceLabel };
     break;
+  case CONNECTDAEMON_FAILURE:
+    type = "Error";
+    action.daemonTimeout ? message = messages["DAEMONCONNECTING_TIMEOUT"] : message = action.error;
+    // values = { originalError: String(action.error) };
```
matheusd: Is this supposed to commented?

app/main_dev/launch.js
```diff
@@ -359,3 +377,70 @@ export const readExesVersion = (app, grpcVersions) => {
...
+  const url = `${rpc_host}:${rpc_port}`;
+  if (dcrdSocket && dcrdSocket.readyState === dcrdSocket.OPEN) {
+    return mainWindow.webContents.send("connectRpcDaemon-response", { connected: true });
+  }
+  dcrdSocket = new webSocket(`wss://${url}/ws`, {
+    headers: {
+      "Authorization": "Basic "+Buffer.from(rpc_user+":"+rpc_pass).toString("base64")
+    },
+    cert: cert,
+    ecdhCurve: "secp521r1",
```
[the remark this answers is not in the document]
vctt94: I found it here: https://github.com/decred/dcrd/blob/master/docs/json_rpc_api.mediawiki#82-example-nodejs-code

On the docs say it is only required for higher node versions `Required for node.js v8.11.1 LTS, not for v10.1.0.`, but testing with `v10.15.3` it is needed.

app/main_dev/launch.js
vctt94: after dcrdsocket is closed it is already on the shutdown, but I liked this idea. I will take a deeper look when refactoring startup.

Maybe that can be good for running more then one instance of decrediton

vctt94:
fixed review comments

app/main_dev/launch.js
matheusd: Right. I meant that as a way of signalling error conditions (eg: dcrd crashed somehow).

app/main_dev/launch.js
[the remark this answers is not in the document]
matheusd: dcrd doesn't currently allow changing the curve (unless you manually generated the cert and key files) so I guess it's fine to use a fixed one here.

app/main_dev/launch.js
vctt94: I see, I think that makes sense. We can also get when disconnecting from remotes daemons. I will take this path.

vctt94:
I removed the block connected method, as we are using `TransactionNotificationsRequest` from https://github.com/decred/dcrwallet/blob/master/rpc/documentation/api.md which I believe makes most sense, because decrediton is towards a spv default mode and this request already contains wallets transactions which will be necessarily anyway, even after getting a new block from dcrd.

PR #2099, opened 2019-04-22
[export] fix wrong total on daily balance export

### LINKED ISSUE #2097: [Export] BUG: Inconsistent balances in Daily Balance CSV
#### Issue Description
I am experiencing two issues with the daily balance CSV export:
1. The daily balances exported from v1.4.0 do not match a previous export I made from v1.3.x. For example, the v1.3.x export may report 102.321 DCR on 12/31/2018 but a subsequent export from v1.4.0 reports 100.876 DCR on 12/31/2018. I know the second balance is wrong because of the second point below.

2. The sum of the `spendable`, `immature`, and `locked` columns in the export do not equal the `total` column. For certain dates the discrepancy can be > 100 DCR.

#### Expected result
For days that have closed, the daily balance in the CSV should not change between exports. The sum of the `spendable`, `immature`, and `locked` columns should equal the `total` column.

#### Versions Affected
Experiencing this issue with Version 1.4.0. I was able to export a correct CSV from Version 1.3.2

#### Steps to reproduce
1. Be running decrediton with staking turned on
2. Export the Daily Balances CSV file from v1.4.0
    * This is sufficient to surface issue 2
3. Export the Daily Balances CSV from v1.3.x and compare to v1.4.0 export
    * Necessary to surface issue 1

vctt94:
fixes #2097

### DIFF
--- a/app/actions/StatisticsActions.js
+++ b/app/actions/StatisticsActions.js
@@ -552,8 +552,8 @@ const txBalancesDelta = async (tx, maturingTxs, liveTickets, walletService, chai
 };
 
 // account for this delta in the balances and call the progress function
-const addDelta = (delta, progressFunction, currentBalance, tsDate) => {
-  progressFunction(tsDate(delta.timestamp), currentBalance);
+const addDelta = (delta, progressFunction, currentBalance, tsDate, backwards) => {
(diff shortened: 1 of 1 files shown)


vctt94:
The other problems with the balance not adding on the beginning, when not backwards. The values are fixed now

Issue #2058, opened 2019-02-25
Responsive views: Restore existing wallet + Fill seed


MariaPleshkova:
Responsive views master issue: #1820

**Restore existing wallet**
Previews: https://xd.adobe.com/view/a5e9c70e-178a-4bf7-7a9a-d3fc818e3769-d3d8/
Specs: https://xd.adobe.com/spec/9fc92412-6dcf-4154-7e6a-eb91232f2fb0-e9f8/

When scaling down to the smallest view, white space of the card is changed to more narrow ( 544px to 355px width) and the whole card is made more compact.

**Fill seed**
Previews: https://xd.adobe.com/view/62c83299-1330-452e-74e8-43eed66707f0-14f9/
Specs: https://xd.adobe.com/spec/302f45df-6ea0-4472-5694-9a9c4edaff72-461e/

When scaling to the smallest size, "Confirm Seed Key" is moved to the card's description place (under the title). Card content is repositioned top down.

Also on both mobile layouts full logo is changed to DCR symbol only.

vctt94:
This issue is lefting the trezor setup. Where should I add it? 

How it looks right now: 

![image](https://user-images.githubusercontent.com/15069783/56586078-6507b280-65b5-11e9-8d34-6fb732726e39.png)

vctt94:
on smaller screens
![image](https://user-images.githubusercontent.com/15069783/57321290-6a95ea00-70d7-11e9-90c9-2e28b27c624c.png)

PR #2100, opened 2019-04-23
Responsive restore seed

vctt94:
This PR starts #2058 and I am doing some clean up on the code, as well. 

It is not finished because it is still lefting the trezor option on the responsive mode, but when addressed, I can finish it.


matheusd:
The words/hex toggle seems to get left out of the screen on smaller widths.

vctt94:
fixed.

Also fixed some stuff I left behind on small width

matheusd:
Seems the create wallet button is now broken on the restore page

vctt94:
fixed

Issue #1949, opened 2018-12-22
Responsive views: Transactions/History


linnutee:
Responsive views master issue: https://github.com/decred/decrediton/issues/1820 

Preview: https://xd.adobe.com/view/4aadd355-8691-4b7c-6cde-c8b39f5d9380-3c9c/
Specs: https://xd.adobe.com/spec/b814d6c2-0f4d-4b53-59c1-a71d8f3ff077-ce25/

Tablet largely same again, scaling the width of rows. 

For mobile use same pattern as in overview (with rows containing less amount of information). `Filter by address ` is scaled in width in order to align with the layout better.

Also referring to updates posted in: https://github.com/decred/decrediton/issues/1942

vctt94:
Is this history tab supposed to show only tx not related with tickets?

PR #2106, opened 2019-05-02
[Responsive] Overview

### LINKED ISSUE #1946: Responsive views: Overview
Responsive views master issue: https://github.com/decred/decrediton/issues/1820 

Preview: https://xd.adobe.com/view/f803fd3f-cbb2-412d-7b9c-b3878e8bcb5d-8e32/
Specs: https://xd.adobe.com/spec/3893e71c-63bc-41b8-77f3-6f7aa0ed6e6f-4ea2/

Starting this with the overview. All following issues will be posted in a similar format, that include these three key views. In i've not covered a specific state of a view that needs more insight, please remind me. 

Briefly the overview needs no adapation in 768px size, as the rows can simply scale down in width and the graph can be re-positioned slightly. 

In 375px mobile view however will need to replace the graph selector with a smaller variant of it, where instead of textual labels are icons. The textual graph data is positioned below the graph; and instead of 14 dates, the graph should display only 12. This way the size of it can remain the same. 

Also the navigation in the mobile view should obviously be on the bottom of the viewport and always visible above the content (just mentioning if the example is confusing).
[decrediton - graph selector icons.zip](https://github.com/decred/decrediton/files/2705383/decrediton.-.graph.selector.icons.zip)

vctt94:
This PR fixes #1946 

It also separates the logic from overview and tx history so it becomes easier to make it responsive. 

We still need to work on the sidebar responsiveness, but I opened an issue for that, as this PR is already getting too big.

Also we need to improve the chart responsiveness, but I am not sure if we are going to still use this package, as @alexlyp  once said he was not liking it very much, so I believe it need some discussion


app/components/views/HomePage/Page.js
```diff
@@ -11,6 +11,46 @@ import TransactionsTab from "./tabs/Transactions";
...
+  <T id="home.tab.balance" m="Balance" />,
+  <T id="home.tab.tickets" m="Tickets" />,
+  <T id="home.tab.transactions" m="Transactions" />,
+];
+
+const tabLink = (i) => {
+  const m = [
+    (
+      <Tooltip text={tabMessages[i]}>
+        <span className="overview-tab-balance">&amp;nbsp;</span>
```
matheusd: Do you really need these `&amp;nbsp;` elements here and in some other spans?

app/style/TxHistory.less
```diff
@@ -31,97 +31,109 @@
 }
 
 .pending-overview-details {
-  background-color: rgba(12, 30, 62, 0.0296);
+  background-color: #edeff1;
```
matheusd: Use a theme var for this. It's fine to define a new one if the correct one does not yet exist.

app/components/views/HomePage/Page.js
vctt94: removed

app/style/TxHistory.less
vctt94: fixed

vctt94:
Yeah, it breaks the layout from the transaction history page, because they were built one above the other, which makes really hard to add responsiveness. 

On this PR I removed mostly the part which was tangled, breaking the history page. 

If you prefer I can fix on this PR, but I was thinking on fix at https://github.com/decred/decrediton/issues/1949 if that is okay to you

PR #2114, opened 2019-05-17
Return array on notification when error is object

vctt94:
I noticed when notification error  are objects, they do not show at the snackbar.

image

This PR transforms the error message to array, so we can show it.

### DIFF
--- a/app/reducers/snackbar.js
+++ b/app/reducers/snackbar.js
@@ -64,6 +64,7 @@ import {
 } from "actions/GovernanceActions";
 
 const WRONG_PASSPHRASE_MSG = "WRONG_PASSPHRASE_MSG";
+const ERROR_IS_OBJECT = "ERROR_IS_OBJECT";
 
 const messages = defineMessages({
   defaultSuccessMessage: {
@@ -289,11 +290,15 @@ const messages = defineMessages({
   TRZ_GETWALLETCREATIONMASTERPUBKEY_FAILED: {
     id: "trezor.getWalletCreationMasterPubKey.failed",
     defaultMessage: "Failed to obtain master extended pubkey from trezor device: {originalError}"
+  },
+  ERROR_IS_OBJECT: {
+    id: "snackbar.errorObject",
+    defaultMessage: "The following error happened: {error}"
   }
 });
 
 export default function snackbar(state = {}, action) {
-  let values, type, message;
+  let values, type, message, error;
 
   const oldMessages = state.messages || [];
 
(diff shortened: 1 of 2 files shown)


app/reducers/snackbar.js
```diff
@@ -289,11 +291,15 @@ const messages = defineMessages({
   TRZ_GETWALLETCREATIONMASTERPUBKEY_FAILED: {
     id: "trezor.getWalletCreationMasterPubKey.failed",
     defaultMessage: "Failed to obtain master extended pubkey from trezor device: {originalError}"
+  },
+  ERROR_IS_OBJECT: {
+    id: "general.error",
```
matheusd: It's probably better to use "snackbar.errorObject" or something  more specific rather than "general.error".

vctt94:
I noticed this error on some politeia's call, so I believe it was a problem at their end. After their last testnet release it did not happen anymore. Here is a print of the error: 
bugpoliteia

I believe this is not something that happens very usual, so I am only printing the object with its `keys: values` , as I believe we do not really handle object errors message so it is only to make debug easier if we see it again.

This is how it looks right now: 
image

matheusd:
Tracking back that specific error, it's triggered by this:
https://github.com/decred/decrediton/blob/cebfc3577fd8a8e57e6e4a6f5250489d9f8d7b08/app/actions/GovernanceActions.js#L80

I just did a quick the test below on a browser and it ends up throwing the object as an error instead of the error string, so that's the original source of the object.

Probably best to change that line to just throw the actual error string (or an an `Error` object).

Otherwise, the PR looks ok.

Test case: 

```js
try { throw "hehehe", {bla: "boo"} } catch (err) { console.log(err) }
```

app/reducers/snackbar.js
```diff
@@ -419,15 +424,21 @@ export default function snackbar(state = {}, action) {
   case TRZ_UPDATEFIRMWARE_FAILED:
   case TRZ_NOCONNECTEDDEVICE:
   case TRZ_GETWALLETCREATIONMASTERPUBKEY_FAILED:
+    type = "Error";
     if (action.error && String(action.error).indexOf("wallet.Unlock: invalid passphrase:: secretkey.DeriveKey") > -1) {
       // intercepting all wrong passphrase errors, independently of which error
       // state was triggered. Not terribly pretty.
       message = messages[WRONG_PASSPHRASE_MSG];
+    } else if (String(action.error).indexOf("[object Object]") > -1) {
```
matheusd: Looking again, it's probably better to test whether this is actually an object (`typeof action.error == "object"`) vs coercing it into a string.

vctt94: I first made like this, the problem with this approach, is that some errors are thrown as objects, so they and on this condition. For example: 

```
Error: 2 UNKNOWN: wallet.PublishUnminedTransactions:: dcrd.jsonrpc.sendrawtransaction: -40: Rejected transaction 4c47d7808aa267f430d2c82a2ff1b78ff1008e2b1cabab0c40bb08c5edec521a: already have transaction 4c47d7808aa267f430d2c82a2ff1b78ff1008e2b1cabab0c40bb08c5edec521a
    at Object.exports.createStatusError (/home/vctt/projects/decred/decrediton/app/node_modules/grpc/src/common.js:91)
    at Object.onReceiveStatus (/home/vctt/projects/decred/decrediton/app/node_modules/grpc/src/client_interceptors.js:1204)
    at InterceptingListener._callNext (/home/vctt/projects/decred/decrediton/app/node_modules/grpc/src/client_interceptors.js:568)
    at InterceptingListener.onReceiveStatus (/home/vctt/projects/decred/decrediton/app/node_modules/grpc/src/client_interceptors.js:618)
    at callback (/home/vctt/projects/decred/decrediton/app/node_modules/grpc/src/client_interceptors.js:845)
```

And when making typeof this error we get an object.

image

vctt94:
I switched to throw an error message. 

It is not clear to me if I should leave the check an error is an object or not.

app/reducers/snackbar.js
matheusd: Got it

matheusd:
It's fine as is. As you mentioned earlier, that's a pretty weird bug to be triggered and means pi returned inconsistent information (or we have a bug somewhere in our code) so a simple error message is fine.

PR #2124, opened 2019-05-21
[responsive] Responsive tx history page

### LINKED ISSUE #1949: Responsive views: Transactions/History

vctt94:
closes #1949 

I am not sure transaction history tab is supposed to show only not stake transactions. As the specs are right now it gives this idea. 

I asked @linnutee at 1949 and after his answer, if that is the case, I will remove the other tx type from it.


app/components/TxHistory/TxHistoryRow/index.js
```diff
@@ -20,6 +21,17 @@ const TxRowByType = { // TODO: use constants instead of string
   "Coinbase": regular("Receive", true),
 };
 
+export const timeMessage = (txTimestamp) => <T
+  id="txHistory.statusSmall.date"
+  defaultMessage="{day} {month} {year} {time}"
```
matheusd: This shouldn't be done like this. Use a single value with a time formatter, such as `{txTimestamp, date, medium}` so that different locales can show the date and time in their preferred format.

If you need a _specific_ custom format for english, then define one in the [main locale file](https://github.com/decred/decrediton/blob/f80e832e55231a8f4cb7b1aa69f6e2faea2709df/app/i18n/locales/index.js#L27) and use _that_ as the formatter (`{txTimestamp, date, day-short-month}`).

If you end up needing to split into finer grained elements of the date, then define a custom format that only outputs that element (such as a new custom date format that shows the 2-digit day but doesn't show any of the other date elements).

vctt94: fixed

PR #2120, opened 2019-05-19
Losslessly compress images, fixes #2118

fulldecent:
Created using https://imageoptim.com/mac

PR #2125, opened 2019-05-22
Losslessly compress vector images

fulldecent:
Created using https://imageoptim.com/mac

Replaces PR #2120


matheusd:
Have you tried optimizing via svgo instead of imageoptim? As I said in the original PR, I'd rather be able to double check that the optimization process for SVGs is deterministic.

Issue #2118, opened 2019-05-19
Images can be losslessly shrunk


fulldecent:
Most/all of the images in the app can be losslessly shrunk. This means they can be encoded better and still have pixel-for-pixel parity with what is currently in this repository.

That saves megabytes off of the final application size.

linnutee:
Noticed it's only addressing the DCR symbol? If the context allows, that too could be as a vector format – svg.

fulldecent:
Here are a list of image optimizers that ImageOptim uses. They are all open source

<img width="775" alt="Screen Shot 2019-05-23 at 15 17 39" src="https://user-images.githubusercontent.com/382183/58280152-feb9af80-7d6d-11e9-9f84-63f4297ede83.png">

PR #2126, opened 2019-05-22
Losslessly compress bitmap images

fulldecent:
Created using https://imageoptim.com/mac

Replaces PR #2120


alexlyp:
nACK. 

Appreciate the PR, but @matheusd and I discussed things a bit and would prefer if you would use a package such as https://www.npmjs.com/package/imagemin and include a script in package.json to have repeatable and verifiable image minification.   Same goes for the PR concering the SVGs as well.

fulldecent:
This PR is not that.

What you are talking about is harder and that is in scope at https://github.com/decred/decrediton/issues/2118

If this PR adds any value it can be merged. Otherwise it can be abandoned. The result of #2118 will be to produce this PR, and to produce it repeatably. But that issue is harder and not available today whereas this PR is available today.

alexlyp:
OK fair enough, closing this PR.

PR #2128, opened 2019-05-27
Fix trying to close remote dcrd

vctt94:
This PR fixes a bug I noticed when switching network at the settings page after commit https://github.com/decred/decrediton/commit/b4125648364222a1cd6968277141a38779f6643e


vctt94:
fixed

Issue #1549, opened 2018-07-25
Move Send button to the right


xaur:
I don't use Decrediton but in #1221 I noticed on the screenshot the Send button is on the left, while the final info is on the right. This is counter-intuitive after most software I used where action buttons are on the right. It would be easier for the eyes to look at the final info block on the left, check everything is correct and move to the right to press Send.

linnutee:
Good notice. This location and left-right/right-left is quite old UX debate which doesn’t really have a consensus. I'm not completely for one or another, but considering this on context basis. 

Imo the case for right alignments is mainly from history. As it’s used a lot in older desktop software UI’s and os level, has formed a habit for many. The reasons behind it are mainly isolating the key interactions (which has been a bigger problem in the past) and building a sense of sequencing (so always moving towards right).

However in Decrediton most of the design fundamentals revolve around a largely left flushed layout, rooting the natural left to right reading logic and Z motion. At the moment trade-offs for changing this would be quite big, as it's not only the send view but should reflect in entire software. Though this can be re-assessed once the mobile/responsiveness gets further, as that brings in another angle. I'll add a note to the dcrdesign repo. 

Points being:
- General left-alignment creates a compact active area for quick navigation
- Changing to right-alignment would add a lot of visual traction (at least in a desktop view) 
- This matter needs to be updated in all of the instances for consistency, so quite big undertaking + as well change to the users 

<img width="747" alt="screen shot 2018-08-09 at 21 23 07" src="https://user-images.githubusercontent.com/17774057/43918007-7641d842-9c1a-11e8-8a1e-131ca072ad7f.png">

xaur:
@linnutee thanks for detailed overview, I see there are more considerations involved.

Indeed right-aligned action buttons originate from older desktop OS and software UI and has formed strong habits for many. But I also see this pattern to this day, and it is not fading at all. Modern desktop software has retained it, and it is expanding to other platforms. In Android action buttons are always to the right. Under the very comment box I'm typing right now the 'Close issue' and 'Comment' buttons are to the right. I wonder how it is on iOS and Mac.

Unlike many deeply rooted habits I'd like to see exterminated (like Windows or inflation), I think this one is quite natural for LTR languages.

[...]

Issue #1956, opened 2018-12-22
Responsive views: Purchase Tickets


linnutee:
Responsive views master issue: https://github.com/decred/decrediton/issues/1820 

Preview: https://xd.adobe.com/view/a00f9e86-6e82-420a-4026-26be4cdc5b7b-d63e/
Specs: https://xd.adobe.com/spec/da214bb6-7903-40e0-7f51-15fa9b643578-ac10/

Few tweaks addressed in: https://github.com/decred/decrediton/issues/1945. Otherwise field titles type sizes reduced in mobile, elements repositioned down-town.

vctt94:
Did the auto purchase ticket move to another place?

linnutee:
Hrmm nope. Just seems to have gone missing on the design. However do keep it as it is.

linnutee:
I'll open another issue for a small alignment fix regarding that on the next week. Seems to have gotten a bit broken over time.

PR #2143, opened 2019-06-27
Visual Design Send View

### LINKED ISSUE #1941: Ixd/Visual Design: Transactions/Send view
Preview: https://xd.adobe.com/view/3fe77764-ecd0-41c7-7e5e-5a6487e983ef-49d6/
Specs: https://xd.adobe.com/spec/ae242c37-0da2-43c1-7e70-df8303e030b7-d815/

Minor updates to the send view in order keep the layout relevant for responsive views and the right-alignment change. 

1) Due to the right alignment and responsive views, have adjusted the card layout to include TX details within it.

2) After some brief user testing, concluded that the actual use of accounts icon reference next to the account field is somewhat obsolete. This also in early Decrediton was intended to display color coded accounts rather than to refer to the accounts view, thus can be removed (elsewhere too).

3) Repositioned `Send to Self` and `Send All Funds` next to the corresponding fields. 

4) Current "Send funds to another account" hover tooltip copy should be changed to "Send funds to self" or something more relevant and clear. When send to self is active, the icon should display an X instead and tooltip display "Cancel sending to funds to self". Also when send to self is active, the plus button should change to disabled state rather than just disappear. 

5) Sending to multiple addresses: took better use of the cards patterns by creating separation between each address/amount set. 

6) Changed the color of the plus button to grey to better fit with the other icons there. 

7) Interaction of `Send Max` should be changed, so it would be disabled when more than one address/amount fields are displayed. (Problem: currently it just closes all the address fields, if a user happens to click on it) 

8) Amount field should have a custom typesetting that displays the dual-sized numbers. Also when `send max` is actived, the field should use a filled pattern (that is not editable) rather than inactive pattern (with the background solid). (Added to work list to do an overhaul of the common components, since these are somewhat inconsistent.)

Icon assets: [decrediton - tx view - icons.zip](https://github.com/decred/decrediton/files/2705327/decrediton.-.tx.view.-.icons.zip)

vctt94:
This PR refactors and update the visual design of transactions send view.

It closes #1941 and #1549


vctt94:
Thanks for reviewing my PR.

overall I like this design better, as well. It is more intuitive. 

I fixed the receive page, the purchase is still broken, but I am working on https://github.com/decred/decrediton/issues/1956 right now, and as the input is a bit different, I will fix on this new PR. 

I am not letting add inputs when sending to another account. This is the proper behavior, we can see by the icons which I had not added, but now I added it. 

and I am clearing the tx after it is published, now.

matheusd:
When trying to send all funds, it seems the input isn't showing the decimal point anymore.

vctt94:
fixed

Issue #2093, opened 2019-04-05
Dark Mode: Settings selection


linnutee:
Once the dark mode is implemented and available in settings, can be referred to as "UI theme" title with a field "Tonality: Light / Dark". E.g.: 

<img width="1179" alt="Screenshot 2019-04-05 at 16 27 34" src="https://user-images.githubusercontent.com/17774057/55631468-c9a6cd00-57c0-11e9-994c-7b10c7afbc76.png">

Ref to master issue: #2089

Issue #2089, opened 2019-04-05
Dark Mode: Master Issue


linnutee:
Created this master issue used for reference to all implementation issues around the dark mode. 

Key drivers and notices: 
- Color scheme: this needs to be separately defined in order not to mix things up. The current (light) scheme's blue and grey have been chosen to only perform for light mode. However for dark mode, the blues/greys making up bulk of the UI structure are a reverse-split with adjusted contrasts and a few tweaks in order for the scheme perform up to accessibility contrast standards as well result in similar but reversed optical look/feel. 

- Listing the only key colors that overlap in both light and dark mode: `#2970FF #2DD8A3 #091440 #69D5F7 #E9F8FE #ED6D47 #FEB8A5 #41BE53 #C4ECCA #FFC84E #FFE4A7`
All dark mode colors used: https://xd.adobe.com/view/63e0674b-87d1-4a81-456c-7d3ea16e1c6a-15d1/

- Iconography: Majority of the icons received a dark mode alternative version. Following a naming convention w/ adding a camelCase Dark at end of icon names. Assuming the presently animated icons will need to be re-rendered also? @kyleFirethought?

- Components: UI components only require change of color, no form or other properties have been changed. 

- /Dcrdesign issue w/ master files https://github.com/decred/dcrdesign/issues/105 

@MariaPleshkova will be posting the individual key views for reference. These should cover all of the common components used in Decrediton. 

Additionally would request for compiling a named list of the current release front-end components (ideally with previews). We can replicate that in the master visual design files to keep better track of those and streamline any future updates. The current visual design component sheet has not seen any updates for a long time and is likely overly synced up with the real implementation. So ideally we'd end up with one for light mode and one for dark mode. Once dark mode is implemented, any updates concerning components will need to reflect in both instances.

MariaPleshkova:
Posted in-app and launcher kew views to decred/decrediton#2122 and decred/decrediton#2123

Issue #2090, opened 2019-04-05
Dark Mode: Navigation


linnutee:
Preview:https://xd.adobe.com/view/c369ffcd-e4a3-41f3-5c0a-1c700070b91e-b558/
Specs: https://xd.adobe.com/spec/16c584cc-34ec-4ad0-757e-6416b05422b5-60f4/
Assets: [decrediton - navigation icons - dark.zip](https://github.com/decred/decrediton/files/3048066/decrediton.-.navigation.icons.-.dark.zip)

Ref to master issue: #2089

PR #2157, opened 2019-07-17
Dark Mode: Settings selection

bgptr:
Solving(#2093)
Ref to master issue: #2089

Dark mode toggle button replaced according to design.


bgptr:
Thanks for the review. 

Do you mean the navigation bar? 
I would like to skin it, but it's in the issue (#2090) and I didn't want to mix the issues.

vctt94:
okay, that's fine. 

There is also another issue I noticed. If you close decrediton it returns to the light theme. You should save this information at the `config.json` file so it can be persisted.

bgptr:
In my implementation, after you choose the theme, you have to click the `Save` button to make the information persistent. 
Of course, it can be saved immediately if you want.

vctt94:
Oh, my bad. I haven't seen the save button, but it is our current behavior at settings. 

If you could reset your commits to cb993b7aa020267ae56bd76dd6d4204b14c53e6f would be good, so we do not have different behaviors at our settings changes. 

Otherwise LGTM.

bgptr:
No problem. I reset the commit.

PR #2164, opened 2019-07-26
fix ticket numeric input showing error at wrong place

vctt94:
After commit 9a5048111fbf21c97be3629cbeeaad02c84bebbb the input error is showing wrongly. 

image

This PR fixes it.


matheusd:
This is working, but the amount field in the send transaction page also seems bugged. Try typing some number then erasing it, the error will overlap with the unit.

vctt94:
Good catch. I fixed it now, thanks.

vctt94:
My bad. Fixed.

Issue #2115, opened 2019-05-18
Update constitution modal with new copy


jholdstock:
Decred constitution has been updated as per [the recent proposal](https://proposals.decred.org/proposals/fd56bb79e0383f40fc2d92f4473634c59f1aa0abda7aabe29079216202c83114)

The old constitution text is included in the file 
https://github.com/decred/decrediton/blob/master/app/components/modals/ConstitutionModalContent.js

The new copy can be found at https://docs.decred.org/governance/decred-constitution/

bgptr:
I'm working on this.

PR #2168, opened 2019-08-07
Update constitution modal with new copy

bgptr:
closes (#2115)


matheusd:
Ideally I think we should switch to using the `Documentation` component (which renders a markdown document in the `i18n` subdir) or a `ReactMarkdown` component directly (if we won't provide translations to the constitution).

bgptr:
Ok, I'm doing the `Documentation` version.

Issue #1497, opened 2018-06-26
Visual Design: Implement Installer


linnutee:
Addressed the visual design for Installer. Use the native Mac OS application icon. Posting the updated Decreditons icon in a follow-up issue. 

Preview: https://xd.adobe.com/view/044a39e3-1256-4779-41e9-9362e821dbb8-727b/
Specs: https://xd.adobe.com/spec/19ff2338-2d69-4c01-5dd3-0b104d4b846e-6fce/

[decreditonInstallerArrow.svg.zip](https://github.com/decred/decrediton/files/2138847/decreditonInstallerArrow.svg.zip)

Jujhar:
Does this still need to be done?

alexlyp:
believe it has been completed.  closing

alexlyp:
Well no, the arrow has not been implemented.   Go ahead @Jujhar  if you think you know how.

Jujhar:
OK

vctt94:
Is this still an issue?

linnutee:
Yep

vctt94:
Closing as #2169 was merged

PR #2169, opened 2019-08-09
Add background to mac osx installer

bgptr:
closes (#1497)


matheusd:
Does the background image really need to be in tiff format?

bgptr:
Probably you are right and it could be e.g. a `background.png` (non-retina) and a `background@2x.png`(retina). It works, but I can not test it on non-retina display. 

In the end, - after a little research - I committed the background image in an optimized multi-resolution TIFF format, which is simply a combination of a standard- and a high-resolution version png. More info about this format: [Package Multiple Versions of Image Resources into One File](https://developer.apple.com/library/mac/#documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Optimizing/Optimizing.html#//apple_ref/doc/uid/TP40012302-CH7-SW13)

Issue #784, opened 2017-10-14
"History" tab not remembering its last Page number/Transaction filter


gravityz3r0:
Not sure if this is intentional built this way, but i would like this feature to be implemented if possible.

In "History" tab, whenever i browse, to example Page 20, and view any transaction detail, it will reset back to Page 1 instead of continuing from the page where i left off. This will require me to scroll all the way back to page 20 all over again.

Likewise, whenever the transaction type filter is applied from the drop down menu (i.e: Vote, Revokes), clicking the back button will send the filter back to default "Regular". I would also request that the drop down menu "remembers" the last applied filter.

jiecut42:
Yup +1. 

I like looking through my vote transactions to see how big they were (which isn't shown without clicking) and also when they'll get 256 confirmations. It's quite a pain because each time you click back you go back to regular.

Yeah so this is one thing I liked about Paymetheus those transaction stats easily. Though some things do need to be sacrificed. Hopefully it can be more streamlined by remembering the last applied filter.

I do like that it shows newest transactions first, and not all the transactions are on one page.

One possible thing though is that maybe vote and ticket transactions could show a bit more information so you don't even need to click on them. There's quite a bit of white space.

rocknet:
Update this to add that as of 1.3.1, the transaction filter stays applied after clicking into the details of a transaction and then clicking back.  The scroll still resets to the top.

PR #2171, opened 2019-08-14
History tab remembers its scroll position

bgptr:
closes (#784)


bgptr:
It is absolutely icky. Maybe the scroll position saving and restoring could be working, but then the scroll jumping would be odd IMHO. Do you have another solution idea?

No specific reason using `useModalForDetails` just didn't want to modify another part of the application. I can revert it in no time if you see it necessary.

matheusd:
Yeah, I think the parametrization can be left out, I don't see us using two different views for transaction details.

bgptr:
Ok, I reverted it.

bgptr:
Thanks! I've fixed it and also removed the fixed positioning.

PR #2175, opened 2019-08-16
Remove deprecated code

vctt94:
After commit https://github.com/decred/decrediton/commit/3d399552856aec933b899ac37113dd5adfb630f6 most of the code being deprecated was removed, but this bits were missed or re-added. 

This PR removes it.

Also, these warnings are showing in decrediton despite all deprecated code being removed. So I believe it is from some dependency we are using:

image

When they get updated we can follow, as theses warnings can be annoying.

### DIFF
--- a/app/components/inputs/Input.js
+++ b/app/components/inputs/Input.js
@@ -19,9 +19,10 @@ class Input extends React.Component{
       this.input.focus();
     }
   }
-  componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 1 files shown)


matheusd:
A quick search seems to imply react-intl and react-motion are two deps still using the old functions. There's a new major version for react-intl which I'll try to see if I can update us to but there isn't a new version for react-motion.

Issue #2056, opened 2019-02-25
Responsive views: Onboarding slides


MariaPleshkova:
Responsive views master issue: #1820

Previews: https://xd.adobe.com/view/2db06669-9b56-4848-40d2-3c6d9aaf0f44-5b81/
Specs: https://xd.adobe.com/spec/c54625d2-1d95-4266-78c2-2dba3ca66946-b377/

When scaling to the smallest view, the image is moved up above the text and stacked in one column with it. Navigation bar with skip/next buttons and pages circles is sticked to the bottom of the screen.

PR #2176, opened 2019-08-16
Responsive views: Onboarding slides

bgptr:
Closes: (#2056)


bgptr:
Thanks for the comment.
I've swapped the buttons and also noticed that the text is in one column format under 768px width so I've fixed that too.

Issue #1998, opened 2019-01-21
Responsive views: Accounts


linnutee:
Responsive views master issue: #1820

Preview: https://xd.adobe.com/view/ddd94038-ecb5-4a9c-7d15-b074a1016d3a-abe8/
Specs: https://xd.adobe.com/spec/78d1896c-9263-4923-5cd8-910a7d914250-0ee0/

No significant adaptations needed for scaling down. Two columns of expanded account data are stacked in a left-to-right, top-down manner. Rename and Hide account icons are in a separate row and centred in the smallest viewport. Spendable should be hidden in the smallest viewport by default and only displayed when expanding the card. Otherwise the smaller card gets quite cluttered.

PR #2178, opened 2019-08-19
[view] Account view

### LINKED ISSUE #1996: Visual Design: Accounts View
Preview: https://xd.adobe.com/view/2b62137b-e877-41f8-4941-2e2a357dc977-5d53/
Specs: https://xd.adobe.com/spec/42a63af8-e1cc-4efb-7d9c-f9b04d13847f-3497/

Smaller updates to improve the experience. 
1) Introduced a subtitle that includes the name of the wallet in use. This helps to clarify the relation between wallets and accounts. 

2) Updated the revealed pubkey address pattern, included copy icon within the blue area. This way it's consistent the similar pattern used in receive view https://github.com/decred/decrediton/issues/1943

3) Would also bring up again the account naming logic. Discussed earlier in https://github.com/decred/decrediton/issues/1520
- Regarding `default` account, simply for better aesthetic could use full capital case Default Account or if possible rename to Primary Account.
- `imported` could be in a disabled state until the user has not started staking. 
- `imported` similarly could be renamed to "Time-locked", "Locked for Staking" or any other option that in a simple way communicates the function of that account.

4) Removed the "recent activity" from design examples as that wasn't implemented. Due to this and responsive views, also brought "Spendable:" closer to the value, as alone it's no longer in relation comp. relation to the expanded data.

vctt94:
This PR closes #1996 

As @linnutee suggested, I updated the default account to `Primary Account` and the imported one to `Locked for Staking`. 

I am also disabling the imported account details when the user have never bought tickets.


app/components/views/AccountsPage/AccountRow/AccountDetails.js
```diff
@@ -117,14 +99,27 @@ const AccountsList = ({
               </div>
               <CopyToClipboard textToCopy={accountExtendedKey} className="account-actions-pubkey-clipboard" />
             </> :
-            <SlateGrayButton className="account-actions-pubkey-button" onClick={onShowPubKey}><T id="account.reveal" m="Reveal"/></SlateGrayButton>
+            <SlateGrayButton className="account-actions-pubkey-button"><T id="account.Hided" m="Hided"/></SlateGrayButton>
```
matheusd: "Hidden"

app/components/views/AccountsPage/AccountRow/Row.js
```diff
@@ -0,0 +1,61 @@
+import { FormattedMessage as T } from "react-intl";
+import { Balance, VerticalAccordion } from "shared";
+import "style/Fonts.less";
+import "style/AccountRow.less";
+
+const isImported = ({ accountName }) => accountName === "imported";
```
matheusd: It's better to check whether it's the imported account by account number (since that's entirely fixed) rather than name.

vctt94:
I see your point, and I think it is makes sense. So I will return it to imported, and if this becomes an issue, we can open a new issue to discuss it, is that fine? 

Also, I am working in the responsiveness on https://github.com/decred/decrediton/issues/1998

That's why I did not deal with it at this PR.

vctt94:
fixed your reviews comments. 

I switched to compare against the imported account's number. For now, I only added a hard coded number, but I will open a new PR organizing better the constants in a single directory and will fix that as well.

xaur:
@matheusd just an idea, if "script" is the only thing that can be "imported", how about `Imported Scripts` as a display name for the `imported` account? Sounds techy, but is more descriptive than just "imported".

PR #2180, opened 2019-08-20
[i18n] Update to react-intl v3

matheusd:
This updates react-intl to major version 3. Due to breaking API changes in
FormattedRelative (now called FormattedRelativeTime) I've re-wrote a
compatible FormattedRelative component in the shared dir.

Also the polyfill for Intl.RelativeTimeFormat can probably be removed after
we upgrade to electron v6.

Also I'm getting an error when rendering the `< 1 minute` message in the sidebar. Please double check if this happening to you as well. I've opened an issue in the upstream react-intl repo.

### DIFF
--- a/app/components/SideBar/LastBlockTime.js
+++ b/app/components/SideBar/LastBlockTime.js
@@ -1,5 +1,6 @@
-import { FormattedMessage as T, FormattedRelative } from "react-intl";
+import { FormattedMessage as T } from "react-intl";
 import ReactTimeout from "react-timeout";
+import { FormattedRelative } from "shared";
 
 @autobind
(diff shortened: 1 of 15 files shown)


vctt94:
So, it looks like react intl does not know how to deal with `<`.

You can remove the < from the message and the error will stop. 

If you return this instead it will work: 

```
<>
  <span>&#60;</span><T id="sidebar.lastBlockIsRecent" m={" 1 minute ago"} />
</>
```

matheusd:
Yeah, that's the one. They have an open issue on that: https://github.com/formatjs/formatjs/issues/148

PR #2181, opened 2019-08-21
Organize constants

vctt94:
Right now decrediton has many constants files around it, making it difficulty to find constants and complicated to know where to add new ones. 

This PR centralize the constants in a single place, and I am also re-adding `babel-plugin-module-resolver`, to make an absolute path of it.


alexlyp:
Needs a yarn re-run

vctt94:
done

PR #2183, opened 2019-08-26
Account view responsiveness

vctt94:
This PR closes https://github.com/decred/decrediton/issues/1998

Right now I am refactoring the `VerticalAccordion` to use its ref's `clientHeight` to make the vertical motion. This way we will end having less trouble to make it responsive. 

It is supposed to close https://github.com/decred/decrediton/issues/1996, but as I touched some other code, like the trezor setup page, I am pushing it so I can get other's opinion, mainly @matheusd , as he coded most of the trezor view.

If others are okay with this approach I can finish the account view responsiveness in this PR.


vctt94:
- Accounts details are now mutually exclusive;
- I did not add overflow: hidden, but I change the effect, can you see if it looks better?

Issue #2061, opened 2019-02-25
Responsive views: Logs


MariaPleshkova:
Responsive views master issue: #1820

Previews: https://xd.adobe.com/view/fe1ba1fc-096f-4299-67b0-d4c2fe379a68-882a/
Specs: https://xd.adobe.com/spec/b4bc18cf-01af-488a-6d7c-9a997d8f5219-c28d/

When scaling down, the width of white area is reduced. The sign, showing estimated time, is deleted (for tablet and mobile).
Also full logo in mobile is changed to DCR symbol only.

Issue #2059, opened 2019-02-25
Responsive views: Launcher overview


MariaPleshkova:
Responsive views master issue: #1820

Previews: https://xd.adobe.com/view/4b423e19-fcf1-44dd-4705-84e02ed3ead0-3d73/
Specs: https://xd.adobe.com/spec/ce37e6a7-cba5-4ee1-5d24-19413a1ac84c-b806/

When scaling down, in tablet if there's only 1 wallet created, all of these square buttons will be in 1 row.
![image](https://user-images.githubusercontent.com/8841018/53337078-626c4200-3911-11e9-8934-f7f16bec7104.png)
But if there are at least 2 wallets created, "Create a new wallet" "Restore existing..." buttons will move down under the first two.

In mobile the size of these buttons was reduced. Launcher was moved to the bottom of the screen. "Settings" and "Logs" items were moved under the Settings icon 
![image](https://user-images.githubusercontent.com/8841018/53337695-2fc34900-3913-11e9-8af6-a811b0bb991f.png)

Also full logo was changed to DCR symbol only.

bgptr:
I have a question. When the "Settings" and "Logs" items move under the Settings icon in mobile, what should happen on click event? Is it just a link to Settings page?

vctt94:
Yeah, right now they are wrapped by an invisible button with an onclick event. 

I believe you can use them to show in the settings icons as well. And if you wrap them by a function, like:

```
const settingsLogsButtons = () => (
  <>
    <InvisibleButton onClick={onShowSettings}>
      <SettingsLinkMsg />
    </InvisibleButton>
    <InvisibleButton onClick={onShowLogs}>
      <LogsLinkMsg />
    </InvisibleButton>
  </>
);
```

I believe it will not be a problem to reuse them.

bgptr:
Ok. Thanks for the quick answer. I'm on it.

bgptr:
@MariaPleshkova, can I get your opinion on the settings icon functioning? There is a conversation already in this PR(#2189)

PR #2189, opened 2019-09-05
Responsive views: Launcher overview and Logs

bgptr:
closes (#2059) and (#2061)

### DIFF
--- a/app/components/views/GetStartedPage/DaemonLoading/Form.js
+++ b/app/components/views/GetStartedPage/DaemonLoading/Form.js
@@ -54,7 +54,6 @@ const DaemonLoadingBody = ({
           </InvisibleButton>
         </>
       </div>
-      <InvisibleButton onClick={onShowSettings} className="settings-button-mobile" />
       <>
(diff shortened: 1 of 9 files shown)


bgptr:
Thanks for the review.

- the overflow-x problem is a good catch, I didn't notice it. It's not always visible, just on certain page widths. I committed a fix for that.
- when clicking at the settings icon, how `logs` and `settings` should show up? 
Like this?
setttings_icon

The settings icon should disappear or it has a toggling mechanism?

MariaPleshkova:
@bgptr you know, I just thought that no need in this icon at all. I mean we can leave them as they are on desktop and tablet views. There are only 2 items under settings icon so I guess they can fit perfectly in one line with the logo even on mobile screens. What do you think?

bgptr:
@MariaPleshkova I like it. Thank you.

bgptr:
I fixed your reviews comments. I changed the menu position from fixed to relative as you requested and changed the logo positioning too.

Issue #1955, opened 2018-12-22
Responsive views: Add a stakepool


linnutee:
Responsive views master issue: https://github.com/decred/decrediton/issues/1820 

Preview: https://xd.adobe.com/view/937f62e2-1ecc-417b-7f8f-b338e3a418d7-e54e/
Specs: https://xd.adobe.com/spec/0196c0e4-c039-42c7-7bca-ff1b825cedb5-77a7/

Card content repositioned top down.

PR #2190, opened 2019-09-09
Responsive views: Add a stakepool

bgptr:
closes (#1955)


bgptr:
Thank you for the review.
- I did not remove the icon, because it has a function: "Manually import a redeem script for ticket". Is this function moved to somewhere else or is it deprecated?
- Sorry, but I don't really understand your second comment: "Also, the content is not with a fixed width. so depending on the size it gets bigger or smaller." Could you explain it a bit more?

bgptr:
I fixed your reviews comments and tweaked a bit the margins.

Issue #2185, opened 2019-08-30
[Governance] Change how proposals are fetched from politeia


vctt94:
Right now, decrediton is fetching proposals by getting the vetted proposals and updating them with the `votestatus` endpoint, and after all that each proposal updates its votestatus individually.

Besides this take a lot of time the `votestatus` endpoint is being deprecated by the `batch` endpoint, so I believe we should change for that as it will simplify how we are fetching proposals right now and make lazy loading easier.

source:
https://github.com/decred/politeia/blob/master/politeiawww/api/www/v1/api.md#proposal-vote-status .

Issue #2015, opened 2019-02-06
Proposal categories inconsistent - 'Under discussion' vs. 'Pre-voting'


dre-tech:
In Decrediton the proposal category is called 'Under Discussion'. In Politeia it is called 'Pre-Voting'. We need to pick one and make it consistent in both.

linnutee:
Referred to Politeiagui repo, better to get it sorted there first.

Issue #2166, opened 2019-07-31
Excessive data download


jholdstock:
#### The problem
When a wallet is opened in Decrediton, proposal voting data is immediately downloaded from Politeia. This includes all vote data for all proposals, which currently represents a >40MB download every single time Decrediton opens a mainnet wallet. 

This is very wasteful, and seems to be getting worse every time voting is opened on a new proposal.

I can see that gzip compression is already being applied to the data, so I don't think we have any easy savings there. Server response time is not great, so there may be savings possible there, but for this issue lets focus on the sheer amount of data downloaded.

#### Analysis of the data

`/api/v1/proposals/{proposal_id}/votes` is called once per proposal ([example](https://proposals.decred.org/api/v1/proposals/5431da8ff4eda8cdbf8f4f2e08566ffa573464b97ef6d6bae78e749f27800d3a/votes)), downloading ~3MB each time. Each of the responses comprises:

- Tickets IDs eligible for voting (80% of total data)
- Votes cast 
	- Signatures (6.5%)
	- Ticket IDs (3.5%)
	- Tokens     (3.5%)

#### Suggestions

- The vast majority of the data, the eligible tickets, is static and known from the moment voting begins. This could be downloaded once and cached by decrediton indefinitely. Downloading it multiple times is wasteful.
- For votes which are known to be concluded, data is static and can also be cached. Only votes which are currently in progress need to be polled.
- Does decrediton even need all of this data? The vote signatures, tokens and ticket IDs for **all** votes?
- When a wallet is closed and re-opened (without closing decrediton) the data already in memory should be persisted and not re-downloaded.
- Perhaps we could defer downloading the data for historical votes which have already concluded? Does historical data really need to be downloaded immediately upon opening the wallet?

lukebp:
Decredition needs the eligible tickets snapshot in order to check if the user has any tickets that are eligible to vote.  This data could be fetched directly from dcrwallet though so that you don't have to fetch such a large payload from the Politeia server.  You can get the vote start block height from the `VoteStatus` endpoint, a lightweight version of the `VoteResults` endpoint that is referenced above, then use it to determine the block height of the vote snapshot.

Note: there are plans to deprecate the `VoteStatus` endpoint in the future.  decred/politeia#953 adds the vote summary information that is returned by `VoteStatus` to a proposal record by default.

lukebp:
[...]

Issue #1945, opened 2018-12-22
Visual Design: Governance subviews tweaks and bug fixes


linnutee:
Preview: https://xd.adobe.com/view/fefd1452-fc57-4418-6610-736fa70425db-88cc/
Specs: https://xd.adobe.com/spec/fb0fd24e-9c3c-4854-656e-c7bc8332e18d-8d3e/

Governance view/all proposals: 
- I've reduced the new proposal row styling to a much more minimal variant, as the previous got quite bloated. Only a 5px stroke on the left side for new proposals, instead of full row coloring.
- New item number ticker for `Under Vote` and `Under Discussion` seems to be missing a fill value. Besides the stroke around it, the circle also needs a solid fill of `#8997A5` in order for the white background to show. ![](https://user-images.githubusercontent.com/17774057/50377719-0d6b3180-062b-11e9-90ed-55bc5d9606ac.png)
- For individual proposals view, added a white background solid to the proposal content area to make the view more cohesive. There is also a 1px seperation between the high level details and the proposal content. 
- The individual proposal view is currently partially implemented, as the content area text styles seem off/missing and the text area container is also too long in width (should be 380px). 
- All bolded text styles should be with Source Sans Pro SemiBold. There should not be any full Bold in Decrediton, only SemiBold.
- Added a state for "Proposal not yet in Voting stage." Icon: [chatBubbles.svg.zip](https://github.com/decred/decrediton/files/2705465/chatBubbles.svg.zip)
- Added a disabled state for the expanded consensus cards. When a consensus vote has been finished, it should be disabled (currently some of the elements are in active states, some not). 
- Updated the expanded consensus cards `My Voting Preference` slightly to make it consistent with the same function in proposals (not sure if it's fully implemented there). In the current release, the radio buttons are slightly off, having too thick strokes. (Note: there should be 5 types of radio buttons now: blue for consensus changes, orange/green for proposals, default grey/blue for settings or generic use and disabled state).

vctt94:
I can fix these

linnutee:
[...]

PR #2192, opened 2019-09-24
Fix governance tweaks

vctt94:
I started this PR working on https://github.com/decred/decrediton/issues/1945 but as the politeia's code had suffered some changes I took the time to update decrediton's code as well.

 closes https://github.com/decred/decrediton/issues/2185 and closes https://github.com/decred/decrediton/issues/2015

Still have some things left to do but as this PR already got too big, I will work on them on https://github.com/decred/decrediton/issues/2166


vctt94:
Good catch. I fixed it now

app/components/buttons/EnableExternalRequestButton.js
```diff
@@ -2,11 +2,11 @@ import { settings } from "connectors";
 import KeyBlueButton from "./KeyBlueButton";
 
 const EnableExternalRequestButton = ({ children, requestType,
-  onAddAllowedRequestType }) => {
+  onAddAllowedRequestType, onClick }) => {
 
-  const onClick = () => onAddAllowedRequestType(requestType);
+  const onClickFunction = () => onClick ? onClick() && onAddAllowedRequestType(requestType) : onAddAllowedRequestType(requestType);
```
matheusd: Shouldn't this be the other way around (`onAddAllowedRequestType(requestType) && onClick`)? So that you're guaranteed that when `onClick()` is triggered the new request type will be allowed to be done?

Otherwise you might end up with a race where onClick() tries to execute a given request but that one isn't allowed yet.

vctt94: yeah, you are right. 

I fixed it as well

PR #2107, opened 2019-05-06
[ln] Initial Version

matheusd:
This is the first version of an embedded LN wallet UI on decrediton.

Current features:
  - Connect to initialized and unlocked ln wallets
  - Fund/Withdraw funds from onchain ln wallet
  - Open/Close channels
  - Create invoices
  - Send payments

Due to current limitations of dcrlnd, this can only be used on wallets that
have been previously unlocked.

It also requires the user to open a regular wallet before connecting to a ln
wallet, which is something we might wanna change in the future.

For the moment, in order to enable support to LN, users need to manually edit
the global decrediton config file to set `"ln_enabled: true"`.

It's likely that this can only be merged after updating the grpc and electron versions (which will come on a separate PR)

PR #2208, opened 2019-10-11
Upgrade to electron 8

matheusd:
Requires #2107


matheusd:
Updated PR to upgrade the deps to electron 8. 

One thing of note is that i had to add an upgrade path so that dcrd's tls cert and key files are removed so they can be recreated using curve P-256 as curve P-521 is no longer supported in electron.

PR #2261, opened 2019-11-07
Accounts quick preview enhancements

### LINKED ISSUE #2233: Visual Design: Accounts quick preview enhancements
Preview: https://xd.adobe.com/view/f1318932-7de3-42cd-5dd6-529bac24dd6b-8700/
Specs: https://xd.adobe.com/spec/7eb6609f-3cee-4e35-77ef-43aa13481c59-0887/
Icons: [Account Icons.zip](https://github.com/decred/decrediton/files/3771566/Account.Icons.zip)

Relates to: https://github.com/decred/decrediton/issues/1520

As it’s not overly clear for new users what the quick account component communicates, added a few enhancements to improve this:
- Added icons to create more continuity and relation to the Accounts view
- Inconsolate font is replaced with Source Sans Pro as part of phasing it out altogether
- Full number and two first decimal spots are semibold, rest of the decimal spots regular (instead of 2 sizes)

bgptr:
closes #2233

### DIFF
--- a/app/style/SideBar.less
+++ b/app/style/SideBar.less
@@ -116,18 +116,10 @@
   font-family: var(--font-family-regular);
   color: var(--input-color-default);
   font-size: 15px;
+  font-weight: 600;
 
-  span {
+  .balance-small {
     font-size: 15px;
-    font-weight: 600;
-    
-    &.balance-small {
-      font-weight: 400;
-    }
-
-    &.currency {
-      display: none;
-    }
   }
 }
 
--- a/app/style/SideBar.less
+++ b/app/style/SideBar.less
@@ -117,15 +117,15 @@
   color: var(--input-color-default);
   font-size: 15px;
 
-  div {
+  span {
+    font-size: 15px;
     font-weight: 600;
-
-    .balance-small {
+    
+    &.balance-small {
       font-weight: 400;
-      font-size: 15px;
     }
 
-    .currency {
+    &.currency {
       display: none;
     }
   }
--- a/app/style/SideBar.less
+++ b/app/style/SideBar.less
@@ -87,13 +87,21 @@
 
 .sidebar-menu-total-balance-extended-bottom {
   padding-right: 18px;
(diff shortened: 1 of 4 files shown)


bgptr:
Interesting. It looks like this on my side:
image
I try to figure out why.

bgptr:
I did some modifications. I hope it'll works for you too.

app/style/SideBar.less
```diff
@@ -105,25 +113,53 @@
 }
 
 .sidebar-menu-total-balance-extended-bottom-account-number {
-  font-family: var(--font-family-monospace);
+  font-family: var(--font-family-regular);
   color: var(--input-color-default);
   font-size: 15px;
+
+  span {
+    font-size: 15px;
```
matheusd: This is the offending rule in my case. Commenting this makes the sizes display correctly.

Weird it's not failing for you. Which OS are you using?

bgptr: I use macOS Catalina. If I remove the suggested lines, then the fonts are too small: 
Screenshot 2019-11-19 at 21 35 46


In the design specs, the bold and the regular part of the text are 15px too. 
I think in your case the normal part is good and the bold is too small somehow.
I will try to reproduce this on Linux.

bgptr: Sorry, but I could not reproduce it on Linux either. Maybe it is because I run it in a virtual machine.
Screenshot 2019-11-20 at 11 31 02

bgptr:
This should be work.

bgptr:
@vctt94 my last commit did not solve the font-size issue on your side?

vctt94:
No, still looks the same size, but bolder:

image

Maybe using pt would be a better approach, as px depends on the resolution

bgptr:
I've changed it to pt. Could you check it, please?

vctt94:
Yeah, it looks better. Thanks! 

image

bgptr:
Thank you!

PR #2280, opened 2019-11-09
Change stakepool occurrence to VSP

vctt94:
closes #2275 and closes #1615
 
I am changing only occurrences that are shown at the screen, as we have many places with stakepool. 

If you guys rather I can change the other occurrences, but I am not sure of the need.

PR #2290, opened 2019-11-11
Fix launcher icon

### LINKED ISSUE #2278: [1.5.0-rc1] No icon displayed for Decrediton in launcher on Linux
Decrediton just shows a transparent block instead of the appropriate logo in the launcher.

Tested on Ubuntu 19.10.

image

vctt94:
closes #2278 

One thing I noticed is when hovering the icon it shows Electron instead of decrediton.

I couldn't find a solution, I looked through their documentation but there is no mention on how to change it: https://electronjs.org/docs/api/browser-window


matheusd:
Hmm, did you test this in ubuntu? I'm getting exactly the opposite behavior: the tooltip shows "Decrediton" but the icon is still transparent.

I'm testing via this binary: https://github.com/matheusd/decred-weekly-builds/releases/tag/v20191115153955

alexlyp:
These changes had already been mistakenly merged in a previous PR: https://github.com/decred/decrediton/pull/2280

These changes were shown to cause electron-builder to not properly find the icon to use during the build process.  If we do change the location of the icons to inside app/ we need to specify that location with the options provided by electron-builder.  Believe it's "buildResources"  

You can read up about it here: https://www.electron.build/configuration/configuration and https://www.electron.build/icons

alexlyp:
I'm not able to properly see the icon on my dev setup here, but I'm thinking it's due to my use of VMs etc.  Maybe someone with an Ubuntu machine can comment.

matheusd:
@alexlyp `yarn dev` doesn't show the icon, but `yarn package-dev-linux` does and adding a rule to generate an appImage (replacing `--linux tar.gz` for `--linux appImage`) also shows an image for me, even in the VM.

vctt94:
rebased

PR #2302, opened 2019-11-13
layout: fix linter warnings

dajohi:
Expected space before "-" operator   function-calc-no-invalid
Expected space after "-" operator    function-calc-no-invalid
Expected space before "-" operator   function-calc-no-invalid
Expected space after "-" operator    function-calc-no-invalid
Expected space before "-" operator   function-calc-no-invalid
Expected space after "-" operator    function-calc-no-invalid

### DIFF
--- a/app/style/Governance.less
+++ b/app/style/Governance.less
@@ -321,7 +321,7 @@
 }
 
 .proposal-details-text {
-  font-family: @font-family-regular-semi-bold;
+  font-family: var(--font-family-regular-semi-bold);
   min-height:100%;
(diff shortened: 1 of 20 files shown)


dajohi:
See https://developer.mozilla.org/en-US/docs/Web/CSS/calc#Examples under the `Nested calc() with CSS Variables` section

dajohi:
Hmm, I wonder if var() is correct for an at-rule

alexlyp:
This needs a rerun to fix the conflicts. @dajohi

PR #2349, opened 2019-11-27
[protoc] Update middlewares

vctt94:
This PR updates our api.proto file to the latest dcrwallet's version. 

I used protoc v3.4.0: https://github.com/protocolbuffers/protobuf/releases/tag/v3.4.0.

When regenerating it switch the generated code from `Buffer()`, which is being deprecated, to `Buffer.from()`


matheusd:
Btw, this is a major step towards #1421

PR #2388, opened 2020-01-06
Disable element selection

degeri:
This fixes an annoying behaviour where text is draggable. Also makes the app feel more "solid"

**Before:**
Peek 2020-01-06 15-31

**After:**
After

Based on https://gist.github.com/msuchodolski/41c9338f3a732379dc570142826ed76e

### DIFF
--- a/app/style/main.less
+++ b/app/style/main.less
@@ -134,7 +134,7 @@ main {
 
 //Exceptions
 
-.standalone-page.txdetails-standalone-page,.snackbar,.purchase-ticket-area,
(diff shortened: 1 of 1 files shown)


degeri:
Ok will try to find a solution for that.

degeri:
Allowed the following:
transaction details page,snackbar,Purchase Tickets ,ticket list,Statistics and Accounts . 

Let me know if I missed something.

degeri:
It both of these already had a copy to clipboard action. Made it selectable now.

PR #2394, opened 2020-01-16
fix closing second instance

### LINKED ISSUE #2381: Cannot close message that appears when multiple instances of decrediton are run.
When opening more than one instances of decrediton by accident the warning message does not close by simply pressing the X button, it requires the user to kill the task via task manager (windows) or xkill command (Linux):

image

zubairzia0:
The second instance message window pops up but cannot be closed, so I added instruction for quitting. 
closes #2381


zubairzia0:
Closing PR  because of the build and test failure. Will open a new PR.

PR #2395, opened 2020-01-16
Import to csv notification quoting fix

### LINKED ISSUE #2356: [Transactions] Import to csv
1.5.0 rc2 debian, mainnet

in notification, filename not found
screeshot

bgptr:
closes #2356


bgptr:
Interesting. I changed it to the unicode characters.

vctt94:
Also, Matheus can correct me if I am wrong, but I believe there is no need to update original.json, as well.

bgptr:
@vctt94 Sorry, I have reverted it.

PR #2398, opened 2020-01-17
[Refactor startup] Fix starting UX and clickables

vctt94:
This PR adds the ability to click at settings and logs and fix some UX feedbacks. 

Right now the initial startup (when not having a config.json file) is not working, but I'll fix it when working in the spv start, as the daemon is started after finishing the SPV choice, at `finishSpvChoice`.

This refactor is now close to an end. After this PR will only be lefting the spv start and some clean up.

 Please let me know if you guys find some inconsistency of the ux or some bug.


vctt94:
I fixed your first two comments: 

```
"Blockchain download estimated complete" never disappears after the download is complete
Opening and closing a wallet keeps an old "Time from last fetched header: ..." log line visible
```

The seed confirmation I will fix in a next PR, which I will also add watching only wallet and trezor wallet creation

And the same for SPV.

PR #2407, opened 2020-01-27
fix treasury balance

### LINKED ISSUE #2341: Governance Tab: Available Treasury Balance
I am using Decrediton - 1.5.0 rc 2 and noticed an issue in the governance tab with the available treasury balance. It says 63,610 while dcrdata is 636,106. Appears the comma and decimal point may be in the wrong spot?

image

### LINKED ISSUE #2406: [1.5.0] Treasury balance is wrong on Governance tab
Screenshot from 2020-01-27 09-53-46

vctt94:
fixes #2406 and fixes #2341. 

The reason this bug only happen sometimes, is because dcrdata sends a floating number, so it can send numbers with less than 8 decimal places, sometimes (maybe some rounding, Idk). 

This PR fixes it.

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -1297,9 +1297,7 @@ export const getTreasuryBalance = () => (dispatch, getState) => {
       const integerPart = splitedTreasuryInfo[0];
       // dcrdata can send numbers with its decimal part less than 8 decimals, so we manually add it.
       let decimalPart = splitedTreasuryInfo[1];
-      while (decimalPart.length < 8) {
-        decimalPart += "0";
-      }
+      decimalPart += "0".repeat(8 - decimalPart.length);
       const treasuryBalance = integerPart + decimalPart;
       dispatch({ treasuryBalance, type: GETTREASURY_BALANCE_SUCCESS });
     });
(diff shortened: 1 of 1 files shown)


app/actions/ClientActions.js
```diff
@@ -1293,7 +1293,14 @@ export const getTreasuryBalance = () => (dispatch, getState) => {
...
     .then(treasuryInfo => {
       // Manually convert DCR to atom amounts to avoid floating point multiplication errors (eg. 589926.57667882*1e8 => 58992657667881.99)
-      const treasuryBalance = parseInt(treasuryInfo["data"]["dcr_unspent"].toString().replace(".",""));
+      const splitedTreasuryInfo = treasuryInfo["data"]["dcr_unspent"].toString().split(".");
+      const integerPart = splitedTreasuryInfo[0];
+      // dcrdata can send numbers with its decimal part less than 8 decimals, so we manually add it.
+      let decimalPart = splitedTreasuryInfo[1];
+      while (decimalPart.length < 8) {
+        decimalPart += "0";
+      }
```
matheusd: No need for the loop. You could use `"0".repeat(8-decimalPart.length) but given that `treasuryBalance` is supposed to be in atoms I believe you can just do `integerPart*1e8 + decimalPart`.

vctt94: I was not aware of this string.repeat method. I switched to it, as to switch to `integerPart*1e8 + decimalPart` I would need to parse them to numbers, so they do not concat.

PR #2411, opened 2020-02-02
[Refactor_Startup] Add Restore watching only and trezor wallets

vctt94:
This PR is built on top of #2398 

- I fixed skipping confirm seed when not in mainnet
- I added restore watching only wallets
- I added a creating wallet state, which right now is adding a `isCreatingWallet` props to true. ~~is only rendering `DecredLoading` component. We can improve this state with some text or other component, but right now it is only to get feedbacks.~~
- ~~I did not add restore trezor wallets, because I do not have a trezor device.~~
- Restore Trezor wallets is working.
- Most of restore, creating wallet should be working by now, so if you guys find any bugs please let me know.

- SPV start still not working. It will be done in the next PR


alexlyp:
The text on the loading bar isn't changing and simply saying Syncing RPC connection...

Should say what step of the syncing process it's on.  This is for a basic watch only wallet.

PR #2415, opened 2020-02-08
Remove slack icon from Help tab

### LINKED ISSUE #2410: Remove Slack from help page
Slack is being retired at the end of February, it should no longer be listed as an option on the help page under "communications".

unimere:
Closes #2410


alexlyp:
So this has a lot of additional stuff in other than removing the icon from the help page.  Please revert all the other translation changes.  Those can be submitted seperately, plus you only need to update the .pot and not every translation reference.

unimere:
I've updated this to only remove the icon from the Help tab, remove the corresponding css/less code, and removed a single slack reference in the readme. I'll open a new PR to update the FatalErrorPage.js and the corresponding .pot files

alexlyp:
Ok great, but there is no need to touch the .pot files.  Those come directly from Transifex.  Just update the string on FatalErrorPage.js and it will get updated in due time.

alexlyp:
Ok great, but there is no need to touch the .pot files.  Those come directly from Transifex.  Just update the string on FatalErrorPage.js and it will get updated in due time.

PR #2416, opened 2020-02-10
Add tx's addresses to overview and transactions view

### LINKED ISSUE #2232: Overview and Transactions Views Addressing
Preview: https://xd.adobe.com/view/759d5f89-ed12-4c01-7bac-f649d443b566-03ac/
Specs: https://xd.adobe.com/spec/ae63447d-0b3e-45ed-7e12-a548eb2330f5-d852/

The account names can be leveraged in the transactions rows to give a bit better idea of what’s what. Applies to both Overview and Transaction History.

vctt94:
closes #2232

As the addresses we need to show are from non wallet inputs and outputs, we need to decode the raw transaction to show them. 

The problem with that approach is that it gets complicated to get addresses of inputs from received transactions, as they are shown as the previous outpoint. So right now I am showing the previous outpoint, which I am not sure if it is a relevant information for the user.

Here it is how it looks at overview page, so you guys can see it: 
image

### DIFF
--- a/app/components/TxHistory/TxHistoryRow/RegularTxRow.js
+++ b/app/components/TxHistory/TxHistoryRow/RegularTxRow.js
@@ -4,20 +4,26 @@ import { createElement as h } from "react";
 import { timeMessage } from "./index";
 import { FormattedMessage as T } from "react-intl";
 
-const accountMessage = <T id="txHistory.account" m=" Account" />;
-
 const RegularTxRow = ({
-  txAmount, txDirection, overview, txAccountName, pending, txTs,
-  txOutputAddresses, ...props
+  txAmount, txDirection, overview, txAccountName, pending, txTs, txOutputAddresses,
+  txAccountNameCredited, txAccountNameDebited, ...props
 }) => (
   <Row {...{ ...props, txAccountName, pending, overview }}>
     <div className="is-row">
       <span className="icon" />
       <span className="transaction-amount-number"><Balance amount={txDirection !== "in" ? -txAmount : txAmount} /></span>
       { !overview && (txDirection === "transfer" ?
-        <div className="transaction-status">
-          <span className="transaction-account-name">{txAccountName}</span>
-        </div> : txDirection !== "in" ? (
+        ( <div className="transaction-info is-row">
+          <T id="txHistory.from" m="From " />
+          <div className="transaction-status">
+            <span className="transaction-account-name">{txAccountNameDebited}</span>
+          </div>
+          <T id="txHistory.to" m=" To " />
+          <div className="transaction-status">
+            <span className="transaction-account-name">{txAccountNameCredited}</span>
+          </div>
+        </div>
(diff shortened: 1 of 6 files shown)


vctt94:
If @hqnnes also can take a look at this PR, as this is the one who created the issue, I'd appreciate it.

matheusd:
To be honest, other than the account name (of which there could actually be multiple in more unusual transactions) I believe the other info is useless for a quick glancing in the overview and I'd say we can probably remove it.

So personally I'd show "to [account name]" in received transactions, "from [account name]" in transactions that were sent and "from [account name] to [account name]" in transfers.

If we wanted to be fancy, we could also try and id split transactions that were used to purchase tickets and label those differently but that's a future improvement.

vctt94:
I agree with you in parts, but I think it is reasonable to have the address in transactions that were sent, as you need to fill the address to send it. 

So I removed the outpoints information and left  `from [account] to [address]` in sent transactions.

At transfers transactions, as it is a fee, there is no way to know the account's name (I believe), so I just left `from account` and transactions that were received I left only `to [account]`. 

I believe it is more reasonable now.

matheusd:
Transactions are labeled as transfers by decrediton ([here](https://github.com/decred/decrediton/blob/c9f4404cbe20c5ce6aba9273e8344d0dc2dc7889/app/wallet/service.js#L108-L116)) if `sum(debits) - sum(credits) == fee`. 

In other words, you need _at least_ one debit and one credit (respectively: one input and one output) in the transaction such that their difference is exactly equal to the fee. So you can figure out the probable destination account by checking the first credit.

Obviously this is only valid for simple transactions. One could manually create, sign and publish a transaction involving credits and debits from multiple (besides non-wallet inputs/outputs) but Decrediton doesn't currently create those so it's ok to not show the full info IMO.

vctt94:
You are right. I added the debited and credited account into transfer txs

app/components/TxHistory/TxHistoryRow/RegularTxRow.js
```diff
@@ -47,8 +53,11 @@ const RegularTxRow = ({
           { txDirection === "transfer" ?
             <>
               <T id="txHistory.from" m="From " />
-              { txAccountName }
-              { accountMessage }
+              { txAccountNameDebited }
+              <T id="txHistory.account" m=" Account" />
+              <T id="txHistory.to" m=" To " />
+              { txAccountNameCredited }
+              <T id="txHistory.account" m=" Account" />
```
matheusd: Much nicer!

For i18n purposes, it's better to use a single parametrized message (so instead of individual `<T>` use a single `<T m="From {srcAccount} to {dstAccount}"`). I believe the `" Account"` suffix is redudant.

vctt94: Agree. I also changed it for tx in and out as well.

PR #2417, opened 2020-02-10
[Refactor_Startup] SPV

vctt94:
This PR is built on top of #2411 

This pr enables spv start and starts to clean up the code. 

As I started to touch too many files while cleaning up the code: reducers, selectors, connectors, etc...

I am opening this PR to make it easier to test only the spv start and the next one I'll deal with all the clean up


alexlyp:
So when I start in non-SPV it starts to sync, then when i switch to SPV while still in app it appears to just hang.  nothing happening in logs.  Just says dcrd exited and it's showing the "syncing Daemon..." in the loading bar.

alexlyp:
When starting in SPV mode I am able to select a wallet to load, but then it appears to hang as well.  

Currently just sitting after the logs say Opened Wallet, and Syncing RPC connection on the loading bar.

PR #2420, opened 2020-02-17
[governance] Check for new politeia proposals

### LINKED ISSUE #2414: List of proposals becomes stale
When I open a wallet and click the governance tab the proposals view loads correctly to show the latest proposals "under discussion" and "voting".
I then leave the governance page, go to any other page, and leave the wallet open for some time. 
When I go back to the governance page it does not refresh the list of proposals "under discussion" and "voting".  For example, new proposals that were started after the wallet was opened will not show in the list and old proposals that have finished will show in the wrong place.

The simple workaround for this is to reload the UI or close and re-open the wallet but someone looking in Decrediton for new proposals won't find them unless they realize that they need to do this.

vctt94:
Closes #2414


app/actions/GovernanceActions.js
```diff
@@ -102,21 +127,53 @@ const getTokenInventory = () => async (dispatch, getState) => {
...
+    const inventory = updateInventoryFromApiData(data);
 
-    inventoryTabs.preVote = data.pre;
-    inventoryTabs.activeVote = data.active;
-    inventoryTabs.finishedVote = [ ...data.approved, ...data.rejected ];
-    inventoryTabs.abandonedVote = data.abandoned;
+    const inventoryDiff = getDefaultInventory();
+    let isDifferent = false;
+
+    Object.keys(oldInventory).forEach( async key => {
```
matheusd: Why `async`? The closure doesn't seem to be waiting for anything.

app/actions/GovernanceActions.js
```diff
@@ -102,21 +127,53 @@ const getTokenInventory = () => async (dispatch, getState) => {
...
+      if (inventory[key].length > oldInventory[key].length) {
```
matheusd: This is broken. 

While unlikely to happen in practice, consider what would happen if (during the 60s interval between calls to this func) a new proposal entered the `preVote` list while 2 proposals left `preVote` and entered `activeVote`: the  new `preVote` list (in `inventory`) would have smaller `length` but new items (which wouldn't get picked up).

If the goal is to fetch the details of any new, as of yet unknown proposal (irrespective of which list it is on) it's probably simpler to just create (or maintain) a map of known proposal hashes (i.e. a map of proposal token => proposal) and then just check (for every token) whether that entry is filled (and if not, request it).

vctt94:
You are right and the logic was wrong. 

I changed it and believe it is working properly now.

app/actions/GovernanceActions.js
```diff
@@ -153,41 +153,47 @@ export const compareInventory = () => async (dispatch, getState) => {
...
-      const diffHashes = inventory[key].filter( token => !oldInventory[key].includes(token) && (isDifferent = true));
+    // create array with all old inventory and inventory token's values
+    const flatOldProps = [oldInventory.activeVote, oldInventory.abandonedVote, oldInventory.finishedVote, oldInventory.preVote].reduce( (acc, v) => {
+      v.forEach(vp => acc.push(vp));
+      return acc;
+    }, []);
+    const flatNewProps = [inventory.activeVote, inventory.abandonedVote, inventory.finishedVote, inventory.preVote].reduce((acc, v) => {
+      v.forEach(vp => acc.push(vp));
+      return acc
+    }, []);
```
matheusd: A more idiomatic way to write this in modern js would be to just do `[...inventory.activeVote, ...inventory.abandonedVote, ...]`. Saves an identation...

app/actions/GovernanceActions.js
```diff
@@ -153,41 +153,47 @@ export const compareInventory = () => async (dispatch, getState) => {
...
-        if (!inventory[key].includes(proposal.token)) {
-          oldProposals[key].splice(index, 1);
+    // also remove it from proposal's list.
+    Object.keys(oldProposals).forEach( key => {
+      newProposalsList[key] = oldProposals[key].reduce((acc, p) => {
+        if(inventory[key].includes(p.token)) {
+          acc.push(p);
+        } else {
+          // Remove token from old inventory so we can re-fetch it.
+          oldInventory[key].splice(oldInventory[key].indexOf(p.token), 1)
```
matheusd: I still think modifying `oldInventory`, even if it doesn't produce local side-effects in this function can cause side-effects in the app. 

As a rule of thumb you should **not** modify any objects returned from the redux state storage (even if they were indirectly returned via a select). I'd suggest duplicating `oldInventory` before starting the main `Object.keys(oldProposals).forEach(...)` loop.

app/actions/GovernanceActions.js
```diff
@@ -153,41 +153,47 @@ export const compareInventory = () => async (dispatch, getState) => {
...
+      return acc
+    }, []);
 
-      if (isDifferent) {
-        isDifferent = false;
-        dispatch(getProposalsAndUpdateVoteStatus(diffHashes));
-      }
+    // Get difference between new inventory and old one, so we can bring a batch
+    // of new proposals.
+    const diffHashes = flatNewProps.filter( token => !flatOldProps.includes(token));
```
matheusd: This is fine for the current number of props but bear in mind this is an O(n^2) algo. It would be better to convert `flatOldProps` to a hashmap so you can directly check for element inclusion (instead of using `includes`) or sort both arrays and check them sequentially.

vctt94: Changed for map instead

app/actions/GovernanceActions.js
vctt94: Making a copy of oldInventory now

app/actions/GovernanceActions.js
vctt94: Changed for map instead

Issue #1491, opened 2018-06-25
Trezor Support


matheusd:
Meta issue for keeping track of work needed to integrate trezor. I'll try to do some of the required work on separate PRs without directly referencing trezor so that we can start integrating in stages, instead of doing only a single very large PR at the end. Even so, it's likely that some of the PRs will be pretty big.

Still subject to a lot of change at the moment. Order of execution is not fixed.

Proof of concept stage
- [ ] Generate address
- [ ] Sign Tx
- [ ] Obtain trezor masterpubkey
- [ ] Support use passphrase to encrypt trezor wallet
- [ ] Sign and Verify messages

Basic decrediton integration
- [ ] Upstream mainnet coin info to trezor firmwares
- [ ] Detect and react to trezor events (trezor connected/disconnected)
- [ ] Import wallet from trezor screen (wallet creation)
- [ ] Change instances of passphrase modal to use trezor confirmation
- [ ] Change or disable unsupported features (message signing/verification, tickets, etc)

Advanced decrediton integration (not everything will be feasible at this moment)
- [ ] Support signing tickets (requires ability to import scripts on watch only wallets)
- [ ] Support multiple accounts (requires ability to import root xpub instead of account xpub)
- [ ] Support trezor passphrase input
- [ ] Support reset and reseed trezor directly
- [ ] Support importing trezor native seed as decred seed (might require separate tooling)

linnutee:
Here's the updated DCR symbol generated for Trezors home screen. 
Also included the complete updated logo and symbol package, which can be used in Trezor's various web instances where needed. 

[dcr - trezor - homescreens.zip](https://github.com/decred/decrediton/files/2138445/dcr.-.trezor.-.homescreens.zip)

[Decred - Logo Package.zip](https://github.com/decred/decrediton/files/2138459/Decred.-.Logo.Package.zip)

Grum1in:
So, Do you do that anybody can buy tickets via Trezor?

alexlyp:
Not yet.

matheusd:
Ticket purchasing Proof of Concept:

Firmware changes: https://github.com/trezor/trezor-mcu/compare/master...matheusd:ticket-poc
Wallet changes: https://github.com/decred/dcrwallet/compare/master...matheusd:trezor-tickets-poc
Client changes: https://github.com/matheusd/trezor-decred-poc/compare/ticket-poc

Rough outline:

[...]

PR #2423, opened 2020-02-19
Import script watch only

vctt94:
This PR adds the ability to watching only wallets import scripts.

It is a work towards https://github.com/decred/decrediton/issues/1491

### DIFF
--- a/app/wallet/control.js
+++ b/app/wallet/control.js
@@ -48,17 +48,14 @@ export const importPrivateKey = (walletService, passphrase, accountNum, wif, res
     walletService.importPrivateKey(request, (err, res) => err ? fail(err) : ok(res));
   });
 
-export const importScript = (walletService, passphrase, script, isWatchingOnly) =>
+export const importScript = (walletService, passphrase, script) =>
   new Promise((ok, fail) => {
     const request = new api.ImportScriptRequest();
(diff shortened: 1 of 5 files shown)


app/wallet/control.js
```diff
@@ -48,14 +48,17 @@ export const importPrivateKey = (walletService, passphrase, accountNum, wif, res
...
     request.setScript(new Uint8Array(Buffer.from(hexToBytes(script))));
-    request.setRescan(rescan);
-    request.setScanFrom(scanFrom);
-    request.setRequireRedeemable(true);
+    request.setRescan(false);
+    request.setScanFrom(0);
+    // watch only wallets are not redeemable of their scripts, therefore we
+    // don't require them to be redeemable.
+    const isRequireRedeem = isWatchingOnly ? false : true;
+    request.setRequireRedeemable(isRequireRedeem);
```
matheusd: This isn't true. The "redeemable" part checks whether one of the addresses of the multisig belogs to the wallet, irrespective of whether the wallet has the private keys for that address or not.

So I think this should remain `true` even for watching only wallets since it ensures the corresponding (cold) wallet will have the keys for at least one of the pubkeys in the script.

vctt94: Yeah, you are right. Fixed

PR #2430, opened 2020-02-28
Disable sandbox

vctt94:
After d3e18e5313b25941574a8cd541a69b6ab885dae2, I hitted an error complaning about sandbox, so this PR deactivate it in dev mode.

Related: https://github.com/electron/electron/issues/17972

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "build": "npm run build-main && npm run build-renderer",
     "rebuild-natives": "cd app && ../node_modules/.bin/electron-rebuild",
     "start": "cross-env NODE_ENV=production electron ./app/ --debug --custombinpath=./bin",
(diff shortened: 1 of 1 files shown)


vctt94:
Yeah, it does

output running decrediton after install decrediton-v1.5.0.deb: 
```
:~/projects/decred/decrediton/release$ decrediton 
[6756:0303/181012.748186:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /opt/decrediton/chrome-sandbox is owned by root and has mode 4755.
Trace/breakpoint trap
```

xaur:
Does disabling sandbox degrade user's security?

vctt94:
sandbox can improve security, but we are not using it in decrediton right now. 

But after talking we decided to not remove it and returned it at: github.com/decred/decrediton/commit/e73ed049e5b0a50b310296381dbe85401713bdbd

Issue #2262, opened 2019-11-08
Visual Design: My Tickets improvements


hqnnes:
Preview: https://xd.adobe.com/view/9624d1d3-0a79-4849-5c49-f1850c2ba60f-e015/
Specs: https://xd.adobe.com/spec/23da73ae-3e63-4d01-498b-5ee66a5e5189-55eb/

To give the user a clearer vision we reorganised the components content from two rows to one row. Also added a table header to make it easier to grasp the info.
* component's content from two rows to one row
* a table header feature (similar to the "Date index feature" in #2252)

linnutee:
@vctt94 this is the follow-up per our discussion on removing tickets from history.

vctt94:
Thanks linnutee!

After our release I will take a look at this!

vctt94:
I am working on this issue now, but I found an information missing. 

Right now ticket votes row are not showing the vote option, in case we have one.

Where should I add it?

Right now this is how it looks: 

![image](https://user-images.githubusercontent.com/15069783/76133553-6c7e5180-5ff7-11ea-9006-b46a375334ec.png)

linnutee:
This makes sense. Only note, should extend the blue solid only as far as the actual preference, not both agenda name + preference. This is due to the blue bg solids only being used to isolate/highlight values.

Also down the line, its probably worth to implement color coding for the consensus change votes,. I will make a separate issue for that.

PR #2440, opened 2020-03-10
My tickets view update

vctt94:
This PR is a work towards: #2262.

I am not closing the issue with this PR, because right now the vote choice is not being showed at this new `my tickets` page.


vctt94:
Can you give it another try?

Issue #2439, opened 2020-03-07
[discussion] CSS modules


amassarwi:
[CSS modules](https://create-react-app.dev/docs/adding-a-css-modules-stylesheet/) are available and used in modern react apps in order to achieve styling encapsulation.

CSS modules will help us maintain very small styling files - where each component gets it's dedicated module/file which ensure no styling clashes cross the app which means no need to come up with very long unmaintainable class names everywhere.

matheusd:
If we're going this route, it's probably best to do a small component first so we can check out what the code layout would look like.

amassarwi:
@matheusd agreed - I'll be doing some POC work and update small part for #2438 per @alexlyp  request, I thought maybe I could also migrate the styling to use CSS modules on the way.

what do you think ?

matheusd:
Yeah, a poc for both is the way to go.

amassarwi:
We have also stuff like:
```
 const className = [
      "trezor-passphrase-modal",
      this.props.isGetStarted ? "get-started" : ""
    ].join(" ");
```

Which can be replaced with: `classNames(classNames(class1 , conditioned class2, ...))` and can be done while migrating `.less` files to css modules

Issue #2438, opened 2020-03-07
[discussion] Functional components & hooks


amassarwi:
I've discussed privately with @vctt94 on matrix, and we thought opening a discussion issue would be the best way to go here, to get more opinions about it.

We would like to use functional components & hooks in decrediton to improve maintainability - many class components life-cycle hooks will be deprecated soon and functional components is the way in modern react apps.

Of course such change will lead a huge *refactor* but of course if can be done incrementally (e.g page by page).

If we get green light to implement this I'd love to contribute to the efforts to make it happen.

vctt94:
I believe it is a good step updating our code basis towards react last updates. As we can see react is clearly going through this direction and each time working less with classes. 

The good part is that we don't need a rush to update everything at once, so I believe we can start updating some small components and we can see how it goes. 

But I'd rather wait and see @alexlyp and @matheusd 's opinions, before starting any work

matheusd:
Seems reasonable. A quick spot check shows that most class-based components use it just to handle internal state changes, so it shouldn't be too bad.

Issue #2428, opened 2020-02-28
[redesign] Eligible tickets section in proposal details page


amassarwi:
See screenshot:
![image](https://user-images.githubusercontent.com/10324528/75542746-b0ce8800-5a20-11ea-9e8d-b1660ced9805.png)

**Update:** 
We ended up getting new design for the eligible tickets section, see comment below: https://github.com/decred/decrediton/issues/2428#issuecomment-606018117

linnutee:
Following up on this one, there's a few issues to consider in the layout. 

1. Wallets Eligible Tickets seems a new component, bit out of sync with the rest of the common components. I've created an example using a accordion within a separated row. This gives it a clearer separation and a bit more harmonious fit. 

Also the copy "Wallet Eligible Tickets" is a bit odd, would propose something in lines of `My Tickets, Eligible for Voting: <ticket-sum> total`
or `Tickets Eligible for Voting: <ticket-sum> tickets in wallet` for example.

If you could also include an example of how the tickets look like when expanded, i'll provide feedback and updated example if needed. 

2. "See proposal comments" on Politeia buttons placement. The problem with the text running too close to the button comes from the text being in full width of the container. The proposal contents text area has actually been in much smaller width in the UI designs, due to the reasons of legibility. Around around 50-70 characters per row makes for an optimal length, anything longer increases the reading difficulty. I assume this was made full length to decrease the vertical height of the page as the proposals are usually quite long. 

So this bit can be addressed by limiting the text area just about 10px smaller in width off the left side of the button. I've reduced the line height quite a bit down to 15px, this should make the text vertically more compact. 

Specs/preview: https://xd.adobe.com/view/cb98a967-40d1-4dc6-568f-90fccf53d9dd-bda9/

amassarwi:
Thanks for the specs @linnutee, looks clean.
Regarding the expanded mode. I've this cropped screeshot of the expanded tickets list not sure if it's good enough:
![image](https://user-images.githubusercontent.com/10324528/77005694-873eb780-6961-11ea-8013-3e239ce013f2.png)
AFAIR @vctt94 has testnet wallet with many voting tickets and may be able to share a better screenshot for the expanded mode.

vctt94:
This is how it looks closed:

![image](https://user-images.githubusercontent.com/15069783/77021404-56f91800-6965-11ea-901b-e5cdb423ac69.png)

This is how it looks openned: 
[...]

PR #2457, opened 2020-04-10
pi-ui POC & css module introduction

amassarwi:
We have decided to try integrate [pi-ui UI library](https://github.com/decred/pi-ui) at this stage, which should improve the molecularity and help us keep a consistent look & feel across decred's different products.

The following `pi-ui` related changes were done:
 -  use `ThemeProvider` to specify the themes & fonts (deleted `themes.less`).
 -  update getting started & settings page to toggle theme using pi-ui's useTheme hook.
 -   use `pi-ui`'s `Button` & `Link` in proposals list & proposal details pages.

The following changes can be done next in pi-ui's context:
 - delete icons.less and provide it using pi-ui's themes object (should be minor, see `styles/themes`)
 - use status bar component for voting progress  
 - use radio button component for voting preference
 - use pi-ui's tabs for governance page tabs 

This diff includes some work towards #2439 by introducing css modules - proposal details page was migrated & proposals list page was partly migrated, we should continue with incremental approach here.
It also includes work from #2438 - hooks polish in proposal details page/components, custom hooks, usage of `useMemo` & `useCallback` hooks, separated files for all helper functional components.

The rest of the governance pages can be refactored next to use pi-ui's components, css modules, hooks & functional components.


app/components/views/LNPage/index.js
```diff
@@ -27,4 +27,5 @@ const LNActivePage = () =>
 const LNPage = ({ lnActive }) =>
   lnActive ? <LNActivePage /> : <LNConnectPage />;
 
+
```
vctt94, line 30: unecessary change

app/components/shared/ExternalLink.js
```diff
@@ -1,15 +1,20 @@
...
 const clicker = (isTestNet, href, hrefTestNet) => () => {
   const url = hrefTestNet && isTestNet ? hrefTestNet : href;
   shell.openExternal(url);
 };
 
-const ExternalLink = ({ className, href, children, hrefTestNet, isTestNet }) => (
-  <a className={className} onClick={clicker(isTestNet, href, hrefTestNet) } href="#">
-    {children}
-  </a>
+const ExternalLink = ({ className, href, children, hrefTestNet, isTestNet, isButton }) => (
```
vctt94, line 10: I believe we can have only one tag here. No need to have the option to this component being a button or a link.

Just make sure it work in all places we use it (it should I believe)

amassarwi: It needed here in order to supper normal links (e.g the example in the following screenshot) - we could make use pi-ui's link here as well tho - I will give a try.

image

tiagoalvesdulce: Here I'll suggest using the Link component from pi-ui. It does have a customComponent property where you can pass the Button component e.g `<Link href="#" customComponent={() => <Button>Click me</Button>} />`. I added it to the docs here: https://github.com/decred/pi-ui/pull/250

app/components/shared/ExternalLink.js
```diff
@@ -1,15 +1,20 @@
...
+  isButton ? (
+    <Button size="sm" className={className} onClick={clicker(isTestNet, href, hrefTestNet) } href="#">
```
tiagoalvesdulce, line 12: It doesn't make sense to pass an href prop to the Button component.

app/components/shared/ExternalLink.js
tiagoalvesdulce: With the above comment, you can probably remove the `isButton` flag. I'm not a big fan of the `isSomething` flag when it's not state related. I'm making an effort to remove them from pgui as well. You could use the same behavior as the Link component and pass a render component to ExternalLink instead. if the component is used multiple times you can create a ExternalLinkAsButton component and it will be basically `<ExternalLink renderComponent={() => <Button ... />}/>`. IMO the code becomes more declarative this way.

app/components/views/ProposalDetails/ChooseVoteOption.jsx
```diff
@@ -76,11 +80,12 @@ function ChooseVoteOption({
   });
 
   const error = state && state.context && getError(state.context.error);
+  // TODO: use memo here! currently this is being recaclulated on each new render!
```
tiagoalvesdulce, line 83: 👍

app/components/views/ProposalDetails/ProposalDetails.jsx
```diff
@@ -0,0 +1,123 @@
...
+      if (!hasEligibleTickets) {
+        return <NoElligibleTicketsVotingInfo {...{ showPurchaseTicketsPage }} />;
+      }
+      return <ChooseVoteOption {...{
+        viewedProposalDetails, voteOptions, setVoteOption, newVoteChoice,
+        eligibleTicketCount, currentVoteChoice, votingComplete: false
+      }} />;
+    }
+    return <ProposalNotVoting />;
+  }, [ proposalStatus, voteStatus, hasTickets, hasEligibleTickets, currentVoteChoice, viewedProposalDetails, eligibleTicketCount, newVoteChoice, setVoteOption, showPurchaseTicketsPage, voteOptions ]);
```
tiagoalvesdulce, line 49: Isn't this line too long? Also, good job memoizing this but I'd rather split it into another component and use the `React.memo()` instead. This component is quite polluted by JSX already.

app/style/Icons.less
```diff
@@ -2,6 +2,9 @@
 @icon-root: '../style/icons';
 @animation-root: '../style/animations';
 
+
+// TODO: load these as theme variables!
```
tiagoalvesdulce, line 6: 👍

tiagoalvesdulce:
The order of my comments was messed up somehow. You can send me a DM on matrix if things are unclear.

amassarwi:
@vctt94 @tiagoalvesdulce thanks for the review guys, will be addressed.

app/components/views/ProposalDetails/ProposalDetails.jsx
amassarwi: Good point mate, I've considered splitting the component here, thought maybe doing it in another PR but actually it make sense to do it at this stage when the css module are there.
I can give it a shot.

matheusd:
My only comment is to be careful in changing the components visuals since it affects the UI as designed by @linnutee and co. Probably best to include some screenshots of the wip.

amassarwi:
@matheusd Sure, we have the specs in mind.
Actually this PR was suppose to include the new eligible tickets components provided here #2428.
We decided to split it and add another PR with the new component.

.github/workflows/nodejs.yml
```diff
@@ -15,7 +15,9 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
 
-    - name: yarn
+    - name: yarn install 1.21.1
```
amassarwi, line 18: This isn't valid and will reverted, was just a try to resolve the build issue!

amassarwi:
> Right now it is possible to change vote option in a finished proposal.

@vctt94: I was able to reproduce the same behavior on master!
I will include the fix for that as well as part of this PR.

app/components/shared/ExternalLink.js
vctt94: Great. I am not really a fan of overcomplicating components as well. Glad you pointed the directions so amass can deal with it :)

app/components/views/GovernancePage/Proposals/index.js
```diff
@@ -54,15 +54,16 @@ function Proposals() {
...
 
   // TODO move reducers which only control local states from reducer/governance.js
   // to here.
   const [ tab, setTab ] = useReducer(() => getProposalsTab(location));
 
-  const dispatch = useDispatch();
   useEffect(() => {
     dispatch(setLastPoliteiaAccessTime());
-  }, []);
+  }, [ dispatch ]);
```
vctt94, line 65: Why you add this dispatch here? I left it empty to be used as a "componentDidMount"

app/components/layout/TabbedPage.js
```diff
@@ -103,7 +103,7 @@ function TabbedPage(props) {
   const [ dir, setDir ] = useState("l2r");
   useEffect( () => {
     setMatchedTab(getMatchedTab(location, props.children));
-  }, []);
+  }, [ props.children, location ]);
```
vctt94, line 106: same here

app/components/layout/TabbedPage.js
```diff
@@ -116,7 +116,7 @@ function TabbedPage(props) {
       : "l2r";
     setDir(dir);
     setMatchedTab(matchedTab);
-  }, [ location ]);
+  }, [ location, previous, props ]);
```
vctt94, line 119: I don't think we need previous here, as if previous values are changing, it means that current values changed

app/components/views/GovernancePage/Proposals/index.js
tiagoalvesdulce: It's safer and obeys the `exhaustive-deps` rule. You can read more about it [here](https://github.com/facebook/react/issues/14920). The dispatch function is created outside the component context so it shouldn't change. It should work exactly the same.

app/components/layout/TabbedPage.js
tiagoalvesdulce: Here is more case dependant, if the deps break the code and you have a really good case for keeping the dep list empty it's worth disabling the lint rule inline with `// eslint-disable-next-line react-hooks/exhaustive-deps`. Sometimes is misleading disabling it, tho. Take a look at the links in [this comment](https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528).

app/components/shared/ExternalLink.js
vctt94: @amassarwi can you add tiago's comment?

amassarwi: Sure, will do!

app/components/views/GovernancePage/Proposals/index.js
amassarwi: @tiagoalvesdulce thanks for reading my mind 👍

app/components/layout/TabbedPage.js
amassarwi: I think we shouldn't disable the rule, I think we either should specify array with *all* deps, or with no array => will run one time like componentdidmount AFAIR.

amassarwi: Here are some insights why we should be strict with the deps rule everywhere:
https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies
https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often

TLDR: we should break the useEffect to several ones instead of omitting deps, see second link above for more insights.

amassarwi: Also, we shouldn't use deep properties like `props.children` in the deps array, instead we should always define another const outside the useEffect and use it instead inside the `useEffect` and in the deps array!

tiagoalvesdulce: Yeah, the useEffect hook should be used for a specific task and you can have multiple useEffect hooks. Even better if you have custom hooks with descriptive names for each of these tasks if they are related or just too long.

app/components/views/GovernancePage/Proposals/index.js
tiagoalvesdulce: 😂😂😂

amassarwi:
@vctt94 

> Right now it is possible to change vote option in a finished proposal.

Was fixed on my latest commit, could you please give it a try on your end.

app/components/views/ProposalDetails/ProposalDetails.jsx
amassarwi: Done mate :mate:

amassarwi:
@vctt94  

> I have found another issue which also happening on master, currently when voting on a proposal successfully the cast vote button is still enabled and isn't turning to disabled, I think we should fix both in this diff as well.

^ Addressed :heavy_check_mark:

amassarwi:
@vctt94 
Just found another issue which I was able to reproduce on master that I would like to address here as well, `purchase tickets` button isn't navigating to the tickets page, see screenshot bellow:
image

app/style/Icons.less
amassarwi: Looks like adding this change to this PR would make extraordinarily big, will create an issue for that and I would like to work on it after merging this one!

app/components/shared/ExternalLink.js
amassarwi: Done, Instead of changing `ExternalLink` which is used in many other parts of the APP, I have refactored `PoliteiaLink` to not rely on `useExternalLink`, and to use combination of `pi-ui`'s `Link` & `Component` depends on the usage.

app/components/shared/ExternalLink.js
```diff
@@ -2,6 +2,7 @@ import { shell } from "electron";
 import { network } from "connectors";
 
 const clicker = (isTestNet, href, hrefTestNet) => () => {
+  console.log(isTestNet);
```
amassarwi, line 5: remove log!

app/components/layout/TabbedPage.js
```diff
@@ -116,8 +117,7 @@ function TabbedPage(props) {
       : "l2r";
     setDir(dir);
     setMatchedTab(matchedTab);
-  }, [ location ]);
-  let { children, header, className } = props;
+  }, [ location, previous, children, onChange ]);
```
vctt94, line 120: I removed previous from here and felt no side effects.

Issue #2131, opened 2019-06-03
[Governance] Improve error display when failing to fetch proposals


vctt94:
When going to governance page, if decrediton fails to fetch proposals, it shows as there is no proposals available. We should show that an error happened instead, to better inform the user what happened.

![image](https://user-images.githubusercontent.com/15069783/58806602-ec9d0400-85ec-11e9-9d0b-e338f157195d.png)

s-ben:
Hey @vctt94, is this issue still valid? Looking to [tweet out good first issues](https://twitter.com/dcrgoodfirst) and this look promising.

vctt94:
Hey @s-ben, yeah it still an issue!

guilhermemntt:
Fixing the issue. I'd like to collaborate, could someone assign this issue to me? Thanks!

PR #2473, opened 2020-04-27
Add snackbar error modal on fetching proposals fail

guilhermemntt:
Closes https://github.com/decred/decrediton/issues/2131

Proposals are cached after the first fetch, so it'll not work properly if the proposals has already been fetched once. Is this the desired behaviour?


matheusd:
Hello @guilhermemntt, thank you for your first contribution!

This is on the right track since this now correctly shows a snackbar error on updates (vs silently failing).

But I believe the idea for the original issue is to display a persistent error message (instead of the erroneous "no proposals available") and possibly also a button to force trying to fetch the proposals again.

guilhermemntt:
I really didin't like the blue button, so I'm thinking about changing it's color to gray.
Should I remove the snackbar?


image

vctt94:
Usually gray is used for disabled state at decrediton, so it may be missinterpreted. I think it is okay with the blue button and the snackbar. 

I think you can also add an `error` class at the text, to reinforce the idea 

We can use some design on this later from @linnutee

app/style/Governance.less
```diff
@@ -52,16 +52,42 @@
   padding: 1em;
 }
 
+.loading-error {
```
amassarwi, line 55: The governance pages are being refactored ATM to use css module instead of `.less`.
So it would be great if the styling of LoadingError component will be in it's own css module.

We have two option:
1. Create components/shared/ErrorLoading.module.css 
2. Create one css module for all shared components for now instead of css module for each component, then you could create `components/shared/shared.module.css`

After adding the module you could import it in your component and style your component - have a look at ProposalDetails.module.css on master.

app/components/shared/LoadingError.js
```diff
@@ -0,0 +1,32 @@
+import { FormattedMessage as T } from "react-intl";
```
amassarwi, line 1: Can we change this file to be `.jsx` ?
I know all other components are with `.js` but we are trying to move away from this and keep components in `.jsx` files and pure JS in `.js` files.

app/style/Governance.less
amassarwi: TBH I'm more fan of the second option here, since all shared components are small enough I guess one module should be enough here.
In case we see in later stage that one component styling is growing to big, we could easily then add a separate css module for it's own.

guilhermemntt: @amassarwi I didn't see your comment before my last commits, so please ignore it. Unfortunately, I had to rebase my branch to the most recent master updates, basically my last commits are the fixes I've done so far. I have two others pull requests opened here, so probably I'll rebase it too.

app/components/shared/LoadingError.js
guilhermemntt: Of course, I really miss Typescript.

app/style/Governance.less
guilhermemntt: Regarding the ```.less``` modules, I totally agree. I'll push the corrections soon.

guilhermemntt:
Implemented in proposals list and proposal detail views.

alexlyp:
Not sure what happened here, but a couple of rogue files were added that need to be removed.

And please resolve conflicts on this and the other PRs that you currently have open.

guilhermemntt:
Rebased.

guilhermemntt:
Fixed. I've added a cancel button to return to previous list. Now, the list keeps its loaded proposals even if the tab changes.

Issue #2454, opened 2020-04-07
[purchase ticket] Add Total Spend 


vctt94:
When purchasing tickets, if buying more than one, decrediton will not show the total being spend. So if someone wants to know how many decreds will be left, after purchase you need to it manually.

We can show the total being spending when purchasing more than one ticket

guilhermemntt:
Currently working on it

PR #2474, opened 2020-04-27
Purchase ticket additional information

guilhermemntt:
Closes https://github.com/decred/decrediton/issues/2454


app/components/inputs/Input.js
```diff
@@ -97,6 +98,13 @@ class Input extends React.Component{
               ) : null}
             </div>
           ) : null}
+          {!invalid && value && validMessage ? (
+            <div className="input-valid-message-area">
+              <div>
```
matheusd, line 103: No need for this extra `div` here. `{validMessage}` can wrap itself however it wants and the previous `<div className="input-valid-message-area">` already provides separation.

app/components/views/TicketsPage/PurchaseTab/PurchaseTickets/Form.js
```diff
@@ -1,12 +1,21 @@
...
 import "style/StakePool.less";
 
 const purchaseLabel = () => <T id="purchaseTickets.purchaseBtn" m="Purchase" />;
 
+const PurchaseValidMessage = ({ spendable, numTicketsToBuy, ticketPrice }) => (
+  <div className="purchase-ticket-valid-message-area">
+    <T id="purchaseTickets.output" m="Output:" />
+    <Balance flat amount={numTicketsToBuy * ticketPrice} />
+    <T id="purchaseTickets.remaining" m="Remaining:" />
+    <Balance flat amount={spendable - numTicketsToBuy * ticketPrice} />
```
matheusd, line 15: Don't write interpolated messages like this, since different languages might want to switch the order of elements (please take a quick look at https://github.com/decred/decrediton/tree/master/app/i18n). If the messages can appear on a single line, write the following:

```js
<T 
    id="purchaseTickets.validMsg" 
    m="Output: {amount} Remaining: {remaining}" 
    values={{
         amount: <Balance flat amount={numTicketsToBuy*ticketPrice} />, 
         remaining: <Balance flat amount={spendable - numTicketsToBuy*ticketPrice} />
    }} 
/>
```

Otherwise if the design calls for two lines (one for "Output:", one for "Remaining:"), then break them into two different messages (but still using interpolated values in the message itself, instead of two different components) and style them directly with CSS (such as by adding a `purchase-ticket-valid-message-area > span` rule).

vctt94: Also I think `total` would be more appropriate than output.

app/style/PurchaseTickets.less
```diff
@@ -57,24 +57,24 @@
   background-size: 6px;
   background-position: 50% 50%;
   background-repeat: no-repeat;
-  background-color: #8997A5;
+  background-color: #8997a5;
```
vctt94, line 60: We have this color at our themes. `Themes.less`

app/components/inputs/Input.js
```diff
@@ -97,6 +98,13 @@ class Input extends React.Component{
               ) : null}
             </div>
           ) : null}
+          {!invalid && value && validMessage ? (
```
vctt94, line 101: I'd rather this message to be added at the `PurchaseTickets/Form.js`. 

I am not really a fan of overcomplicating components and as `validMessage` is only used at this place, I think it is worth to not add it to `Input.js` component.

guilhermemntt:
Just pushed the requested changes. Thank you!

Issue #2455, opened 2020-04-07
[cspp] Integrate cspp into decrediton


vctt94:
This is a list of what is needed so far for finishing the cspp integration with decrediton.

- [ ] Add Privacy Page - only allow to see if changed in settings
- [ ] Add Privacy tutorial for beginners
- [ ] Add method to create necessary initial accounts
- [ ] Add indicator mixer is running
- [ ] Flag wallet as mixed, which will have different behaviors such as: 
    - [ ] Prevent receiving funds on mixing account
    - [ ] Prevent spending from unmixed account
- [ ] Allow to restore private wallet

This was discussed at matrix but still open for debate, if someone has a different opinion, please let us know.

matheusd:
Some ideas, in no particular order.

- [ ] Flag mix transactions as such (ideally this would be done by dcrwallet itself since it knows which transactions were done in conjunction with the cspp server but we can use some heuristics to do that ourselves)
- [ ] Improve txdetails ui for mix transactions (too many addresses clutter the ui, there are some visual issues in the page when showing mix transactions)
- [ ] Have some CSPP-specific stats/logs (how many mixes we participated in, existing utxos)

xaur:
> - Prevent receiving funds on mixing account  
> - Prevent spending from unmixed account

It would be nice to have these protections in dcrwallet and reused from there to avoid duplication.

PR #2475, opened 2020-04-27
[cspp] integrating cspp into decrediton

vctt94:
This PR is a continuation of the integration of cspp at decrediton. #2455 

-  It makes `mixedAccounts` not being able to receive funds (not showing addresses at receive page);
- It makes only possible to send funds from `mixedAccount` at send page; 
- It adds a `isMixTx` method which will help us determine which tx is mixed or not, so we will be able to flag them

    This way we will be able to restore private wallets and guess which account is the mixed account. 

    This code is based on https://github.com/decred/dcrdata/blob/master/txhelpers/cspp.go (thanks @chappjc ).

I think it would be good to add the ability to send funds from other accounts in case the user want to (mark a checkbox, which notifies him about the lack of privacy or the possibility of reducing the anonimous set where these inputs may be spent). 

After this being merged, we will still need to:
 - restore private wallets;
 - Add tutorial for begginers;
 - styling


app/constants/Decred.js
```diff
@@ -45,3 +45,44 @@ export const MAX_DCR_AMOUNT = 21e14;
...
+//   - 72 bytes DER signature + 1 byte sighash
+//   - OP_DATA_33
+//   - 33 bytes serialized compressed pubkey
+export const	RedeemP2PKHSigScriptSize = 1 + 73 + 1 + 33;
+
+// AtomsPerCoin is the number of atomic units in one coin.
+const AtomsPerCoin = 1e8;
+
+// MaxAmount is the maximum transaction amount allowed in atoms.
+// Decred - Changeme for release
```
matheusd, line 63: No need for this comment :P

app/helpers/transactions.js
```diff
@@ -0,0 +1,147 @@
...
+        mixedOuts.set(val, 1);
+      }
+    }
+  });
+
+  mixedOuts.forEach( (count, val) => {
+    console.log(val);
+    console.log(count);
+    console.log(mixedOuts);
+    console.log(mixedOuts.get(val));
```
matheusd: Extraneaous logs

matheusd: Actually I see you've removed on a later commit (was reviewing by commit)

alexlyp:
Needs a rebase, then should be good to go.

PR #2491, opened 2020-05-21
[LN] Allow multiple in-flight payments

matheusd:
LN payments aren't guaranteed to be completed quickly so we need to allow
multiple outstanding payments.

This doesn't yet handle the case of in-flight payments across re-starts (this
will be handled in a future PR).


matheusd:
Updated to fix various issues found while testing with @vctt94

PR #2523, opened 2020-06-10
Decode sstxcommitment script

vctt94:
After merging https://github.com/decred/decrediton/pull/2468, we removed our grpc call to decodeRawTransactions. 

When working at https://github.com/decred/decrediton/pull/2516, I noticed we are not properly decoding sstxcommitment scripts, which are needed for signing the tx message for the vsp.

This PR adds it.

### Diff description: 

 - I added the opcode constants file, which have all opcodes described on the btc wiki, in bitcoin core and in most if not all other references and software related  to handling DCR scripts.
 - at our `helpers/script.js` I am adding the `parseScript` method, which was removed from dcrd due to optmizations at the txscript on https://github.com/decred/dcrd/pull/1656.
   - The reason I am doing that is for convinience. This way we need to import less code from dcrd into decrediton, and as we do not decode big scripts in decrediton, it is fine for us, for now.
  - added `addrFromSStxPkScrCommitment` so we can get addressses from sstxcommitment scripts
  - fixed some of our tests after the new additions, move some test data to different folders, for better naming and increased out test coveraged. 
    - I added a new `wallet/service.spec.js` which is a integration test of our `decodeRawTransaction` methods from wallet service. It is in unit tests because right now our package.json consider our unit/test folder as test's rootDir. We need to add integration tests to it.


test/data/decodedTransactions.js
```diff
@@ -0,0 +1,100 @@
+export const decodedPurchasedTicketTx = {
+  "version": 1,
+  "serType": 0,
+  "numInputs": 1,
+  "inputs": [
+    {
+      "opRawHash": Buffer.from([132, 107, 192, 162, 131, 217, 8, 228, 140, 36, 137, 147, 128, 220, 21, 80, 233, 149, 72, 230, 4, 157, 215, 148, 83, 194, 10, 148, 245, 249, 167, 105]),
```
matheusd, line 7: I'd recommend creating a helper to convert from an hex string to a Buffer for these functions, since it makes it easier to review.

Or maybe dropping `opRawHash` altogether and working off of `prevTxId` (since having two fields for the same info could lead to using inconsistent data)

vctt94: Agree. I removed the opRawHash from decodedTransaction object.

PR #2532, opened 2020-06-15
Fix multi tx bugs at transaction detail page

### LINKED ISSUE #2522: [tickets] some voted tickets not showing `reward` & `vote time` info
image

vctt94:
This PR depends on #2523.

It fixes #2522. The problem was that as we were using the tx status to show it, the status of a purchased ticket was voted, showing as voted. Now it will show as purchased, but the icon will change based on the status. Here is an example with a live, voted and revoked ticket: 

image

I am fixing the vote outputs  which @matheusd pointed when merging #2468. Now I am showing the script: 

image

I am adding the asm to decoded scripts, but right now I am not fully showing it at the scripts, but I belive we can. Toughts on that?

 I am fixing a bug which it tries to get input amount from received transaction (which the wallet don't recognize as it is not a wallet input), causing it to fail.

But still not showing the amount on received txs: 

image


app/helpers/scripts.js
```diff
@@ -258,14 +261,36 @@ export const extractPkScriptAddrs = (version, pkScript, chainParams) => {
...
+    for (let i = 0; i < retScript.length; i++) {
+      disbuf += printPop(retScript[i]);
+      disbuf += " ";
+    }
+    // remove last space (" ").
+    if (disbuf.length > 0) {
+      disbuf = disbuf.slice(0, -1);
+    }
+  }
+  // console.log(pkScript)
```
amassarwi, line 280: cleanup

app/components/views/TransactionPage/Page.js
```diff
@@ -15,7 +15,7 @@ import * as sel from "selectors";
 import "style/TxDetails.less";
 
 function mapNonWalletOutput(output) {
-  const address = output.decodedScript.address || "[script]";
+  const address = output.decodedScript.address || "[script] - " + output.decodedScript.asm;
```
amassarwi, line 18: string literal

test/data/script.js
```diff
@@ -42,18 +42,40 @@ export const scriptDataTest = [
...
+    expected: {
+      "address": "DcgYx6SzsWsaTFYEHwZ83wyKntCMiJYrJ3M",
+      "requiredSig": 1,
+      "scriptClass": 3,
+      "asm": "OP_HASH160 OP_DATA_20 63bcc565f9e68ee0189dd5cc67f1b0e5f02f45cb OP_EQUAL"
+    },
     // reqSigs: 1,
     // class: ScriptHashTy == 3,
   },
+  // {"version":1,"serType":0,"numInputs":1,"inputs":
```
amassarwi, line 67: intended ?

matheusd:
> I am adding the asm to decoded scripts, but right now I am not fully showing it at the scripts, but I belive we can. Toughts on that?

I think it would be better for ticket transactions to show the address (and maybe a flag that it's a ticket commiment?) since that seems more useful to me (for example, to allow a user to verify the commitment address belongs to the wallet).

For unrecognized transactions or other types of outputs, then showing the script seems more reasonable.

One thing I think we need is to add the full address/script to a an alt text and maybe copy to the clipboard upon clicking the address/output so that you can actually verify it in the security settings (showing only a partial address prevents this and forces you to go to the explorer to see the full address).

vctt94:
I agree. I am adding a new method at https://github.com/decred/decrediton/pull/2577/files#diff-649256bad84fbd0681cc5737f4abb310R323 which will get sstx commitment addresses.

after merging 2577 it will be easy enough to add the commitment address at the TransactionPage view. 

So if we can get that in, I can add that in a further PR, if that's fine for you @matheusd

PR #2534, opened 2020-06-16
Fix ticket naming when saving proposal votes

vctt94:
@amassarwi reported to me a bug after voting on latest master. 

After f31cc4b768ee1ff25c5a7ff1e94e34f7bf0f834f, the ticket name was changed to `txHash` before saving it to our politeia directory:
https://github.com/decred/decrediton/commit/f31cc4b768ee1ff25c5a7ff1e94e34f7bf0f834f#diff-c8399d09f1bddd336993b884a3e8d512R623

When opening a proposal with this kind of votes, it would fail to open with the following message, as it is looking for tickets instead of txHash:

image

This PR fixes it.

### DIFF
--- a/app/components/views/ProposalDetailsPage/ProposalDetails.jsx
+++ b/app/components/views/ProposalDetailsPage/ProposalDetails.jsx
@@ -43,6 +43,11 @@ const ProposalDetails = ({
   goBackHistory,
   eligibleTicketCount
 }) => {
+  walletEligibleTickets = walletEligibleTickets.map((et, i) => {
+    walletEligibleTickets[i].txHash = et.ticket;
+    return walletEligibleTickets[i];
+  });
+
   const { tsDate, hasTickets, isTestnet } = useProposalDetails();
   const { themeName } = useTheme();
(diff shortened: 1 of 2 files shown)


vctt94:
fixed

app/components/views/ProposalDetailsPage/ProposalDetails.jsx
```diff
@@ -43,6 +43,11 @@ const ProposalDetails = ({
   goBackHistory,
   eligibleTicketCount
 }) => {
+  walletEligibleTickets = walletEligibleTickets.map((et, i) => {
```
amassarwi, line 46: I would suggest moving this to GovernanceAction.js, to keep all proposal data logic in one place.

vctt94: I thought about that. The reason I didn't do it, is to avoid saving duplicated info into our `vote.json` file

amassarwi: Oh, I see that's was the initial issue, the extra info

PR #2535, opened 2020-06-16
[LN] Add SCB backup info and restore option

matheusd:
This adds info about the SCB file that dcrlnd continously generates for backup
purposes. It also adds the possibility of generating an ad-hoc multi-channel
backup and verifying it.

During LN wallet creation, users can now specify an SCB file to apply the
backup after restoring a wallet from seed.

Closes #2527, resolves #2529

PR #2543, opened 2020-06-23
Fix ln connect button loading state

### LINKED ISSUE #2500: [LN] If connection fails, button does not reset state
If ln fails to connect, the button does not leaves the "loading" state:
image

bgptr:
closes #2500


matheusd:
Right, #2535 changed some of the startup sequence so I'd wait until that is merged and verify again whether that's still an issue.

vctt94:
I believe we can close this PR, as this error was already handled by a prior PR.

PR #2548, opened 2020-06-27
Polish tx row pending indicator

### LINKED ISSUE #2544: [overview] Pending tx indicator may be out of length
At our overview, it is possible the pending tx indicator be out of the row:

image

I believe this happens on sent transactions, which we show the address and it may occupy the whole length (this is a guess)

amassarwi:
This diff improves the pending indicator which is displayed in both txs history & overview lists.
 - Use flex to align content of the row.
 - Use pi-ui's `Tooltip`.
 - Use pi-ui's `Text` component with `truncate` prop to ellipsis too wide text. 

Closes #2544 

After:

image

image

image

### DIFF
--- a/app/components/shared/TxHistory/RegularTxRow.jsx
+++ b/app/components/shared/TxHistory/RegularTxRow.jsx
@@ -6,7 +6,7 @@ import {
   TRANSACTION_DIR_TRANSFERRED
 } from "constants";
 import styles from "./TxHistory.module.css";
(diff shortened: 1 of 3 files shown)


alexlyp:
Hrm I'm seeing animated pending icon floating to the left near the transaction instead of the right edge of the column.  Also when hovered I'm seeing "content" instead of some other information.

Issue #2519, opened 2020-06-08
[pi middleware] Promissify politeia's requests


vctt94:
Right now we are using the methods of our politeia middlware direct into our GovernanceActions.

It would be better if we promissify them, like we do with our vsp requests.

guilhermemntt:
I'm currently working on it.

PR #2557, opened 2020-07-03
Promisify politeia API requests

guilhermemntt:
Closes https://github.com/decred/decrediton/issues/2519.

### DIFF
--- a/app/middleware/politeiaapi.js
+++ b/app/middleware/politeiaapi.js
@@ -38,24 +38,64 @@ function POST(piURL, path, payload) {
   });
 }
 
-export const getActiveVotes = (piURL) => GET(piURL, "/v1/proposals/activevote");
-export const getVetted = (piURL) => GET(piURL, "/v1/proposals/vetted");
-export const getVotesStatus = (piURL) => GET(piURL, "/v1/proposals/votestatus");
-export const getProposal = (piURL, token) =>
-  GET(piURL, "/v1/proposals/" + token);
-export const getProposalVotes = (piURL, token) =>
-  GET(piURL, "/v1/proposals/" + token + "/votes");
-export const getProposalVoteStatus = (piURL, token) =>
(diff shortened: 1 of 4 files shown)


app/middleware/politeiaapi.js
```diff
@@ -38,24 +38,64 @@ function POST(piURL, path, payload) {
...
-export const getVotesStatus = (piURL) => GET(piURL, "/v1/proposals/votestatus");
-export const getProposal = (piURL, token) =>
-  GET(piURL, "/v1/proposals/" + token);
-export const getProposalVotes = (piURL, token) =>
-  GET(piURL, "/v1/proposals/" + token + "/votes");
-export const getProposalVoteStatus = (piURL, token) =>
-  GET(piURL, "/v1/proposals/" + token + "/votestatus");
-export const getTokenInventory = (piURL) =>
-  GET(piURL, "/v1/proposals/tokeninventory");
+export const getActiveVotes = ({ piURL }, cb) => GET(piURL, "/v1/proposals/activevote").then(function (response) {
```
matheusd, line 41: Isn't the return of `GET` already a promise? Seems redundant to me to switch to callback style here and then back to promises at the wallet level.

matheusd: Also, I somewhat dislike changing parameters from a simple list to an object since it makes figuring out the required parameters harder. This is usually reserved for situations where the parameter list is large or changing very rapidly.

Any particular reason you're switching the parameters to this style here?

guilhermemntt: You're totally rigth, ``axios.get`` already returns a ``Promise``. I've just followed the pattern defined in https://github.com/decred/decrediton/pull/1536. By doing it this way, we can take advantage of ``log`` function as well.

matheusd: Right, as we discussed in a DM, given the previous work was also done in this way I think it's reasonable to keep following it.

PR #2560, opened 2020-07-03
Deal with error in case no tx details is found

vctt94:
I noticed if a wallet does not know some tx details it will crash decrediton when starting. This may happen for many reasons, one we discussed at matrix is that, mixer is set to ignore gap limit. So if someone mixes his outputs with a different wallet.db, decrediton may not recognize it. 

This PR deals with this case, so decrediton can still starts and we can further investigate the cause.

### DIFF
--- a/app/selectors.js
+++ b/app/selectors.js
@@ -421,8 +421,16 @@ export const ticketNormalizer = createSelector(
         .reduce((s, c) => (s + isTicketChange(c) ? c.getAmount() : 0), 0);
 
       // ticket investment is the full amount paid by the wallet on the ticket purchase
+      let accountName = "";
       const debitList = ticketTx.getDebitsList();
-      const accountName = getAccountName(debitList[0].getPreviousAccount());
+      if (debitList.length > 0) {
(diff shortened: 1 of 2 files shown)


app/selectors.js
```diff
@@ -421,8 +421,16 @@ export const ticketNormalizer = createSelector(
...
+      let accountName = "";
       const debitList = ticketTx.getDebitsList();
-      const accountName = getAccountName(debitList[0].getPreviousAccount());
+      if (debitList.length > 0) {
+        accountName = getAccountName(debitList[0].getPreviousAccount());
+      } else {
+        // Mixer is set to ignore gap limit. If a wallet does not know tx
+        // details for some reason, can be due that or some other reason which
+        // needs to be investigated.
+        throw "Tx debit list is empty. Something is wrong.";
```
matheusd, line 432: Other situations where this (debitList.length == 0) might happen: pool fee wallet (where the wallet records the ticket but the pool did not in fact have any debits), ticket being purchased from a different wallet into the new one (not sure if applicable currently)

Issue #2566, opened 2020-07-06
[txHistory] vote info is showing enter timestamp


vctt94:
Right now at our overview or my tickets page, we are showing the enter timestamp (which is the date the ticket was purchased). We should show the leave timestamp (the date the vote was voted) 

example: (this ticket was voted today and it is showing may 20:

![image](https://user-images.githubusercontent.com/15069783/86651624-0f50d780-bfba-11ea-94a5-21e43ba006c5.png)

guilhermemntt:
On it.

PR #2569, opened 2020-07-09
Vote info showing leave timestamp

guilhermemntt:
Changes:

- Overview newly tickets is now showing the leave timestamp.
- A new column "Refunded" has been added to "My Tickets" page.

Closes https://github.com/decred/decrediton/issues/2566.

image

### DIFF
--- a/app/components/shared/TxHistory/StakeTxRow.jsx
+++ b/app/components/shared/TxHistory/StakeTxRow.jsx
@@ -17,7 +17,6 @@ const StakeTxRow = ({
   leaveTimestamp,
   enterTimestamp,
   pending,
-  tsDate,
   txTs,
   accountName,
   txType,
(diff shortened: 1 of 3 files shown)


guilhermemntt:
Maybe @linnutee can help us.

linnutee:
Hi sorry for taking a while. 

The refund column doesn't seem like a good addition since it only applies Revoked tickets, thus eating up space for most of the time. 

The scenario here around the Revoked tickets comes down to users not knowing the technical detail around 256 block case, as i've noticed this is somewhat common question "where are my funds?" after revocation. 

Off the bat: I would first try to communicate the revoking time matter in the individual ticket view – since there's the most real-estate to work with. Alternatively, we could have a custom component in the "Vote Time" column, that instead of the clock icon has something referring to revocation, as well include a tooltip for that. I'll shoot something over during this week. 

Also worth to double check the type sizes (table headers) and the component on Account column seems to have spacing issues; as well Reward row on Revoked has "--" infront of 0.00

guilhermemntt:
I'm closing this issue by now. https://github.com/decred/decrediton/pull/2569#issuecomment-662511564 is still valid, but discussing with @vctt94, it seems better to implement this changes at the next release. Note that some changes in this direction have been made in https://github.com/decred/decrediton/pull/2819, considering the new VSP approach.

PR #2575, opened 2020-07-10
[LN] Show basic network info

matheusd:
This adds basic network info plus the ability to query for specific node info
from the local dcrlnd node.

This should help debug split views on the network.


matheusd:
Updated to handle search errors

PR #2578, opened 2020-07-14
Fix startup error on macos

amassarwi:
I was experiencing the following error after pulling latest master with electron `v8.2.*`:
https://github.com/electron/electron/issues/22023#issuecomment-652338308 
Which leads to: https://www.electronjs.org/docs/tutorial/using-native-node-modules#using-npm

Running the suggested commands resulted in some yarn.lock changes, see below commands:
```
$ export npm_config_target=1.2.3                                    [23:36:39]
# The architecture of Electron, see https://electronjs.org/docs/tutorial/support\#supported-platforms
# for supported architectures.
export npm_config_arch=x64
export npm_config_target_arch=x64
# Download headers for Electron.
export npm_config_disturl=https://electronjs.org/headers
# Tell node-pre-gyp that we are building for Electron.
export npm_config_runtime=electron
# Tell node-pre-gyp to build module from source code.
export npm_config_build_from_source=true
# Install all dependencies, and store cache to ~/.electron-gyp.
sudo = HOME=~/.electron-gyp yarn
```

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12,581 +12,467 @@
   resolved "https://registry.yarnpkg.com/7zip/-/7zip-0.0.6.tgz#9cafb171af82329490353b4816f03347aa150a30"
   integrity sha1-nK+xca+CMpSQNTtIFvAzR6oVCjA=
 
-"@babel/code-frame@^7.0.0":
-  version "7.10.1"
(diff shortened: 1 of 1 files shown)


amassarwi:
After long fight with yarn, node & the gang, I ended up upgrading my node
to 14.* and recloning the repo, after that running latest master worked just fine.

Issue #2552, opened 2020-06-29
"Shutting down decredtion", advanced startup & settings ui issues


amassarwi:
Currently animation & message aren't centered properly.
I guess this was introduced with #2537.

amassarwi:
Also related to #2537 and reported by @vctt94 on matrix - advanced startup layout is broken:
![image](https://user-images.githubusercontent.com/10324528/86132343-0b6b1400-bae7-11ea-890c-cfd8f60aff5e.png)

amassarwi:
![image](https://user-images.githubusercontent.com/10324528/86479680-e4ae1700-bd4c-11ea-9cdb-0d28eec7eb22.png)

amassarwi:
@vctt94 reported another small ui issue on matrix:

> Choosing wallets is not with cursor pointer

vctt94:
@matheusd reported when decrediton starting going to tutorial page, decrediton crashes

amassarwi:
> @matheusd reported when decrediton starting going to tutorial page, decrediton crashes

on it

vctt94:
On small widths, when there is a error to start, the error goes to the bottom of the page.

![image](https://user-images.githubusercontent.com/15069783/88580886-b3aac480-d022-11ea-96cd-f8449de0883a.png)

amassarwi:
@vctt94 i think we can close this issue, as we fixed most of the listed issues and the remaining alignment issue
is being addressed by @guilhermemntt in #2639

vctt94:
yeah, sure

PR #2580, opened 2020-07-15
Shutdown: Functional component, custom hook & ui fix 

amassarwi:
This diff refactors the shutdown page to a functional component.
In addition, it replaces `shutdownpage` connector with `useShutdown`
hook.
It also fixes a ui alignment issue reported on #2552.


app/components/views/ShutdownPage/ShutdownPage.jsx
```diff
@@ -0,0 +1,24 @@
+import { FormattedMessage as T } from "react-intl";
+import { useShutdown } from "./hooks";
+import { useMountEffect } from "hooks";
+import { DecredLoading } from "indicators";
+import { classNames } from "pi-ui";
+import styles from "../GetStartedPage/GetStarted.module.css";
```
amassarwi, line 6: This isn't ideal, the following classes in GetStarted.module.css are shared between different files, 
`<div className={classNames(styles.pageBody, styles.getstarted)}>` we should wrap this div in a component 
by it's own, as it used in many places, then we will avoid importing this module here, I will keep it here for know and will open a separate issue regarding this reused div/classes.

vctt94: fine by me

amassarwi:
Linking to functional refactoring issue #2438

PR #2581, opened 2020-07-16
Fix minor bugs in modal components

guilhermemntt:
This PR fixes two minor bugs I've just checked in modal components:
1. ``PassphraseModal`` sometimes does not reset the passphrase input content.
2. ``ChangePasswordModal`` never works properly.

They were probably generated during refactoring.

### DIFF
--- a/app/components/modals/PassphraseModal/hooks.js
+++ b/app/components/modals/PassphraseModal/hooks.js
@@ -15,10 +15,6 @@ function usePassphraseModal(
     setHasFailedAttempt(false);
   }, []);
 
-  useEffect(() => {
-    if (triggerSubmit) onSubmit();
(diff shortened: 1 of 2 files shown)


app/components/modals/PassphraseModal/hooks.js
```diff
@@ -44,6 +40,10 @@ function usePassphraseModal(
     resetState();
   }, [passPhrase, validationFailed, isValidCallback, onSubmit, resetState]);
 
+  useEffect(() => {
+    if (triggerSubmit) onSubmitCallback();
```
amassarwi, line 44:
```suggestion
    triggerSubmit && onSubmitCallback()
```

PR #2589, opened 2020-07-21
Update en translations to display values

### LINKED ISSUE #2462: [1.5.1] Blank voted/earned stats on ticket page
On the "Overview" tab whilst displaying the "Tickets" chart. To the left of the chart, and under where the total balance is displayed, I am seeing:

> X active tickets
> With a total value of Y DCR
> 
> voted tickets
> Earned

The number of voted tickets and earned DCR are not populated.

jholdstock:
Attempted to fix #2462 because it is an annoyance for me personally, but as I am working on it I realise this is a part of a much wider change, and I am not totally familiar with the i18n workflow for decrediton. I have pushed what I have so far, but it is incomplete because it only fixes two of the affected fields, and only affects English language.

Even if the other fixes are not appropriate, using the correct var name in `Tickets.js` (`totalValueOfActiveTickets` => `totalValueOfLiveTickets`) could possibly be merged in isolation, 

fyi @matheusd

### DIFF
--- a/app/components/views/HomePage/tabs/Tickets.js
+++ b/app/components/views/HomePage/tabs/Tickets.js
@@ -30,7 +30,7 @@ const HomePage = ({
           <div className="is-row">
             <div className="overview-balance-spendable-locked-text">
               <T
-                id="home.totalValueOfActiveTickets"
+                id="home.totalValueOfLiveTickets"
                 m="With a total value of {value}"
                 values={{
                   value: (


matheusd:
Right, this kind of work (changing .pot files) is only done near the release to avoid churn. You're seeing this because the identifier was actually changed in #2389 but the translation files were not updated.

Try doing a `yarn i18n-prepare-untranslated` and I think you'll see the correct message

jholdstock:
`yarn i18n-prepare-untranslated` was helpful, thanks Matheus. 

I have reduced this PR to only include the variable renaming, which seems to be a valid change on its own. The old string `totalValueOfActiveTickets` does not appear anywhere else in the code base, and results in console errors being logged at runtime. Updating to `totalValueOfLiveTickets` matches an existing key in the i18n files and removes the error.

matheusd:
But that's the thing: it's the value in the pot file that's outdated (vs the one in the source file) but the pot files are only getting updated by the transifex translations, closer to release date. 

The workflow for translations is roughly:

- During development, keys in the source files change, are added, removed, etc
- Close to the release date I run `yarn i18n-prepare-untranslated`
- Upload source po files to transifex
- Translations are done
- Download generated pot files
- Send a PR updating everything

So translations are naturally broken during the standard dev cycle and the contents of the po/pot/json files are _not_ to be taken as authoritative.

jholdstock:
Thanks for explaining. This was actually my initial understanding of how things worked, but when I ran `yarn i18n-prepare-untranslated` with a clean repo, the string `totalValueOfActiveTickets` was not added to any of the updated files, so I am a bit confused.

Closing this PR because I assume it will be resolved when the new translations come in for 1.6.

PR #2599, opened 2020-07-23
Use decrediton own rpc.cert and rpc.key instead of the ones from dcrd

### LINKED ISSUE #2463: Backup rpc.cert and rpc.key before overwriting them
I've just run decrediton with changes from #2208 and it overwrote my `~/.dcrd/rpc.cert` and `~/.dcrd/rpc.key` without giving me any warning, and without creating a backup.

It would be great if a backup of these files could be created before they are deleted (probably silently, because the vast majority of users will not care).

relevant conversation from Matrix: https://matrix.to/#/!HEeJkbPRpAqgAwhXWO:decred.org/$158739936816638aEnfy:decred.org?via=decred.org&via=matrix.org&via=pintobyte.com

vctt94:
This PR fixes #2463 

With this change we are going to use the rpc.cert and rpc.key from decrediton's config directory, instead of the ones at dcrd, when starting decrediton with regular mode. 

If it is an advanced start and a different appdata, it will still need to update the rpc.cert and rpc.key, as we can't start dcrd with electron 8 as it doesn't support curve P-521.

~~Right now we are silently removing the files, as explained here https://github.com/decred/decrediton/pull/2599/files#diff-2c2e316b04b5e082a42f1340fc886b19R283~~

~~But I believe we can have a better approach for that. I believe we can make some kind of warn for the user so he makes a backup of his rpc.cert and rpc.key before removing them, what do you guys think?~~

Now, I am saving the rpc.cert and rpc.key at a directory called backup located at decrediton's config directory (on linux: `.config/decrediton/backup`)

I am also changing the config name from `upgd_electron8` to `is_electron8`, because I think `upgd_electron8` can be confusing as if we should upgrade it to electron 8 or if it is already updated

### DIFF
--- a/app/helpers/files.js
+++ b/app/helpers/files.js
@@ -1,3 +1,4 @@
+import fs from "fs-extra";
 
 // readFileBackward reads a file backward and if maxSize is specified it will
 // only read until reach that size in bytes.
@@ -36,29 +37,22 @@ export function readFileBackward(path, maxSize, end) {
 // makeFileBackup makes a backup of the file on the directory specified.
 // If the directory does not exists, it will be created.
 export function makeFileBackup(file, directory) {
-  try {
-    if (!fs.existsSync(file)) {
-      throw "File does not exists";
-    }
-    // get file name. Which probably is everything after the last /
-    // ex: /home/.dcrd/rpc.cert
-    const fileName = file.substr(file.lastIndexOf('/') + 1);
-    console.log(fileName)
-    // if directory does not exists, create it.
-    if (!fs.existsSync(directory)) {
-      fs.mkdirSync(directory);
-    }
-    // copy it to directory specified
-    fs.copyFileSync(file, `${directory}/${fileName}`, err => {
-      if (err) {
-        throw err;
-        return;
-      };
-    });
-
-    return true;
-  } catch (err) {
-    throw err;
-    return false;
+  if (!fs.existsSync(file)) {
+    throw "File does not exists";
   }
+  // get file name. Which probably is everything after the last /
+  // ex: /home/.dcrd/rpc.cert
+  const fileName = file.substr(file.lastIndexOf("/") + 1);
+  // if directory does not exists, create it.
+  if (!fs.existsSync(directory)) {
+    fs.mkdirSync(directory);
+  }
+  // copy it to directory specified
+  fs.copyFileSync(file, `${directory}/${fileName}`, err => {
+    if (err) {
+      throw err;
+    };
+  });
+
+  return true;
 }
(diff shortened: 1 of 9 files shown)


app/main_dev/launch.js
```diff
@@ -335,6 +371,125 @@ export const launchDCRD = (reactIPC, testnet, appdata) =>
...
+        const splitListen = value.split(":");
+        if (splitListen.length >= 2) {
+          rpc_host = splitListen[0];
+          rpc_port = splitListen[1];
+        }
+      }
+      if (!userFound && !passFound) {
+        // If user and pass aren't found on the top level, look through all
+        // next level config entries
+        for (const [key2, value2] of Object.entries(value)) {
```
matheusd, line 458: I'm confused: what sort of ini file requires parsing like this?

vctt94: That's a good question, tbh I only copied the code from our config.js to here. 

This was implement by @alexlyp  on https://github.com/decred/decrediton/pull/1111 maybe he recalls?

alexlyp: Don't remember the specifics, but I believe there were some corner cases that weren't allowing for the user/pass to be plucked from the config files properly.  So if they weren't found on the top level, then they would be looked through all the key/value config pairs for any hits.

app/helpers/files.js
```diff
@@ -0,0 +1,58 @@
...
+
+// makeFileBackup makes a backup of the file on the directory specified.
+// If the directory does not exists, it will be created.
+export function makeFileBackup(file, directory) {
+  if (!fs.existsSync(file)) {
+    throw "File does not exists";
+  }
+  // get file name. Which probably is everything after the last /
+  // ex: /home/.dcrd/rpc.cert
+  const fileName = file.substr(file.lastIndexOf("/") + 1);
```
matheusd, line 45: Can this break under windows or is `file` guaranteed to use `/` as separator? Usually the correct function to use in these situations is `basename()` (see [path.basename](https://nodejs.org/api/path.html#path_path_basename_path_ext)) but there may be considerations for not using that.

vctt94: Yeah, you are right, that will be a problem. What considerations are that?

vctt94:
I changed how I am getting file names to path.basename, instead. Plz let me know if you have any objections

app/helpers/files.js
matheusd: For example, if `file` is guaranteed to be coming from a hardcoded constant or was previously converted to use slashes independently of the platform, then `basename()` wouldn't work because it would determine the basename by the last `\` in windows.

PR #2600, opened 2020-07-23
[getstarted] adv. startup: ui fixes & tweaks

amassarwi:
This diff fixes advanced startup ui issues reported on #2552 

I replaced the text toggle which was used with the one we have in pi-ui.
As discussed with @vctt94 ideally we should replace all toggles with the one in pi-ui.

We already have a component called `TextToggle` so I left 2 TODOs in case we would like
to replace them as well.

As you can see we didn't use the component directly in the advanced startup form instead we used some classes from `TextToggle.less`, after refactoring to css modules, the classes defined in `TextToggle.less` won't work in GetStarted pages as they are using the hashed classes from the css module, this why the form styling was broken after the refactoring.
I added the missing styling in addition to the toggle replacement and now it looks much better.

After:
image

image


amassarwi:
@vctt94 fixed bud

PR #2606, opened 2020-07-27
SideBar: automated tests

bgptr:
We have discussed with @vctt94 on matrix, that the decrediton test coverage is pretty small, and it would be nice if we could add some automated tests. This diff is the first attempt in this direction. I have implemented tests on the `Sidebar` component, which is a relatively small part of the app and has already refactored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ bin/
 dist/
 app/i18n/extracted/app
 /*.node
+coverage/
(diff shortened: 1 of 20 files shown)


bgptr:
Thank you for the review @vctt94. Following your suggested link, I replaced most of`getByTestId` queries to `getByRole` which is the top-recommended query.

PR #2594, opened 2020-07-22
fix tx going to wrong place

vctt94:
fixes #2570

PR #2618, opened 2020-07-29
Fix transaction wrong path

guilhermemntt:
This PR corrects the path of the last ticket transaction, setting it to redirect to the right location. This bug has been found in https://github.com/decred/decrediton/pull/2594 as well. After this PR, there will be no further occurrence of ``transactions/history/txHash``.


guilhermemntt:
It seems like a network fail in Github when testing **yarn**. Thanks for the warning, I'll try to fix it.

guilhermemntt:
It looks good now. Github issues...

PR #2619, opened 2020-07-30
[cspp] not allow small balances to start mixer

vctt94:
This PR makes the mixer throws an error and not start if the change account balance has its balances smaller than the min rellay fee.


app/actions/ClientActions.js
```diff
@@ -675,3 +675,8 @@ export const abandonTransactionAttempt = (txid) => (dispatch, getState) => {
     })
     .catch((error) => dispatch({ error, type: ABANDONTRANSACTION_FAILED }));
 };
+
+export const getAcctSpendebleBalance = (acctId) => async (dispatch, getState) => {
```
alexlyp, line 679: spello --> Spendable

PR #2628, opened 2020-08-04
[governance] show eligible tickets only when voting is active or finished

### LINKED ISSUE #2610: [governance] proposal not yet voting showing eligible tickets
Right now we are showing eligible tickets for a proposal not yet on voting staging, example:

image

So as it is not on voting stage, it does not have eligible tickets, so we should not show this info on proposals on that stage.

amassarwi:
This diff closes #2610, it ensures eligible tickets section is displayed only when 
proposal voting is active or finished.

Also, this diff includes changes related to prettifying, as I ran `yarn prettify`.

### DIFF
--- a/app/components/views/ProposalDetailsPage/ProposalDetails.jsx
+++ b/app/components/views/ProposalDetailsPage/ProposalDetails.jsx
@@ -53,6 +53,9 @@ const ProposalDetails = ({
   const isDarkTheme = themeName === "theme-dark";
   const shortToken = token.substring(0, 7);
   const proposalPath = `/proposals/${shortToken}`;
+  const votingActiveOrFinished =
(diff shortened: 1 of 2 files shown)


app/components/views/ProposalDetailsPage/ProposalDetails.jsx
```diff
@@ -141,7 +143,7 @@ const ProposalDetails = ({
           />
         )}
       </div>
-      {walletEligibleTickets && (
+      {votingActiveOrFinished && walletEligibleTickets && (
```
amassarwi, line 146: That's was the actual the fix, all other changes are prettify changes!

amassarwi:
Sure, will do

amassarwi:
@vctt94 done mate

PR #2641, opened 2020-08-27
[LN] UI/UX improvements

fguisso:
- Use module css in all tabs pages
- add Inbound balance in the invoices tab
- add Outbound balance in the payments tab
- remove the total balance from channels tab
- add a progress bar in open channels to show the difference in Remote and Local balance(outbound/inbound individual balance)
- improve the wallet tab UI and backup UX


fguisso:
Payments Tab:
image

fguisso:
Invoices tab:
image

fguisso:
Channels tab:
image

matheusd (commented):
Looking nice!

A few suggestions to improve the design:

Use the same color consistently for total, inbound and outbound bandwidth. 

For example, in the "Channels" tab you show the local and remote balances in different colors. You should use the same (corresponding) colors on the max outbound/max inbound panels so that it's clear that the max inbound corresponds to the sum of remote balances and max outbound corresponds to the sum of local balances.

I don't think you need to show the Total Bandwidth in the "Invoices" and "Payments" tab (only the corresponding balance).

fguisso:
Wallet tab:
image

linnutee:
Hey @fguisso could you drop some screenshots of all the subviews. More insight to give some feedback on.

fguisso:
@linnutee Let me finish this and I make a video with all subviews and their functionality. I'm in the last task.

app/components/views/LNPage/NetworkTab/RoutesInfo.jsx
```diff
@@ -0,0 +1,57 @@
+import { FormattedMessage as T } from "react-intl";
+import { Subtitle, Balance } from "shared";
+import styles from "./NetworkTab.module.css";
+
+const Route = ({ route }) => (
+  <div className={styles.route}>
+    <span>Total amount</span>
```
matheusd, line 7: This (and the following labels) need to use the `<T ...` elements.

matheusd (commented):
Latest iteration is looking very nice :P

A few additional things I think would be worth modifying it here:

- Allow copying the node id in:
  - The channel list page
  - The payments page, after decoding an invoice
  - The network page, after listing a node (that is, allow to copy the counterparty when listing a channel)
  - On the query routes page (hops)
- Link the channel points in the network tab to the block explorer

matheusd (approved):

app/reducers/ln.js
```diff
@@ -242,12 +252,36 @@ export default function ln(state = {}, action) {
...
+      return {
+        ...state,
+        getRoutesInfoAttempt: true,
+        routes: null
+      };
+    case LNWALLET_GETROUTESINFO_SUCCESS:
+      return {
+        ...state,
+        getRoutesInfoAttempt: false,
+        routes: action.routes
```
matheusd, line 271: You need to set `nodeInfo: null` otherwise if  `LNWALLET_GETROUTESINFO_FAILED` is triggered once `nodeInfo` isn't cleared of the error and the route is never displayed.

app/components/views/LNPage/NetworkTab/NetworkTab.jsx
```diff
@@ -56,17 +59,116 @@ const QueryNode = ({ getNodeInfo }) => {
...
+    routesInfo,
+    tsDate,
+    getNodeInfo,
+    getNodeInfoAttempt,
+    getRoutesInfoAttempt,
+    getRoutesInfo
+  } = useNetworkTab();
+  const [activeTab, setActiveTab] = useState(0);
+  const [nodeID, setNodeID] = useState("");
+  const [amount, setAmount] = useState(0);
```
matheusd, line 150: Probably best to use `useState(1)` because it seems querying the route with a zero amount always fails.

There's a bug on `DcrInput` that prevents correctly displaying the initial value of 1 atom (https://github.com/decred/decrediton/issues/2682) but once that's fixed this should display correctly.

matheusd (changes requested):
Actually, I found two other nits.

app/components/views/LNPage/NetworkTab/RoutesInfo.jsx
```diff
@@ -0,0 +1,57 @@
...
+    <span><T id="ln.routesInfo.routeAmount" m="Total amount" /></span>
+    <div className={styles.infoContent}><Balance amount={route.totalAmt} /></div>
+    <span><T id="ln.routesInfo.routeFees" m="Total fees" /></span>
+    <div className={styles.infoContent}><Balance amount={route.totalFees} /></div>
+    <div className={styles.hopsGrid}>
+      <span><T id="ln.routesInfo.hop" m="Hop" /></span>
+      <span><T id="ln.routesInfo.hopFee" m="Fee" /></span>
+      <span><T id="ln.routesInfo.hopPubkey" m="Public Key" /></span>
+      {route.hopsList.map((hop, i) => (
+        <>
```
matheusd, line 17: This needs a `key` prop, otherwise it triggers an error.

fguisso: I don't know how key works, but here I need to use the `<>` empty tag to ensure that elements don't render inside another `<div>`.
There is another away?

matheusd: https://reactjs.org/docs/fragments.html#keyed-fragments

fguisso: I fix it in another way, declaring a new component and use it with `key` props:
```
<Hop key={...} />
```

app/components/views/LNPage/PaymentsTab/Page.js
```diff
@@ -296,11 +296,11 @@ export default ({
                     <span>Fee</span>
                     <span>PubKey</span>
                     {htlc.route.hopsList.map((hop, i) => (
-                      <>
+                      <React.Fragment key={"hop-"+i}>
```
amassarwi, line 299:
```suggestion
                      <React.Fragment key={`hop-${i}`}>
```

matheusd (approved):

PR #2642, opened 2020-08-27
[recent txs] Fix recent regular txs showing twice

vctt94:
It is possible right now, that a regular tx which is not mined for some reason, show twice at our recent regular txs selector. 

This PR fixes it.


alexlyp (approved):
ACK

PR #2645, opened 2020-08-31
[startup] Better deal with errors

vctt94:
I noticed on startup, if decrediton fails to get available wallets or if it fails to start a wallet, when choosing one, no feedback is given.

This PR fixes it. Example:

image


alexlyp (approved):

xaur:
If the screenshot above shows how it works _after_ the patch, this will be confusing for the users.

vctt94:
I think that's possible. @bgptr is working on https://github.com/decred/decrediton/pull/2683 to make the recover seed error more use friendly. 

I think we can do the same here

PR #2647, opened 2020-08-31
[launch] Return to daemon connection on error

### LINKED ISSUE #2621: [launch] After different network error decrediton goes to wrong state
After having a different network error (a mainnet daemon, connecting to a testnet decrediton, for example), decrediton goes to wallet selection, instead of going back to daemon connection

amassarwi:
Closes #2621

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -496,6 +496,7 @@ export const checkNetworkMatch = () => (dispatch, getState) =>
             error: DIFF_CONNECTION_ERROR,
             type: CHECK_NETWORKMATCH_FAILED
           });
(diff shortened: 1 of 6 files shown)


amassarwi:
@vctt94 I was able to connect after changing to the right testnet running daemon.

vctt94 (commented):
The issue described was fixed but we found another issue which amass is fixing

amassarwi:
@vctt94 added config files path to network mismatch error's action suggestion:
image

vctt94 (approved):
Great! Thanks for that amass

alexlyp (approved):

PR #2649, opened 2020-09-01
Fix extract P2SH from ticket

amassarwi:
This diff fixes a bug I experienced with my testnet wallet which was failing on startup while extracting P2SH for a ticket.


amassarwi:
@vctt94 Done mate

vctt94 (approved):
Great, thanks!

alexlyp (approved):

PR #2655, opened 2020-09-04
[deps] Update deps to fix dot-prop vuln

alexlyp:
This fixes the vulnerability alerts for dot-prop we are currently seeing.  As far as I can tell all previous issues we had with the locked version of electron-store and conf are no longer problems.

https://github.com/decred/decrediton/pull/1015 was the previous PR that rolled back the conf/electron store versions.


vctt94 (approved):
Tested and found no issues with yarn or the build

PR #2658, opened 2020-09-04
Fix spv wallet not starting

### LINKED ISSUE #2656: [spv] Wallets aren't loading into the overview while in SPV
On current master f61842f wallets on both testnet and mainnet are not properly loading the overview screen.  Appears to be fine in normal/non-SPV mode.

vctt94:
closes #2656


alexlyp (approved):
tACK

PR #2664, opened 2020-09-09
 Add set account passphrase

vctt94:
~~**contains db upgrade -- backup db before testing**~~

~~This PR depends on https://github.com/decred/decrediton/pull/2665 and on https://github.com/decred/dcrwallet/pull/1823~~

~~This PR depends on https://github.com/decred/dcrwallet/pull/1979~~

~~depends on #3337~~

**MAKE A DB BACKUP BEFORE TESTING, AS IT WILL SET ALL ACCOUNTS YOUR WALLET PASSPHRASE**

This PR adds a new step into the setup wallet machine, which sets all account's passphrase as the wallet passhrase.

Right now what I changed to use lock/unlock account instead of wallet:

 - send tx;
 - purchase ticket and auto buyer (legacy and non legacy);
 - privacy - running the mixer.
 
 The rest of the wallet operations still are with old functionalities


alexlyp:
Can rebase this now that 2665 is in

alexlyp:
Hrm looks like there's another rebase that's needed, sorry

alexlyp:
Now that 1.6.1 is pretty much done, a rebase is due on this one.

app/actions/ControlActions.js
```diff
@@ -840,50 +866,57 @@ export const startTicketBuyerV2Attempt = (
...
-      if (status.indexOf("Cancelled") < 0) {
-        if (status.indexOf("invalid passphrase") > 0) {
-          dispatch({ error: status, type: STARTTICKETBUYERV2_FAILED });
-        }
-      } else {
-        dispatch({ type: STOPTICKETBUYERV2_SUCCESS });
+  const { ticketBuyerService } = getState().grpc;
+  dispatch({ ticketBuyerConfig, type: STARTTICKETBUYERV2_ATTEMPT });
+  const accountNum = account.encrypted ? account.value : null;
+  const error = await dispatch(unlockWalletOrAcct(passphrase, accountNum));
```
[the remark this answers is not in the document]
vctt94: Now it is not possible to do other actions while mixer or auto buyer is running, anymore.

app/wallet/control.js
```diff
@@ -461,9 +443,11 @@ export const setVspdAgendaChoices = (
 export const unlockWallet = (walletService, passphrase) =>
   new Promise((resolve, reject) => {
     const unlockReq = new api.UnlockWalletRequest();
+    console.log(console.log(passphrase));
```
amassarwi, line 446: console

app/wallet/control.js
```diff
@@ -461,9 +443,11 @@ export const setVspdAgendaChoices = (
...
     unlockReq.setPassphrase(new Uint8Array(Buffer.from(passphrase)));
     // Unlock wallet so we can call the request.
     walletService.unlockWallet(unlockReq, (error) => {
+      console.log(error);
```
amassarwi, line 450: console

app/wallet/control.js
```diff
@@ -160,69 +157,54 @@ export const purchaseTickets = (
...
-        // Lock wallet and return response from the request.
-        const lockReq = new api.LockWalletRequest();
-        walletService.lockWallet(lockReq, (error) => {
-          if (error) {
-            reject(error);
-          }
-          resolve(response);
-        });
-      });
+
```
amassarwi, line 206: unnecessary empty line

amassarwi (commented):
Code looks good overall, left some minors.

alexlyp:
On the Account passphrase page during start up with Skip and X button are shown.  These should all be hidden since we are going to force the upgrade.

app/actions/ControlActions.js
```diff
@@ -220,23 +220,29 @@ export const SIGNTX_ATTEMPT = "SIGNTX_ATTEMPT";
...
+    );
+
+    dispatch({
+      signTransactionResponse: signTransactionResponse,
+      type: SIGNTX_SUCCESS
+    });
+    dispatch(
+      publishTransactionAttempt(signTransactionResponse.getTransaction())
+    );
+    await dispatch(lockWalletOrAcct(acctNumber));
```
matheusd, line 242: Is this still needed now that signing happens inside `unlockAcctAndExecFn`?

app/actions/ControlActions.js
```diff
@@ -324,27 +332,32 @@ export const purchaseTicketsAttempt = (
...
-      stakepool,
-      !dontSignTx
-    );
-    if (dontSignTx) {
-      return dispatch({
-        purchaseTicketsResponse,
-        type: CREATE_UNSIGNEDTICKETS_SUCCESS
-      });
-    }
+    await dispatch(lockWalletOrAcct(accountNum));
```
matheusd, line 360: Still needed?

app/actions/ControlActions.js
```diff
@@ -924,3 +946,175 @@ export const getPeerInfo = () => (dispatch, getState) => {
...
+  } catch (error) {
+    dispatch({ error, type: SETACCOUNTSPASSPHRASE_FAILED });
+  }
+};
+
+export const UNLOCKACCTORWALLET_ATTEMPT = "UNLOCKACCTORWALLET_ATTEMPT";
+export const UNLOCKACCTORWALLET_FAILED = "UNLOCKACCTORWALLET_FAILED";
+export const UNLOCKACCTORWALLET_SUCCESS = "UNLOCKACCTORWALLET_SUCCESS";
+
+// unlockAcctAndExecFn unlocks the account and perform some acction. Locks the
```
matheusd, line 1013: perform => performs
acction => action

app/actions/ControlActions.js
```diff
@@ -924,3 +946,175 @@ export const getPeerInfo = () => (dispatch, getState) => {
...
+  // execute method
+  try {
+    res = await fn();
+  } catch (error) {
+    fnError = error;
+  }
+
+  if (fnError !== null) {
+    throw fnError;
+  }
```
matheusd, line 1058: This will leave the account unlocked if `fnError` is filled in the block above. This needs to happen after the wallet is locked (if `leaveUnlock` is false).

vctt94: true, good catch.

app/actions/ControlActions.js
```diff
@@ -924,3 +943,171 @@ export const getPeerInfo = () => (dispatch, getState) => {
...
+
+  // execute method
+  try {
+    res = await fn();
+  } catch (error) {
+    fnError = error;
+  }
+
+  if (leaveUnlock) {
+    return res;
```
matheusd, line 1059: You also need to check for `fnError` here, otherwise you could return `res` when the actual `fn()` call failed.

matheusd:
- [ ] Block start ln wallet buttons if ticketbuyer/mixer/etc is running (same as txs)
- [ ] Improve copy for initial account passphrase setup (see below)
- [ ] Never lock dex account

Current copy for the initial account setup:

> Your wallet have accounts with no passphrase. Enter your passphrase so we can update it

This could be a bit scary for users ("accounts with no passphrase"). I suggest something along the lines of:

> (Title) Migrate to per-account passphrases

> Decrediton now uses per-account locking, which requires a one time migration. Enter your current passphrase to perform this upgrade. Visit docs.decred.org/wallets/decrediton/migrations to know more.

app/actions/ControlActions.js
```diff
@@ -1090,6 +1090,17 @@ export const unlockAcctAndExecFn = (
 
   // lock account
   try {
+    // do not allow locking of the dex account, as it isn't supposed to lock.
+    const dexAccountName = sel.dexAccount(getState());
```
matheusd, line 1094: You need to handle the case when the wallet does not have a dex account

matheusd (approved):

alexlyp (approved):

PR #2666, opened 2020-09-11
spv: spv mode indicator on sidebar

### LINKED ISSUE #2640: Add SPV mode label
As suggested by @davecgh:

> I think it would be a good idea to make the mode (SPV versus Full) more prominent.  Maybe down next to the block height.  e.g. Latest Block: 478390 (SPV or Full Mode Icon).

victorgcramos:
This diff closes #2640 by adding the SPV icon on "Latest Block" sidebar section.

Screen Shot 2020-09-11 at 10 31 06 AM

(on hover)
Screen Shot 2020-09-11 at 10 31 00 AM


vctt94 (commented):
Thanks for this PR. I think it looks good, but can you also add it when the sidebar is reduced or in the bottom?

victorgcramos:
So I think that we should use the label SPV instead of the icon. What do you think @vctt94

alexlyp:
Hrm I think the icon with a tooltip should suffice for now.  We can tweak if people seem to be confused after release.

victorgcramos:
IMO, on mobile view, adding the spv icon would pollute the current UI. Maybe we should go with the icon + tooltip for the full-width view, and use the label for collapsed views.

vctt94:
Fine by me

victorgcramos:
Collapsed view looks like this:
Screen Shot 2020-09-17 at 11 37 23 AM

vctt94 (approved):
LGTM

alexlyp (approved):

PR #2669, opened 2020-09-15
[accounts] fix: small improvements on privacy accounts flow

victorgcramos:
Noticed that after creating the default accounts required for mixing, there are no changes on the UI. This Diff redirects it to the correct privacy view, and removes some useless `useCallback` hooks.


vctt94 (approved):
LGTM

alexlyp (approved):
ACK

Issue #2636, opened 2020-08-21
config.json gets erased during shutdown


matheusd:
When running a packaged version on master, while shutting down the main `config.json` will sometimes get completely erased (as in: the file contents become empty).

There's probably some race condition somewhere that's causing the file not to be written or flushed to disk before the shutdown completes, which causes this.

This might be one of the root causes of the "Config File Error: There was an error while trying to load the config file, the format is invalid." in the wild.

PR #2671, opened 2020-09-16
make config singleton

vctt94:
This PR is an attempt to fix https://github.com/decred/decrediton/issues/2636.

I wasn't able to reproduce that bug, but starting a new config store each time getGlobalCfg() is called doesn't seem right to me. This PR changes it so we only start it once

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -6,15 +6,17 @@ import {
   getGlobalCfgPath,
   getWalletPath,
   dcrwalletConf,
-  getDcrdRpcCert
+  getDcrdRpcCert,
+  getBackupDirectory
 } from "./main_dev/paths";
 import * as cfgConstants from "constants/config";
+import { makeFileBackup } from "helpers";
 import { DCR } from "constants";
 
 let config = null;
 export function getGlobalCfg() {
   if (!config) {
-    config = initGlobalCfg()
+    config = initGlobalCfg();
   }
 
   return config;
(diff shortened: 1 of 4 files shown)


app/config.js
```diff
@@ -11,8 +11,13 @@ import {
 import * as cfgConstants from "constants/config";
 import { DCR } from "constants";
 
+let config = null;
 export function getGlobalCfg() {
-  const config = new Store();
+  if (!config) {
+    config = new Store();
+    return config;
```
matheusd, line 18: This return is unnecessary, since the next line will already return.

matheusd (commented):
Seems reasonable but I'm not sure this will ultimately fix that, though I haven't been able to consistently reproduce that issue anyway.

You should also change the reference in `initGlobalCfg()` to use the same config object.

One thing might be useful is creating a backup with the contents when `validateGlobalCfgFile()` fails.

vctt94:
Addressed your comment and now I am making a backup at decrediton's backup directory, of the validated config.json. 

At least now if it happens, we can recover the last valid config.

matheusd (commented):
Seems reasonable to add a backup where you put it but I'm afraid it'll end up being useless due to the likely sequence of events:

- Decrediton shuts down and the config file gets messed up
- Decrediton starts and user is presented with the "config file error" message. Decrediton shuts down.
- Decrediton restarts and creates the default config file (thus overwriting the backup)

Since we've added the backup, maybe we should add that if `validateGlobalCfgFile` fails for the current config file, you try to apply the backup instead (if it exists and is itself valid).

vctt94:
That's a possible scenario. Now I am trying to recover it when failing, if the parse is valid. If the content is lost, before the backup, I think the isn't much we can do, but testing with just a broken config, seems to work. 

I'd appreciate your feedback about it :)

alexlyp:
Closing for age, plus I'm not sure this was the correct approach to fix the linked issue.

PR #2672, opened 2020-09-16
[Trezor] Tx Signing Fixes

matheusd:
Some bugfixes for the current master for transaction signing when using trezor.

Individual commits should be easy to review individually.

### DIFF
--- a/app/helpers/msgTx.js
+++ b/app/helpers/msgTx.js
@@ -9,6 +9,10 @@ import {
 } from "./byteActions";
 import { Uint64LE } from "int64-buffer";
 
+// const SERTYPE_FULL = 0
+const SERTYPE_NOWITNESS = 1;
+const SERTYPE_ONLYWITNESS = 2;
+
 // selializeNoWitnessEncode gets a decoded tx and serialize it.
 // it returns its bytes buffer.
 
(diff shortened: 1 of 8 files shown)


app/helpers/msgTx.js
```diff
@@ -182,84 +186,136 @@ export function decodeRawTransaction(rawTx) {
...
-  position += 1;
-  switch (first) {
-    case 0xfd:
-      tx.numInputs = rawTx.readUInt16LE(position);
-      position += 2;
-      break;
-    case 0xfe:
-      tx.numInputs = rawTx.readUInt32LE(position);
+
+  if (tx.serType != SERTYPE_ONLYWITNESS) {
```
victorgcramos, line 190: I think we should use a strict comparison here instead of loose one.

victorgcramos (commented):
Code looks good! Just found one minor detail, see my inline comment. Nice job (as always) 😄

app/helpers/msgTx.js
matheusd: Nice catch. Updated.

vctt94 (approved):
I tested on latest trezor-firmware master, commit: cd1d5195fd0f414c7bcb41dfda27fe2f1c4fd064. 
I noticed a bug when initializing trezor, but signing is working fine!

LGTM.

PR #2673, opened 2020-09-17
[MyTickets] functional components, hooks & css modules 

### LINKED ISSUE #2521: TicketsPage: functional components, css module & custom hooks
To try document better our incremental efforts on the two epics #2438 & #2439, I thought we could create sub-issue for each page we are planning to work on. 
After migrating some parts of the app, we could follow the same pattern:
 - class components -> functional components
 - `*.less` files -> css modules
 - redux connectors -> custom hooks 

For more insights checkout the parts which have been refactored: #2486, #2457

TicketsPage folder:
image

### LINKED ISSUE #2670: [My Tickets] Header looks weird with sidebar reduced
When the sidebar is reduced, the headers of the table looks weird: 

image

victorgcramos:
This diff closes #2670 and closes #2521 by refactoring MyTickets page to a functional approach, by using hooks and css modules.

### DIFF
--- a/app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
+++ b/app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
@@ -78,7 +78,7 @@ const selectTicketTypeFromFilter = (filter) => {
   return ticketType && ticketType.key;
 };
 
-const MyTickets = (props) => {
+const MyTickets = () => {
   const {
     tickets,
     tsDate,
@@ -117,7 +117,6 @@ const MyTickets = (props) => {
   return (
     <TicketListPage
       {...{
-        ...props,
         selectedTicketTypeKey,
         selectedSortOrderKey,
         loadMoreThreshold,
--- a/app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
+++ b/app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
@@ -2,48 +2,73 @@ import { useState } from "react";
 import TicketListPage from "./Page";
 import { FormattedMessage as T } from "react-intl";
 import { useTicketsList } from "../hooks";
+import * as txTypes from "constants/Decrediton";
 
 const labels = {
-  unknown: <T id="ticket.status.multiple.unknown" m="unknown" />,
-  unmined: <T id="ticket.status.multiple.unmined" m="unmined" />,
-  immature: <T id="ticket.status.multiple.immature" m="immature" />,
-  live: <T id="ticket.status.multiple.live" m="live" />,
-  voted: <T id="ticket.status.multiple.voted" m="voted" />,
-  missed: <T id="ticket.status.multiple.missed" m="missed" />,
-  expired: <T id="ticket.status.multiple.expired" m="expired" />,
-  revoked: <T id="ticket.status.multiple.revoked" m="revoked" />
+  [txTypes.UNKNOWN]: <T id="ticket.status.multiple.unknown" m="unknown" />,
+  [txTypes.UNMINED]: <T id="ticket.status.multiple.unmined" m="unmined" />,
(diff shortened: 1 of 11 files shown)


vctt94 (changes requested):
So, it looks like this PR breaks the overview tx history row:

image

vctt94:
I also think it should show the tx name. If it is a vote, a ticket purchase, etc.

victorgcramos:
@vctt94 I think we should use the home page overview mode on my tickets page when the width is collapsed... Check my changes and give me your thoughts! Thanks for the review!

vctt94 (changes requested):
I agree and think that it looks good with the overview visual, but one thing, I noticed is the statistics page is broken now: 

image

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -1,120 +1,140 @@
...
 const labels = {
-  unknown: <T id="ticket.status.multiple.unknown" m="unknown" />,
-  unmined: <T id="ticket.status.multiple.unmined" m="unmined" />,
-  immature: <T id="ticket.status.multiple.immature" m="immature" />,
-  live: <T id="ticket.status.multiple.live" m="live" />,
-  voted: <T id="ticket.status.multiple.voted" m="voted" />,
-  missed: <T id="ticket.status.multiple.missed" m="missed" />,
-  expired: <T id="ticket.status.multiple.expired" m="expired" />,
-  revoked: <T id="ticket.status.multiple.revoked" m="revoked" />
+  [txTypes.UNKNOWN]: <T id="ticket.status.multiple.unknown" m="unknown" />,
```
vctt94, line 8: Thanks for that!

vctt94 (commented):
Thanks, it is mostly looking good. One thing I noticed is that the no more tickets bar has a different width: 

image

Beyond that I believe it is mostly looking good.

vctt94 (approved):
LGTM!

amassarwi:
Referencing relevant parent issues:
#2438  #2439

alexlyp:
needs a conflict resolved. then should be gtg

amassarwi:
@victorgcramos Please reference the corresponding issue as well. 
Closes #2521

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -1,120 +1,140 @@
-import { ticketsList } from "connectors";
```
amassarwi, line 1: Please delete the connector from connectors dir as well.

amassarwi:
Also, there are few class components inside TicketsPage sub folder as well, they all need to be refactored to functional components as well.

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
amassarwi: Indeed thanks 🔥 

Kill all these TODOs 🔪

amassarwi: @victorgcramos

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -1,120 +1,140 @@
...
+    onChangeFilter({ listDirection: type.value });
+    setSortOrderKey(type.value);
+  };
+
+  const loadMoreThreshold = 90 + Math.max(0, window.innerHeight - 765);
+
+  return (
+    <TicketListPage
+      {...{
+        ...props,
```
amassarwi, line 120: this isn't needed anymore, the props in this case was only the connector files, you can ditch it

alexlyp (approved):

PR #2674, opened 2020-09-17
[launcher] not possible to open a decrypted wallet

bgptr:
Uncaught TypeError on `OpenWallet` component: `onOpenWallet is not a function`


vctt94 (approved):
Great, thanks for fixing that! LGTM

alexlyp (approved):

PR #2677, opened 2020-09-24
[startup] avoid initial proposals batch double requests

victorgcramos:
This is minor, but I noticed that we always request the politeia api twice, one for active proposals and other for pre-voting proposals. Since the proposals batch request accepts an array of tokens as a param, we can make only one request using both.

We only need to double-request if active + pre-voting length is greater than 20 (max batch size), and this won't happen very often.


vctt94 (approved):
LGTM

alexlyp (approved):

PR #2683, opened 2020-10-01
Improve error handling on existing seed view

bgptr:
This diff improves the error handling on the existing seed view by changing the input's border color to red if the seed word is not at valid position.
ksnip_20201001-200949

### DIFF
--- a/app/components/views/GetStartedPage/CreateWalletPage/ExistingSeed/ExistingSeed.jsx
+++ b/app/components/views/GetStartedPage/CreateWalletPage/ExistingSeed/ExistingSeed.jsx
@@ -52,52 +52,89 @@ const ExistingSeed = ({
   );
 
   const onError = useCallback(
-    (seedError, seedWord) => {
+    (seedError) => {
       const seedErrorStr = seedError.details || "";
 
-      const countWords = () => {
-        let count = 0;
-        seedWords.forEach((wordObj) => {
-          if (wordObj.word && wordObj.word.length > 0) count++;
-        });
-        return count;
+      const populatedSeedWords = seedWords.filter(
+        (seedWord) => seedWord.word && seedWord.word.length > 0
+      );
+
+      // hide all error messages and indicators
+      const hideError = () => {
+        setError("");
+        if (seedWords.filter((seedWord) => seedWord.error).length > 0) {
+          setSeedWords(
+            seedWords.map((seedWord) => {
+              return {
+                ...seedWord,
+                error: false
+              };
+            })
+          );
+        }
       };
-      if (countWords() <= 1) {
+
+      if (populatedSeedWords.length <= 1) {
         // Weird errors with one word, better to avoid them.
         return;
       }
-      // if user has not completed filing all words we avoid showing invalid decoding.
-      if (seedErrorStr.includes(MISMATCH_ERROR) && countWords() < 33) {
+
+      if (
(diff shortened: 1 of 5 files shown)


app/components/views/GetStartedPage/CreateWalletPage/ExistingSeed/ExistingSeed.jsx
```diff
@@ -53,7 +53,7 @@ const ExistingSeed = ({
 
   const onError = useCallback(
```
vctt94, line 54: Can remove this use callback

app/components/views/GetStartedPage/CreateWalletPage/ExistingSeed/ExistingSeed.jsx
```diff
@@ -70,21 +70,24 @@ const ExistingSeed = ({
...
-        const index = errorStr.indexOf(POSITION_ERROR);
-        const numberPosition = index + POSITION_ERROR.length + 1;
-        const endErrorStr = errorStr.slice(numberPosition + 1);
-        const beginErrorStr = errorStr.slice(0, numberPosition);
-        return beginErrorStr + (seedWord.index + 1) + endErrorStr;
-      };
-      if (seedErrorStr.includes(POSITION_ERROR)) {
-        seedErrorStr = fixPositionError(seedErrorStr, seedWord);
+      if (seedErrorStr.includes(POSITION_ERROR) && seedWord) {
+        const regexp = new RegExp(`${POSITION_ERROR} (\\d+)`, "g");
```
vctt94, line 74: I think adding a comment here explaining what is happening would be good

app/components/views/GetStartedPage/CreateWalletPage/ExistingSeed/ExistingSeed.jsx
```diff
@@ -70,21 +70,24 @@ const ExistingSeed = ({
...
+        if (regexpArray != null && typeof regexpArray[1] !== "undefined") {
+          const updatedSeedWords = [...seedWords];
+          updatedSeedWords[seedWord.index] = seedWord;
+          updatedSeedWords[regexpArray[1]] = {
+            ...updatedSeedWords[regexpArray[1]],
+            error: true
+          };
+          setSeedWords(updatedSeedWords);
+        } else {
+          console.log("invalid position error msg format");
```
vctt94, line 85: why this log?

vctt94 (commented):
Thanks for this PR. 

I left a couple of comments, but I like the direction this is going.

bgptr:
Thank you for your review. I fixed your comments.

vctt94 (commented):
Thanks for fixing my comments.

I have a couple more comments. 

I noticed it is possible when entering a word in its wrong place, the wallet throws an error, but it does not show right after it. It only shows after you entering the next word, example:

Here the absurd word already errored, as you can see at decrediton logs:

image

But it only shows as wrong if I add the third word:

image

And if I delete the word from its position, the field border gets missed: 

image

 I also think it would be good keeping the error message instead of just making the word red, as it can be a bit confusing. Making it red and showing an error message like "word on position n is wrong" would make it clearer. What do you think?

bgptr:
Thank you for your review. I fixed the mentioned issues and cleaned up the component a bit. Also, I added the suggested error message. It is much more straightforward now.

vctt94 (approved):
LGTM. Good job on this PR!

PR #2684, opened 2020-10-05
trezor: Allow message signing.

JoeGruffins:
It doesn't look like the button has an `onSubmit`: https://github.com/decred/pi-ui/blob/master/src/components/Button/Button.jsx
Also was not passing needed params I believe. Please check me.


vctt94 (approved):
tACK

amassarwi (approved):

PR #2695, opened 2020-10-07
snackbar: Fix trezor methods.

JoeGruffins:
It doesn't look like you can use single quotes inside the messages. I also tried escaping with a backslash, but it still just prints out ...`{label}`

For the case, move items around so that everthing is called correctly. Add `TRZ_TOGGLEPINPROTECTION_SUCCESS` to the outter case and remove it from the very bottom. Add a similar condition to `TRZ_TOGGLEPASSPHRASEPROTECTION_SUCCESS` that toggles enabled/disabled. Add `TRZ_CHANGELABEL_SUCCESS` to the outer case.

Toggling the password and pin actually isn't working for me on master. Not sure why, but it should be fixed upon the conversion to trezor/connect.

### DIFF
--- a/app/reducers/snackbar.js
+++ b/app/reducers/snackbar.js
@@ -310,20 +310,20 @@ const messages = defineMessages({
   },
   TRZ_TOGGLEPINPROTECTION_SUCCESS_ENABLED: {
     id: "trezor.pinProtectionSuccess.enabled",
-    defaultMessage: "Pin protection has been enabled in trezor '{label}'"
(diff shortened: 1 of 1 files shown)


vctt94 (commented):
I wasn't able to fully test this PR because it looks like toggling pin protection on master is broken, but the code looks okay. 

Is toggling pin protection working for you?

JoeGruffins:
@vctt94 I implemented this in #2705 where everything should work. Thank you for looking, please continue testing there.

PR #2705, opened 2020-10-09
multi: Replace trezor.js with connect. 

JoeGruffins:
trezor.js is being deprecated. Switch to using trezor/connect to
communicate with a trezor device.

Swap trezor.js deps for connect. Pulling connect from my repository
temporarily.

connect has no way of listing connected devices or choosing a device (I believe).

There is no good outline of the API that I can find. It is defined in the code here:
https://github.com/trezor/connect/blob/develop/src/js/index.js

So, remove the device list. Also do not create separate
sessions. There is not only the one session or transport layer that we
communicate over.

Changes by file:

TrezorActions.js
 - Use a sigular session for communication. Adjust all methods
   appropriately.
 - Pull constants from trezor/connect.
 - Drop all references to a device list.
 - Store only the device ID and label.
 - Add a connect method.
 - Add initial init method settings.
 - Replace device listeners.
 - Add a listener for not backed up devices.
 - Adjust deviceRun to use a single session, error on no device, and
   also to catch errors in the returned payload.
 - Add getFeatures convenience method. No longer use stored features
   but request features from the device whenever they are needed.
 - Add noDevice convenience method.
 - Remove clearing device sessions. While this is a wanted feature, it
   is not yet clear how to achieve with connect. Leave a TODO to this
   affect.
 - Use boolean for canceled param in pin/passwork/word callbacks.
 - Move all attempt notifications to the beginning of methods, before
   the first possible failure.
 - Use named objects in session methods.
 - Add hex argument to signMessage and remove hex2b64 conversion.
 - Toggle methods and homescreen are now all part of applySettings.
   Other method names have also changed.
 - Add backup method.

Modals.js
 - Use local PassPhraseModal.
 - Replace reloadDeviceList with connect.

ConfigButtons.js
 - Use red button for passphrase toggle.
 - Remove clear device session button. While this feature would be good
   to have, it is not yet clear how to clear the session and/or logout.

RecoveryButtons.js
 - Use blue button for new backup option.

TrezorPage/index.js
 - Replace reloadDeviceList with connect.

connectors/trezor.js
 - Replace loadDeviceList and reloadDeviceList with connect.
 - Add backup device.

constants/trezor.js
 - Add Model T homescreen.
 - Removed now unused trezor-link constants.

helpers/trezor.js
 - If an output has an address_n it cannot have even a null address and
   vice-versa.
 - Amounts are now type checked to be strings.
 - script_sig must be in the bin outputs, and not in the outputs.

externalRequests.js
 - Now only communicating with trezor connect.

reducers/trezor.js
 - Remove device list constants.
 - Add connect, backup, and transport start/stop constants.
 - Do not store a device list. Store device ID and label.


JoeGruffins:
Tested on model T virtual environment and real model One. 

I hope it's alright to enable firmware installation in a separate pr. I was getting some real weird results and need some more time on that.

matheusd (approved):
Needs a rebase, but basic stuff is working with the emulators.

Updating the firmware through decrediton indeed isn't working but it seems that it's due to some change in firmware headers, so we can roll that change forward.

JoeGruffins:
Thank you. Also tested on Windows with a real device (I only have model 1) and working as expected there as well.

PR #2706, opened 2020-10-10
[transactions] fix: align transaction history page items

### LINKED ISSUE #2690: Uneven spacing on Transaction history page
Screenshot from 2020-10-07 10-26-15

The spacing and alignment is all over the place on this page, makes it pretty difficult to read. Maybe the data should be aligned as columns? 

I noticed that #2252 has a mockup for the transaction history page: https://xd.adobe.com/spec/c069846d-7dee-4deb-401b-8ccd3632078b-b658/screen/32bac0ce-68d1-409b-832d-87e583f8cb69/

Perhaps there is some quick fix for alignment which could be put in if #2252 is too large to be included in 1.6

victorgcramos:
This diff closes #2690 by adding some CSS tweaks to improve the fields organization in the TxHistory row.

### Before

**Regular view:**
    Screen Shot 2020-10-10 at 9 26 35 AM

**Small-width view:**
    Screen Shot 2020-10-10 at 9 33 20 AM

### After
**Regular view:**
    Screen Shot 2020-10-10 at 9 26 14 AM

**Small-width view:**
    Screen Shot 2020-10-10 at 9 32 25 AM

### DIFF
--- a/app/components/shared/TxHistory/RegularTxRow.jsx
+++ b/app/components/shared/TxHistory/RegularTxRow.jsx
@@ -6,21 +6,19 @@ import {
   TRANSACTION_DIR_TRANSFERRED
 } from "constants";
 import styles from "./TxHistory.module.css";
-import { classNames, Text } from "pi-ui";
+import { classNames } from "pi-ui";
 
-const TxDirection = ({ account, isCred, flat }) => (
-  <span className={"is-row"}>
+const TxDirection = ({ account, isCred }) => (
+  <span className="is-row">
     {isCred ? (
       <T
         id="txHistory.out.tx"
         m="To {acc}"
         values={{
-          acc: !flat ? (
+          acc: (
             <div className={styles.status}>
               <div className={styles.accountName}>{account}</div>
             </div>
-          ) : (
-            account
           )
         }}
       />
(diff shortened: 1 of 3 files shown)


amassarwi:
@victorgcramos Thanks for the fix.
Just thought while looking at your screenshots it may make sense to always show these tables in 100% width?

It may improve the ux in this case, what do you think ? maybe @linnutee can share his thoughts as well.

victorgcramos:
hey @amassarwi thanks for the feedback! I agree with you!

jholdstock:
Really nice improvement over what we currently have. Fixes #2690

linnutee:
Few small notices:
- Type sizes on the three elements From/To; Address Data and Date should all be the same size. Currently "From/To" is larger.
- The address data bg container is missing padding rules – vertical misalignment and possibly missing a bit of padding on the left/right side. Best to double check specs.
- The address data container uses a shorthand version of displaying first four (could be more) and last four characters of the address in the design files. Pro's to this is saving a fair bit of space while still allowing to identify each address by checking the beginning or ending of it. If that's deemed too short, I would recommend having the cutoff "…" in the middle, rather than the end.

@amassarwi there's some work done on figuring out wider views approach that would rather use several columns – I think it may be better to pursuit that. Adding a full-width rule would need to reflect on the rest of the app and that would look off on some of the views ie. adding too much empty space. Check these: https://github.com/decred/dcrdesign/issues/182 

Screenshot 2020-10-16 at 22 58 01

Screenshot 2020-10-16 at 22 58 56

PR #2707, opened 2020-10-10
Functional EyeFilterMenu & EyeFilterMenuWithSlider components

### LINKED ISSUE #2696: Dark mode: white panel on transaction history page
White panel stands out against everything else, I assume it was something which was missed whilst implementing dark mode.
"Amount Range" in black is difficult to read on dark blue background.

As a seperate thing - why is this panel closable with an x icon? It doesnt seem necessary, and its also impossible to restore the panel after it has been closed which is inconvenient.

Screenshot from 2020-10-07 12-23-01

amassarwi:
This diff first intention was to fix the issue reported here #2696.

I noticed that it's a good opportunity to refactor the two buttons involved to be functional and to get rid of some more `.less`
files(see #2438, #2439).

Includes the following changes:
 - Refactor buttons/EyeFilterMenu component to functional.
 - Refactor buttons/EyeFilterMenuWithSlider component to functional. 
- Introduce `useClickOutside` custom hook to replace `eventOutsideComponent` helper function used in 
class components.
 - Ditch EyeFilterMenu.less (replace with to css module, one for each button)
 - Ditch HistoryPage.less legacy file - Partially moved to EyeFilterMenuWithSlider.module.css and NoUiSlider global
styling moved to NoUISlider.css separate file (not a css module imprted as plain css). 
 - Fixes #2696.

After:
image

image


app/helpers/dom.js
```diff
@@ -14,6 +14,10 @@ export const isDescendant = (el, target) => {
 // `target` is outside the provided `el`. Eg: this is used to determine if
 // the target of an onMouseUp event is outside a root element.
 export const eventOutsideElement = (el, target) => {
+  console.log({
```
amassarwi, line 17: revert this!

jholdstock:
Tested this. It definitely fixes #2696 - the colors look correct and the unnecessary X button is removed, nice.

PR #2717, opened 2020-10-13
Dcr input func comp

### LINKED ISSUE #2682: [dcrInput] DcrInput with inputs smaller than 1 is buggy
It is possible that when the amount on dcrInput is smaller than 1 it shows as 0.

I think that maybe it can also be a good oportunity to refactor it to a functional component

vctt94:
This PR fixes #2682 

~~It is still a wip because right now after a transaction is sent, its value is kept, which I still haven't figured why.~~


matheusd (approved):
Working as intended, and definitely improving things, but doesn't yet solve the second part of https://github.com/decred/decrediton/issues/2682#issuecomment-704971929 (initial values aren't being respected).

That can be addressed in a separate PR.

PR #2731, opened 2020-10-16
[security] fix: improve sign message and address validation handlers

### LINKED ISSUE #2700: Sign Message validation triggering on Validate address field
- Enter a valid address into the Validate Address field. "Owned address" result appears - expected.
- Modify the Address field in the sign message form - the Validate Address result below changes to "Invalid address", even though the field value has not changed

Peek 2020-10-07 14-31

victorgcramos:
This diff Closes #2700, which reported an unexpected behavior regarding address validation on security page.

I figured out that the addressValidation stuff was using the redux tree to manage the address validation state. But since we use the `useValidateAddress` on multiple components, each one behaving differently according to the `address`, it makes no sense to store the responses on the redux tree.

I thought of caching those address in the redux tree, but I think it's an overkill fo this feature, since there are not much cases we need to keep the validation responses stored.

Therefore, the solution was to keep the validation response as a state variable, and remove the address validation away from the redux tree.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -151,10 +151,7 @@ export const IMPORTSCRIPT_SUCCESS = "IMPORTSCRIPT_SUCCESS";
 
 // importScriptAttempt tries to import the given script into the wallet. It will
 // throw an exception in case of errors.
-export const importScriptAttempt = (script) => async (
-  dispatch,
-  getState
-) => {
+export const importScriptAttempt = (script) => async (dispatch, getState) => {
   dispatch({ type: IMPORTSCRIPT_ATTEMPT });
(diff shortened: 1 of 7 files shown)


amassarwi:
@victorgcramos 🔥

vctt94 (approved):
Thanks for this PR. That's a great improvement!

jholdstock:
Looks good but seems to have broken the verify message functionality - opened #2750

victorgcramos:
@jholdstock actually, what broke the verify message functionality was the addition of these 2 args on `getServiceClient` function https://github.com/decred/decrediton/blob/b49caf51f0382ac5440decfe9836fa4138ce655c/app/middleware/grpc/client.js#L14-L15

PR #2733, opened 2020-10-16
fix windows build

vctt94:
This PR is built with the help of @jrick


jrick (commented):
module.cc changes look correct given node 14.14.0, but i'll leave a full review to the decrediton maintainers.

app/modules/win32ipc/module.cc
```diff
@@ -46,14 +46,15 @@ void CreatePipe(v8::FunctionCallbackInfo<v8::Value> const& args) {
     if ((uintptr_t)(double)pipe.read_end_handle != pipe.read_end_handle ||
         (uintptr_t)(double)pipe.write_end_handle != pipe.write_end_handle) {
         isolate->ThrowException(v8::Exception::Error(
-            v8::String::NewFromUtf8(isolate, "Handle is too large for double")));
+            v8::String::NewFromUtf8(isolate, "Handle is too large for double").ToLocalChecked()));
```
jrick, line 49: this isn't wrong but can use `NewFromUtf8Literal` instead, and would be consistent with the rest of the file that creates string objects from literals.

jrick: looks like all of the NewFromUtf8Literal calls have to be removed anyways in order to target node 12.13.0 (the version this electron release requires).

app/modules/win32ipc/module.cc
```diff
@@ -6,17 +6,24 @@
...
+        return v.ToLocalChecked();
+}
+
 void CreatePipe(v8::FunctionCallbackInfo<v8::Value> const& args) {
     auto isolate = v8::Isolate::GetCurrent();
 
     if (args.Length() != 1) {
         isolate->ThrowException(v8::Exception::TypeError(
-            v8::String::NewFromUtf8(isolate, "Wrong number of arguments")));
+            v8::String::NewFromUtf8(isolate, "Wrong number of arguments", NORMAL_STRING)));
```
jrick, line 21: you should use the function declared above instead for all of these. i'm guessing that you probably see build warnings because this ends up calling the deprecated function still.

app/modules/win32ipc/module.cc
```diff
@@ -6,17 +6,24 @@
 
 #include "pipe_wrapper.h"
 
+#define NORMAL_STRING v8::String::NewStringType::kNormalString
+v8::Local<v8::String> newFromUtf8(v8::Isolate* isolate, char const* str) {
```
jrick, line 10: newStringFromUtf8 is likely a better name (since it's no longer being called prepended by the class name)

alexlyp (approved):
tACK, great work!

PR #2748, opened 2020-10-20
[overview] fix: overview page adjustments

### LINKED ISSUE #2679: [responsiveness] overview looks weird in small widths:
The icon which shows the menu active looks weird and we should put the tables in the same column, so we can see them

image

victorgcramos:
This diff fixes some details on overview pages, such as tabs behavior, transactions and stake cards alignment and positioning

Closes #2679 

Screen Shot 2020-10-19 at 9 43 39 PM


matheusd (approved):

app/components/shared/TabsHeader/index.js
```diff
@@ -0,0 +1 @@
+export { default } from "./TabsHeader";
```
vctt94, line 1: This index is not necessary, you can use the file directly

vctt94 (commented):
Thanks for this PR, left only a small comment

vctt94 (approved):
Thanks. LGTM

PR #2777, opened 2020-10-23
Disable the send button and set it to the loading state on the manual ticket purchase page

### LINKED ISSUE #2762: Manual ticket purchase should block button and show loading indicator
Otherwise there's no indication the purchase process is being done.

bgptr:
Closes #2762


bgptr:
It seems #2762 was fixed already in #2773, so I close this PR.

PR #2783, opened 2020-10-23
[TransactionPage] fix: clear abandoned transaction from tx history lists

### LINKED ISSUE #2772: Abandoning tx does not remove it from list of pending txs
After reloading the tx does disappear and if the tx is eventually mined, it reappears correctly.

victorgcramos:
Closes #2772


vctt94 (commented):
Thanks for this PR, but can you move the logic from the reducer to the `abandonTransactionAttempt`, we are avoiding adding logic into our reducer files

vctt94 (approved):
LGTM

PR #2787, opened 2020-10-24
Multi: fix "Is Legacy" control misalignment, set its default value to checked, and make it persistent

### LINKED ISSUE #2784: VSP "Is Legacy" checkbox moves when it is checked
Peek 2020-10-24 10-00

### LINKED ISSUE #2785: Is Legacy should be checked by default
The new and legacy VSP forms are very similar, and when you open the Tickets tab its not immediately obvious which one you are looking at.

In order to keep a smooth UX for users of legacys VSPs (ie. everybody right now) - the legacy VSP form should be the default, and switching to the new VSP form should be a conscious choice.

As a bonus - decrediton should remember which one you had open most recently. Eg. If I buy legacy tickets, next time I open decrediton it should open the legacy form.

bgptr:
Closes #2785 
Closes #2784

### DIFF
--- a/app/actions/VSPActions.js
+++ b/app/actions/VSPActions.js
@@ -1,6 +1,6 @@
 // @flow
 import Promise from "promise";
-import { getWalletCfg, updateStakePoolConfig } from "config";
+import { getWalletCfg, getGlobalCfg, updateStakePoolConfig } from "config";
 import { importScriptAttempt, rescanAttempt } from "./ControlActions";
 import * as sel from "../selectors";
 import * as wallet from "wallet";
@@ -475,6 +475,8 @@ export const discoverAvailableStakepools = () => async (dispatch, getState) => {
 
 export const TOGGLE_ISLEGACY = "TOGGLE_ISLEGACY";
 export const toggleIsLegacy = (isLegacy) => (dispatch) => {
+  const config = getGlobalCfg();
+  config.set("vsp_is_legacy", isLegacy);
   dispatch({
     type: TOGGLE_ISLEGACY,
     isLegacy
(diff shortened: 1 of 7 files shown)


app/actions/VSPActions.js
```diff
@@ -467,6 +467,8 @@ export const discoverAvailableStakepools = () => async (dispatch, getState) => {
 
 export const TOGGLE_ISLEGACY = "TOGGLE_ISLEGACY";
 export const toggleIsLegacy = (isLegacy) => (dispatch) => {
+  const config = getGlobalCfg();
```
vctt94, line 470: wallet config

vctt94 (changes requested):
Same of #2789

bgptr:
Thank you for thre review. 
I moved the isLegacy param to the wallet config.

app/config.js
```diff
@@ -55,6 +55,9 @@ export function initWalletCfg(testnet, walletPath) {
   if (!config.has("trezor")) {
     config.set("trezor", false);
   }
+  if (!config.has("vsp_is_legacy")) {
+    config.set("vsp_is_legacy", true);
```
vctt94, line 59: Thinking if this should be true or false. 

What do you think @alexlyp ?

vctt94 (commented):
Looks good, but I am not sure if the ticket default value should be legacy or not.

I'd rather wait on ay-p's opinion

app/config.js
alexlyp: We should be defaulting to use the new VSP ticket purchasing system. (non-legacy)

bgptr: Alright. I've changed it to false and resolved the merge conflicts.

vctt94 (approved):
LGTM

PR #2789, opened 2020-10-26
[VSP] save last used VSP host

### LINKED ISSUE #2763: Remember last used VSP
On the new, per-VSP, accountless ticket purchase mode, it's probably best to remember the last VSP used for ticket purchasing so you don't have to select it again.

bgptr:
This update saves the last used VSP host to config after a successful ticket purchase on the new VSP page and sets the VSPSelects when the user arrives next time.
Closes #2763


app/actions/ControlActions.js
```diff
@@ -380,7 +381,9 @@ export const newPurchaseTicketsAttempt = (
         type: CREATE_UNSIGNEDTICKETS_SUCCESS
       });
     }
-    dispatch({ purchaseTicketsResponse, type: PURCHASETICKETS_SUCCESS });
+    const config = getGlobalCfg();
```
vctt94, line 384: This shouldn't be a global config, but a wallet config instead

vctt94 (changes requested):
Thanks for this PR.

The vsp should be a wallet config and not a global config, as each wallet can have different vsps selected, as default.

bgptr:
Thank you for thre review. 
I moved the vsp_host param to the wallet config.

degeri:
can we have this as an optional clickable setting ? worried about centralization risks here. We want people to feel like they can choose any VSP they want.

vctt94:
I think that's a valid point @degeri. Maybe we can have a checkbox near the ticket purchase area saying `use default`. What do you think @bgptr ?

bgptr:
@vctt94 @degeri Sure thing, I have added the checkbox.

jholdstock:
I've tried out the checkbox, I like the idea but it needs refining. I suggest

- Checkbox label should read "Always use this VSP"
- If no VSP is selected (ie. the dropdown says "Select VSP...") the checkbox should be hidden/deactivated. When a VSP is selected, checkbox appears/activates.
- When the checkbox is checked, the dropdown is deactivated to indicate that the value can not be changed. Checkbox must be unchecked again before dropdown value can change.

bgptr:
@jholdstock I like your suggestions. I've implemented them.

jholdstock:
Looking better but it seems like the selected vsp and checkbox are resetting when I closed/reopen decrediton.

Edit: Actually, I dont even need to restart decrediton. Just opening a new tab resets the values.

bgptr:
@jholdstock You are right. Probably it's because I saved the VSP host only after successful ticket purchasing. 

I revamped (and also cleaned up) the code to save immediately after clicking on the checkbox. 
It's more user-friendly now. Thank you.

vctt94:
needs rebase

bgptr:
rebased

vctt94 (approved):
I think that's a good start and we can start from here. Thanks @bgptr

Issue #2790, opened 2020-10-26
[launch] settings showing wallet option


vctt94:
On the launch settings, there is a button for update private passphrase, which shouldn't show because the wallet still not started

guilhermemntt:
On it.

PR #2792, opened 2020-10-26
Do not show change passphrase option when launching

guilhermemntt:
Closes https://github.com/decred/decrediton/issues/2790.

### DIFF
--- a/app/components/views/GetStartedPage/PrivacyPage/CustomPrivacyOptions.jsx
+++ b/app/components/views/GetStartedPage/PrivacyPage/CustomPrivacyOptions.jsx
@@ -17,7 +17,7 @@ export default ({ toggleCustomPrivacy, acceptCustomPrivacy, ...props }) => (
       />
     </div>
     <div className={styles.customPrivacySettings}>
-      <PrivacySettings {...props} />
+      <PrivacySettings {...{ ...props, showChangePassPhraseOption: false }} />
       <div className={styles.privacyButtons}>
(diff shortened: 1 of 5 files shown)


app/components/views/GetStartedPage/PrivacyPage/CustomPrivacyOptions.jsx
```diff
@@ -17,7 +17,7 @@ export default ({ toggleCustomPrivacy, acceptCustomPrivacy, ...props }) => (
       />
     </div>
     <div className={styles.customPrivacySettings}>
-      <PrivacySettings {...props} />
+      <PrivacySettings {...{ ...props, showChangePassPhraseOption: false }} />
```
vctt94, line 20: Thanks, but we already have a props which can be used for that. `walletReady`

vctt94 (commented):
Thanks for this PR, I left a comment about a props which can be used for that, instead of adding a new one into the component.

app/components/views/GetStartedPage/PrivacyPage/CustomPrivacyOptions.jsx
guilhermemntt: @vctt94 Great! That's good to know. I'll fix it.

guilhermemntt:
It seems to be ok now, @vctt94.

vctt94 (approved):
Thanks, LGTM

PR #2796, opened 2020-10-26
Vertical align rescan button with neighbouring text.

jholdstock:
Before 
Screenshot from 2020-10-26 17-52-09
After
Screenshot from 2020-10-26 17-50-03


vctt94 (approved):
Thanks for this PR. LGTM

PR #2809, opened 2020-10-27
[CSS] Improve LN Warning style

### LINKED ISSUE #2798: LN Warning style
This page stands out as not conforming to the Decrediton style. Eg:

- Colours are wrong - for a good example of correct colour/style for a markdown document, check "Help" > "Constitution"
- Links are barely visible in dark mode
- Left padding is insufficient when window is slightly shrunk horizontally
- Button needs more top padding
- Capitalizing every word on the button is odd. Would be better as  "I understand and accept the risks"

In addition, a bit of padding between each bullet will make the page look less like an imposing wall of text, and more like easily digestible chunks. This should increase the chance of people actually reading it.

Screenshot from 2020-10-27 08-41-49
Screenshot from 2020-10-27 08-42-45

bgptr:
Closes #2798

### DIFF
--- a/app/style/LN.less
+++ b/app/style/LN.less
@@ -35,13 +35,39 @@
 
 .ln-createwallet-warning-doc {
   max-width: 36em;
+  color: var(--modal-text);
 
-  p, ul, h1 {
+  h1 {
+    font-size: 27px;
+    line-height: 34px;
+    font-weight: 400;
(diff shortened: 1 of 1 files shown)


matheusd (approved):
Looks good to me.
/cc @jholdstock

jholdstock:
lgtm

Issue #2791, opened 2020-10-26
[LN] Balance tiles is too big


fguisso:
Balance tiles don't fit in the little screens, make them more responsive.

PR #2810, opened 2020-10-27
Make LN tiles more responsive

guilhermemntt:
This PR adjusts LN tiles responsiveness to make them fit better on small screens and resolutions.

Before:

image

After:

image

Closes https://github.com/decred/decrediton/issues/2791.

### DIFF
--- a/app/components/views/LNPage/ChannelsTab/ChannelsTab.module.css
+++ b/app/components/views/LNPage/ChannelsTab/ChannelsTab.module.css
@@ -107,14 +107,15 @@
   flex-direction: row;
   flex-wrap: wrap;
   justify-content: space-between;
-  margin-bottom: 40px;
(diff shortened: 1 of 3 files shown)


jholdstock:
The responsiveness works well for me, but the aesthetic still does seem to fit the rest of decrediton. The balance numbers are too big, and they dont use the correct monospace font (Source Code Pro). Check the main balance on the Overview tab for a good example of how this might look.

I expect the css for this already exists in decrediton somewhere and it can be reused, but if not, you could refer to https://github.com/decred/dcrdesign/files/2160926/decred.-.visual.toolkit.-.0618.pdf

The .pdf has a whole page about how to properly display DCR balances.

vctt94 (approved):
I think that's an improvement from what we have now, so for that reason LGTM

PR #2815, opened 2020-10-27
Ask for passphrase if account discovery is needed

matheusd:
This fixes a bug in SPV wallets where account discovery was broken on restored wallets which had more than one account. Discovery wasn't happening due to the passphrase not being passed along to the SpvSync.

It also modifies the get started sequence to ask for the private passphrase if account discovery hasn't completed yet, so that restarting a wallet restore after some blocks have been synced but before accounts have been discovered correctly finds the additional accounts.

Testing the passphrase request can be done by starting a wallet restore, then closing Decrediton during the initial block sync.

There are four different combinations for being presented the passphrase request view: 

- SPV Sync vs RPC Sync
- Wallet with vs without public passphrase

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -333,7 +333,7 @@ export const closeDaemonRequest = () => async (dispatch, getState) => {
   }
 };
 
-export const startWallet = (selectedWallet) => (dispatch, getState) =>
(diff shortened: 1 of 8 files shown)


alexlyp (approved):
ah sweet great catch!

Issue #2681, opened 2020-09-28
multi: Enable staking on trezor.


JoeGruffins:
This purpose of this issue is to track changes needed to enable staking through decrediton on the trezor hardware wallet.

Depends on changes tracked by #2680 

Staking also depends on changes being merged in the trezor firmware. Those changes are here:
https://github.com/trezor/trezor-firmware/pull/1249

Work on this in decrediton cannot begin until the above two changes are completed.

Ticket purchasing on testnet has been reliably achieved using a combination of these diffs with the above firmware changes:
https://github.com/trezor/trezor-utxo-lib/pull/9
https://github.com/trezor/connect/pull/678
https://github.com/matheusd/trezor-decred-poc/pull/6

JoeGruffins:
Related to #1491

JoeGruffins:
We have to decided to roll our own firmware as we wait for Trezor to merge. #2705 is in so we can continue. We were not able to get firmware updates working in that pr, so there will be an intermediate pr to get that working, then this issue can be solved for those who wish to use custom firmware rather soon I believe.

xaur:
Every once in a while people ask about the status of Decred staking with Trezor. This is the most relevant issue I could find. Having a short status for 2022 would make it a good reference to link to.

JoeGruffins:
All work on the trezor side is done actually! Will update this OP after a refresh on what still needs to be done over here and in dcrwallet.

JoeGruffins:
Well, I thought trezor was done for staking, but there is a bug in version v2.5.1 through v2.5.3 and am trying to get fixed [here](https://github.com/trezor/trezor-firmware/pull/2703)

#3848 allows purchasing on testnet. There is already a follow up pr ready that handles creating a new voting key that will allow vsps to vote for the trezor with deterministic voting addresses based on data produced from signing a message with the trezor.

The last piece that is still in development is error handling when fee payment does not go smoothly. I would like to handle this in decrediton, since it will require the user's input to attempt to re-pay a fee. We cannot do this passively in dcrwallet as I believe is the case normally.

xaur:
dcrwallet's VSP client (or the autobuyer?) has machinery for tracking VSP tickets throughout their lifecycle, doesn't it handle failed fee payments already?

JoeGruffins:
You have to pay the fee with trezor. So you have to confirm the signing of the transaction on the device. You can't just leave it unlocked like dcrwallet.

[...]

PR #2816, opened 2020-10-28
TrezorActions: Allow updating firmware.

JoeGruffins:
Part of #2681

We are unable to update firmware from TrezorActions.js. This is believed
to be due to the size of the firmware and the limitations of this
electron child process. We send the firmware to the main process to push
to the device. The main process uses a temporary connection to trezor in
order to accomplish this.

### DIFF
--- a/app/actions/TrezorActions.js
+++ b/app/actions/TrezorActions.js
@@ -782,7 +782,7 @@ export const updateFirmware = (path) => async (dispatch, getState) => {
   // Attempting to update the firmware while already updating will cause the
   // trezor to lock up.
   const {
-    trezor: { performingUpdate }
+    trezor: { performingUpdate, device }
   } = getState();
   if (performingUpdate) {
     console.log("already updating firmware");
@@ -800,9 +800,13 @@ export const updateFirmware = (path) => async (dispatch, getState) => {
   }
 
   try {
+    if (device != BOOTLOADER_MODE) throw "device must be in bootloader mode";
     // Ask main.development.js to send the firmware for us.
-    const errorStr = await ipcRenderer.invoke("upload-firmware", path);
-    if (errorStr) throw errorStr;
+    const { error, started } = await ipcRenderer.invoke("upload-firmware", path);
+    // If the updated started, the device must be disconnected before further
+    // use.
+    if (started) alertNoConnectedDevice()(dispatch);
+    if (error) throw error;
     dispatch({ type: TRZ_UPDATEFIRMWARE_SUCCESS });
   } catch (error) {
     dispatch({ error, type: TRZ_UPDATEFIRMWARE_FAILED });
--- a/app/actions/TrezorActions.js
+++ b/app/actions/TrezorActions.js
@@ -1,7 +1,7 @@
 import * as wallet from "wallet";
 import * as selectors from "selectors";
 import fs from "fs";
-import { hexToBytes, str2utf8hex } from "helpers";
+import { rawToHex, hexToBytes, str2utf8hex } from "helpers";
 import {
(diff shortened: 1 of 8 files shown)


app/actions/TrezorActions.js
```diff
@@ -111,20 +114,26 @@ export const TRZ_SELECTEDDEVICE_CHANGED = "TRZ_SELECTEDDEVICE_CHANGED";
 export const TRZ_NOCONNECTEDDEVICE = "TRZ_NOCONNECTEDDEVICE";
 
 function onChange(dispatch, getState, features) {
-  if (features == null) throw "no features on change";
+  if (features == null) return;
```
matheusd, line 117: Do the new versions not return a `features` object in boot loader mode? Why are you changing these from throws (which should ultimately at least log the error) to simple returns?

app/actions/TrezorActions.js
```diff
@@ -794,13 +803,11 @@ export const updateFirmware = (path) => async (dispatch, getState) => {
...
-
-    await deviceRun(dispatch, getState, async () => {
-      const res = await session.firmwareUpdate({
-        binary: rawFirmware.buffer
-      });
-      return res.payload;
-    });
+    const hexFirmware = rawToHex(rawFirmware);
+    // Ask main.development.js to send the firmware for us.
+    const errorStr = await ipcRenderer.invoke("upload-firmware", hexFirmware);
```
matheusd, line 808: It's best to send the path to the firmware file instead of the entire firmware contents to the ipc main thread.

matheusd (commented):
Going in the right direction.

Tested with the manually fixed firmware file (stripping the first 256 bytes of the firmware) and it worked on a physical model one.

Only thing I missed, which I'm not sure it's possible, is some feedback that the firmware install process is being performed.

Also, trying to install the wrong firmware makes the trezor set screen hang without detecting a device disconnection.

app/actions/TrezorActions.js
JoeGruffins: This probably didn't need to be changed. Reverting.

JoeGruffins:
> Also, trying to install the wrong firmware makes the trezor set screen hang without detecting a device disconnection.

Even after disconnecting the device? I am not seeing this. Does the trezor have a message about bad headers?

JoeGruffins:
Making a simple log of progress, but probably not where we want. For trezor 1 only progress 0 and 100 are announced, but I think model T might give increments in between. I would be good to put a progress bar for the user, but can I make an issue for now?

matheusd:
> Even after disconnecting the device? I am not seeing this. Does the trezor have a message about bad headers?

Correct. The sequence of events is:

- Put trezor in bootloader mode and connect it to computer
- Decrediton detects it
- Start firmware update with an invalid firmware
- (button to update firmware still enabled instead of disabled with a loading indicator, btw)
- Trezor complains about invalid header
- Disconnect trezor from computer
- Trezor Setup screen is still showing

After that, reconnecting the trezor device will "flash" the setup screen (i.e. it disappears for a moment, then reapears).

matheusd:
Sorry for the erroneous close, misclicked the button.

matheusd:
> I would be good to put a progress bar for the user, but can I make an issue for now?

Fair enough.

JoeGruffins:
Thank for the process. I understand now.

JoeGruffins:
Hopefully this handles failures better.

matheusd (approved):
Working great now (modulo the header thing)!

xaur:
This was reverted by #2931 due to build errors. Fixed version merged in #2932. (posting to link stuff together)

PR #2820, opened 2020-10-28
New Privacy menu item

### LINKED ISSUE #2806: Privacy should have its own tab
Privacy is a headline feature of the 1.6 release. It should be in its own tab, and have its own entry in the nav menu. It should not be hidden away in the Accounts tab.

bgptr:
This update renames the Security menu item to Privacy and moves the Accounts/Privacy tab as its first tab. The former Security page moves as its second tab.
Closes #2806


app/components/views/PrivacyPage/PrivacyPage.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+const PrivacyPageHeader = React.memo(
+  () => (
+    <StandaloneHeader
+    iconClassName="security"
+    title={<T id="privacypage.title" m="Privacy and Security Center" />}
+    description={
+      <T
+        id="privacy.description"
+        m="Various tools that help in different aspects of crypto currency privacy and security will be located here."
```
jholdstock, line 16:
```suggestion
        m="Tools that help in different aspects of cryptocurrency privacy and security."
```

vctt94 (approved):
After discussing on matrix, we decided to go with security on the privacy menu for now, so privacy can be a top menu. Therefore, LGTM after the suggestions from jhold being applied.

app/components/views/PrivacyPage/PrivacyPage.jsx
bgptr: Thank you for your review. I've changed the text.

alexlyp (approved):

PR #2821, opened 2020-10-28
Fix not updating feeStatus value after registering it successfully

vctt94:
This PR removes the warning of required props and update ticket fee status value, if vsp sync was successfully.


alexlyp (approved):
tACK

PR #2822, opened 2020-10-28
[TX Send] Fix `Send all funds` issue

### LINKED ISSUE #2799: "Send all funds" button does not activate Send button
- Open Transactions tab
- Enter a valid recipient address
- Click "Send all funds"
- Send button does not activate, even though the form is filled with valid data.

Peek 2020-10-27 09-01

bgptr:
Closes #2799


jholdstock:
Still seeing an issue here, to recreate:

- Paste a valid address into To field
- Type "1" into the Amount field. Hit backspace to delete it.
- Press "Send all" funds

Send button does not activate

bgptr:
Good catch, thank you. I've fixed it. 

Also, I found and fixed an issue too: 

- Send funds to another account
- Type valid amount
- Press "Send all" funds

After changing the `From` account the Tx details did not update. e.g the `Total amount sending` remained the same.

alexlyp (approved):

PR #2823, opened 2020-10-29
LN: Styling for backup panel.

jholdstock:
- Give the panel a background.
- Dont hide buttons behind an accordion. There is no benefit in making such a small panel collapsible, especially if we want to encourage users to use it regularly.
- Use a proper Decred icon rather than a stock icon from the web. I've also created a dark mode version.

## Before

Screenshot from 2020-10-29 12-07-07
Screenshot from 2020-10-29 12-07-22

## After

Screenshot from 2020-10-29 11-56-38
Screenshot from 2020-10-29 11-57-24
Screenshot from 2020-10-29 11-58-27


matheusd (approved):
Looking good! Thank you for this.

vctt94 (approved):
Nice! That's a good improvement

PR #2837, opened 2020-10-30
Increase threshold size, so it also loads more on smaller screens

### LINKED ISSUE #2697: Cannot change order of transaction history
Changing order from "Newest" to "Oldest" results in an infinite stakey dance. "Loading more transactions..." never stops.

Screenshot from 2020-10-07 12-29-26

### LINKED ISSUE #2766: Loading more transactions broken on small lists
It doesn't disappear when the wallet has very few transactions, such that all are fetched on the first batch.

vctt94:
closes #2697 and closes #2766


jholdstock:
This resolves #2697. Tested on wallets with large and small amounts of transactions.

I have not tested #2766 because I have never been able to reproduce that one

matheusd (approved):

PR #2839, opened 2020-10-31
Privacy: Hide mixed account branch.

### LINKED ISSUE #2807: Privacy: Hide Account Branch
Account Branch is a very technical detail to be exposing to users.

Given that the value is not even configurable, it is hard coded to zero, I suggest we do not display it to avoid confusion.

amassarwi:
Closes #2807


vctt94 (approved):
LGTM

PR #2840, opened 2020-11-01
build: test against latest node version

dajohi:


vctt94 (approved):
Tested on windows and linux with node 14.5

alexlyp (approved):

PR #2841, opened 2020-11-01
pi-ui: use default themes' names & override toggle bg

### LINKED ISSUE #2829: Toggle "Different Local Daemon Location" has no background
Toggle "Different Local Daemon Location" in start page has no background when using dark theme.

image

amassarwi:
While working on decred/decrediton#2829 I noticed that decrediton is using different theme names as the default ones used in pi-ui & pigui, this is a bit problematic and may cause some styling bugs, as we check internally in pi-ui if default dark theme is used by checking the current theme name: `const isDarkTheme = themeName === "dark"`, and using different theme names as we do in decrediton would cause some pi-ui's components to show wrong styling.

So as solution https://github.com/decred/pi-ui/pull/304 exports two constants for default themes names: `DEFAULT_LIGHT_THEME_NAME` & `DEFAULT_DARK_THEME_NAME`, and uses them when it determines if default dark styling should be used internally in pi-ui's components. After this commit consumers of pi-ui lib would need to provide these default themes names if they are interested in default components' dark/light styling.

This diff adjusts `decrediton` code base to work with the new exported default themes' names & overrides the dark 
toggle bar background color.

### Dependencies

Closes #2829 
Needs https://github.com/decred/pi-ui/pull/304

### DIFF
--- a/app/components/views/GetStartedPage/Settings/Settings.jsx
+++ b/app/components/views/GetStartedPage/Settings/Settings.jsx
@@ -11,7 +11,11 @@ import { KeyBlueButton } from "buttons";
 import { GoBackMsg } from "../messages";
 import { getGlobalCfg } from "config";
 import * as configConstants from "constants/config";
-import { useTheme } from "pi-ui";
+import {
+  useTheme,
+  DEFAULT_LIGHT_THEME_NAME,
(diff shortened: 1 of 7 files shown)


vctt94 (approved):
LGTM

PR #2844, opened 2020-11-02
[GetStarted] Don't ask for pwd on watching only wallets

### LINKED ISSUE #2842: watching only wallet: Asked for password to discover accounts.
This must be due to something recent. When creating a watching only wallet I am asked to provide a passphrase which does not exist to discover accounts that can't be discovered. Afterwards I am unable to open the wallet at all with the error: 
```
react_devtools_backend.js:6 Error: 12 UNIMPLEMENTED: wallet.Unlock: watching only wallet: watching wallets can not be unlocked
```

discoveraccounts

matheusd:
Fix #2842

### DIFF
--- a/app/components/views/GetStartedPage/hooks.js
+++ b/app/components/views/GetStartedPage/hooks.js
@@ -145,12 +145,13 @@ export const useGetStarted = () => {
         console.log("is At Start Wallet");
         const { selectedWallet } = context;
         const { passPhrase } = context;
+        const { isWatchingOnly, isTrezor } = selectedWallet.value;
         const hasPassPhrase = !!passPhrase;
         onStartWallet(selectedWallet, hasPassPhrase)
           .then((discoverAccountsComplete) => {
             setSelectedWallet(selectedWallet);
             const { passPhrase } = context;
-            if (!discoverAccountsComplete && !passPhrase) {
+            if (!discoverAccountsComplete && !passPhrase && !isWatchingOnly && !isTrezor) {
               // Need to discover accounts and the passphrase isn't stored in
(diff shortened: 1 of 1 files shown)


vctt94 (approved):
LGTM

app/components/views/GetStartedPage/hooks.js
```diff
@@ -145,12 +145,13 @@ export const useGetStarted = () => {
...
         const { selectedWallet } = context;
         const { passPhrase } = context;
+        const isWatchingOnly = selectedWallet.value.isWatchingOnly;
         const hasPassPhrase = !!passPhrase;
         onStartWallet(selectedWallet, hasPassPhrase)
           .then((discoverAccountsComplete) => {
             setSelectedWallet(selectedWallet);
             const { passPhrase } = context;
-            if (!discoverAccountsComplete && !passPhrase) {
+            if (!discoverAccountsComplete && !passPhrase && !isWatchingOnly) {
```
JoeGruffins, line 154: I guess you also need to add isTrezor, because it still asks me for the password if I check trezor but not wathing only.

matheusd: Trezors are also considered watching only wallets, and I tested the PR with a physical trezor and wasn't asked. What exactly was the flow you used that made you hit it?

JoeGruffins (approved):
No longer being asked.

app/components/views/GetStartedPage/hooks.js
matheusd: Updated to account for trezor wallets.

PR #2845, opened 2020-11-02
sidebar: Improve peers count

### LINKED ISSUE #2801: Peer count is ugly
The peer count is really ugly, the circle border does not match any other style in Decrediton. Large text, always visible at the top of the navbar is just not right for this.

I would suggest to put this at the bottom of the navbar, similar to the Latest block label.

Screenshot from 2020-10-27 09-18-37

amassarwi:
This diff improves the peers count a bit by moving it to the bottom of the sidebar, 
and showing it next to the latest block.

After:
image

image

---

Closes #2801

### DIFF
--- a/app/components/SideBar/Logo/Logo.jsx
+++ b/app/components/SideBar/Logo/Logo.jsx
@@ -10,12 +10,9 @@ const Logo = React.memo(
     onReduceSideBar,
     onExpandSideBar,
     isWatchingOnly,
-    getRunningIndicator,
-    peersCount
+    getRunningIndicator
   }) => (
-    <div
-      className={expandSideBar ? style.logo : style.reducedLogo}
-    >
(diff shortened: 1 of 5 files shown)


jholdstock:
Should we add an icon as well, so it has a similar look to the balance and rescan?

[peersIcon.zip](https://github.com/decred/decrediton/files/5480624/peersIcon.zip)

Screenshot from 2020-11-03 10-37-14

amassarwi:
I totally agree with both comments, will add the icon & play a bit with the spv icon positioning.

amassarwi:
Update:

image

Dark:

image

@jholdstock what do you think ?

jholdstock:
Big improvement imo, nice work

vctt94 (approved):
Thanks for this PR, I also think that's a good improvement!

PR #2850, opened 2020-11-03
Add darkmode rescan hover icon.

jholdstock:
Previously there was no darkmode icon, and the lightmode icon was barely visible.

## Before

Screenshot from 2020-11-03 11-03-42

## After

Screenshot from 2020-11-03 11-00-58


vctt94 (approved):
LGTM

Issue #2803, opened 2020-10-27
New VSP form issues


jholdstock:
- It is possible to select 0 tickets by pressing the "minus" button. 1 should be the minimum.
![Screenshot from 2020-10-27 09-37-48](https://user-images.githubusercontent.com/6762864/97283721-16854500-1838-11eb-85d2-4b6e9b32dc6e.png)

- "From Account", "From VSP" and "Amount" labels are too big. They should use the same style as the labels in the Automatic Ticket buyer form

![Screenshot from 2020-10-27 09-38-18](https://user-images.githubusercontent.com/6762864/97283781-2e5cc900-1838-11eb-8e9d-4b08a0a80312.png)

- "Staking Overview" and "Purchase ticket" sections are merged into one section, would be nice to keep them seperated

- Selecting a VSP which never responds - eg vspd.realprivacy.cc -  "Loading" indicator has no padding and no animation. The timeout is too long - seems to be 60+ seconds which is far too generous - 5 seconds is plenty. This happens in both the manual purchase form and the automated ticket buyer form.

![Screenshot from 2020-10-27 09-40-57](https://user-images.githubusercontent.com/6762864/97284129-90b5c980-1838-11eb-8ed2-688e08bb194a.png)

- The "i" menu which opens Ticket Purchase Information is no longer relevant for new vsp. Almost all of the information here is for legacy only, so lets hide it on the new form

victorgcramos:
I'd like to jump into this one after finishing #2786

jholdstock:
Hey @victorgcramos, I am making some changes on this form, so please message me before you start anything to prevent us overlapping.

victorgcramos:
no problem, @jholdstock! Thanks!

jholdstock:
I've updated the description above to remove items which were resolved by #2853

PR #2853, opened 2020-11-03
Refine purchase ticket form.

jholdstock:
- Only show "i" information icon on the legacy form.
- Splt "Staking Overview" and "Purchase Ticket" panels. Give them their own headers.
- Change "From Account" to "Account" and "From VSP" to "VSP". This matches the old form.
- Update "Is Legacy" to "Use Legacy VSP" and provide a tooltip for more information.
- Give Auto ticket buyer the same title style regardless of legacy/new.

Fixes most of #2803 - I will update that issue with the outstanding issues once this PR is merged.


victorgcramos (approved):
Really good improvements! Looking much better! Just a quick preview of the improved view in comparison to the old one:

**Improved view**
Screen Shot 2020-11-03 at 3 36 03 PM

**Current view**
Screen Shot 2020-11-03 at 3 36 31 PM

vctt94 (commented):
I think this is a good improvement, but I think it can be misleading with privacy ticket purchase. Here how it looks on this PR:

image

That's the reason we didn't use subtitlte on this view, so we can better "separate" them, into categories. So now the `privacy settings` looks weird. 

What do you think?

Also I think we can add a vertical margin between VSP and Account.

jholdstock:
Ah, I hadnt seen these privacy options on the ticket purchase tab before.

Aren't these options only relevant for the auto-buyer? Do they have any effect when making single, manual ticket purchases?

vctt94:
No, they are relevant for manual purchase, as well. For them to showing you just need to configure privacy

vctt94 (commented):
Thanks! This is looking good. 
I have a couple small nits

Can you put the checkbox inside the tooltip, so it shows when hovering it. Also can you make the mouse pointer when on the checkbox. 

image

jholdstock:
Now using pi-ui checkboxes, which works much nicer with the tooltip. Tooltip and mouse pointer now cover both the checkbox and the label.

vctt94 (approved):
Thanks, looking much better now! 

LGTM

PR #2854, opened 2020-11-03
[governance] "Finished Vote" tab filters

### LINKED ISSUE #2786: "Finished Voting" proposals ordering
Proposals in the "Finished Voting" list are ordered by their status - all of the "Passed" proposals come first, and then all of the "Declined/Quorum not met" proposals. I would expect this list to be time ordered. To provide an example, if I want to check the status of the D.R.E.A.M proposal, which finished voting only a few days ago, I need to scroll past every "Passed" proposal to find it, which is a long way down. The proposals are loaded in small batches, so scrolling down a long way is slow.

victorgcramos:
This diff adds the possibility to filter finished voting proposals by its approved or rejected status.

I think we cannot filter proposals by timestamp, since timestamp filtering is not implemented on politeia, but we can add some filters into the **Finished Vote** tab, so we could filter props by **Approved** and **Rejected** status.

Closes #2786

### DIFF
--- a/app/components/views/GovernancePage/Proposals/hooks.js
+++ b/app/components/views/GovernancePage/Proposals/hooks.js
@@ -7,6 +7,8 @@ import * as gov from "actions/GovernanceActions";
 import { usePrevious } from "hooks";
 import { setLastPoliteiaAccessTime } from "actions/WalletLoaderActions";
 
+const MAX_PAGE_SIZE = 20; // TODO: Get proposallistpagesize from politeia's request: /v1/policy
+
 export function useProposalsTab() {
   // TODO: move reducers which only control local states from reducer/governance.js to here.
   const activeVoteCount = useSelector(sel.newActiveVoteProposalsCount);
@@ -18,7 +20,7 @@ export function useProposalsTab() {
   const [tab, setTab] = useReducer(() => getProposalsTab(location));
 
   useEffect(() => {
-    dispatch(setLastPoliteiaAccessTime());
+    return () => dispatch(setLastPoliteiaAccessTime());
   }, [dispatch]);
   const getTokenAndInitialBatch = () => dispatch(gov.getTokenAndInitialBatch());
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


jholdstock (commented):
The filters buttons work nicely and solve my issue reported in #2786, but there is an expanse of blank space now (red on screenshot). I think this can be reduced by reducing the height of the buttons and the margin above/below them.

Screenshot from 2020-11-10 14-25-08

jholdstock:
lgmt

app/components/views/GovernancePage/Proposals/hooks.js
```diff
@@ -7,6 +7,8 @@ import * as gov from "actions/GovernanceActions";
 import { usePrevious } from "hooks";
 import { setLastPoliteiaAccessTime } from "actions/WalletLoaderActions";
 
+const MAX_PAGE_SIZE = 20; // TODO: Get proposallistpagesize from politeia's request: /v1/policy
```
amassarwi, line 10: Worth creating an issue

app/components/views/GovernancePage/Proposals/hooks.js
```diff
@@ -18,7 +20,7 @@ export function useProposalsTab() {
   const [tab, setTab] = useReducer(() => getProposalsTab(location));
 
   useEffect(() => {
-    dispatch(setLastPoliteiaAccessTime());
+    return () => dispatch(setLastPoliteiaAccessTime());
```
amassarwi, line 23: 👍

app/components/views/GovernancePage/Proposals/hooks.js
```diff
@@ -157,7 +156,7 @@ const onLoadMoreProposals = async (
 
   const proposalBatch = inventory.slice(proposalLength, proposalNumber);
   try {
-    await getProposalsAndUpdateVoteStatus(proposalBatch);
+    return await getProposalsAndUpdateVoteStatus(proposalBatch);
```
amassarwi, line 159: can you please ditch the console.log in the catch block ?

amassarwi (approved):
LGTM, added couple of small suggestions.

vctt94 (approved):
LGTM

PR #2873, opened 2020-11-06
[Remote daemon] Improve error message

### LINKED ISSUE #2855: Existing remote dcrd certificate unsupported curve
Virtually all existing dcrd setups use certificates with the P-521 curve.  Unfortunately, it appears that v1.6.0 of Decrediton no longer supports that curve (probably because Chromium which electron uses removed support for it).

The consequence is that all existing setups that have Decrediton configured to connect to the remote dcrd instance now fail as follows:

image

I **highly** suggest Decrediton detect the error and give the users clear instructions on how to resolve it  I've put up a PR to dcrd to change the default curve to P-256, but that obviously won't fix all of the existing certs and setups.  The way the user will need to fix it is to remove the `rpc.cert` and `rpc.key` and restart dcrd with the `--tlscurve=P-256` param (or that will be the default curve once the aforementioned PR lands) to allow it to generate a cert and key with that supported curve.

Ideally, I think Decrediton itself should just examine the certificate the user specifies to see if it has a supported curve, and if not, provide a clear message why the cert can't be used and how to resolve it.  Barring that, it should at least detect the error when trying to connect to dcrd.

bgptr:
Closes #2855
This update adds a more detailed error message with instructions when Decrediton tries to connect to a daemon, which uses the unsupported P-521 curve.
ksnip_20201106-183013

### DIFF
--- a/app/components/views/GetStartedPage/hooks.js
+++ b/app/components/views/GetStartedPage/hooks.js
@@ -12,7 +12,10 @@ import WalletPubpassInput from "./OpenWallet/OpenWallet";
 import DiscoverAccounts from "./OpenWallet/DiscoverAccounts";
 import ReleaseNotes from "./ReleaseNotes/ReleaseNotes";
 import { ipcRenderer } from "electron";
-import { OPENWALLET_INPUT, OPENWALLET_INPUTPRIVPASS } from "actions/WalletLoaderActions";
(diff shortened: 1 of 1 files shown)


vctt94 (approved):
Thanks! LGTM

PR #2882, opened 2020-11-09
Introduce generic ToggleSwitch with customizable text

jholdstock:
- Fix an issue where the tooltip on the mixer toggle said "Turn on auto buyer".
- Rename the css from `autobuyer-switch` to `toggle-switch` and move it to a dedicated file.
- Remove `RestoreWalletSwitch` which was a duplicate of the `AutoBuyerSwitch`.
- Remove unused `NetworkSwitch`.


app/components/buttons/ToggleSwitch.js
```diff
@@ -0,0 +1,27 @@
+import "style/ToggleSwitch.less";
```
amassarwi, line 1: Could you please change this with css module ? have a look `*.module.css` files and how we use them in `.jsx` files.

app/components/buttons/ToggleSwitch.js
```diff
@@ -0,0 +1,27 @@
...
+// enabled shows the switch as on or off.
+// disabled unable to trigger the onClick method.
+// enabledText and notEnabledText the tooltop contents when the switch is on/off.
+const ToggleSwitch = ({ enabled, onClick, disabled, enabledText, notEnabledText }) => (
+  <Tooltip
+    text={ enabled ? enabledText : notEnabledText }>
+    <div className="toggle-switch">
+      <div
+        className={
+          enabled ? "toggle-switch-enabled" : "toggle-switch-disabled"
```
amassarwi, line 13: use `classNames` function which can be imported from `pi-ui` to use conditioned classes in a cleaner way

app/style/ToggleSwitch.less
```diff
@@ -0,0 +1,47 @@
+@import (reference) "./main.less";
```
amassarwi, line 1: The should be a css module, and should be moved to the component's dir

app/components/buttons/ToggleSwitch.js
amassarwi, line 1: All new or refactored components should have `.jsx` extension to make it clear which includes plain `.js` and which `.jsx`

amassarwi (commented):
LGTM overall, left some comments.

app/components/buttons/ToggleSwitch/ToggleSwitch.module.css
```diff
@@ -0,0 +1,47 @@
+.toggleSwitch {
+  display: inline-block;
+  vertical-align: middle;
+  margin-right: 0.5em;
+
+  margin-bottom: 10px;
+  width: 85px;
+}
+
+.toggleSwitchEnabled,
```
amassarwi, line 10: i would get rid of the `toggleSwitch` prefix in all classnames this is needed in .less files but not in a css module
as classes are hashed and we are safe against classes conflicts

app/components/buttons/ToggleSwitch/ToggleSwitch.jsx
```diff
@@ -0,0 +1,30 @@
...
+// enabled shows the switch as on or off.
+// disabled unable to trigger the onClick method.
+// enabledText and notEnabledText the tooltop contents when the switch is on/off.
+const ToggleSwitch = ({ enabled, onClick, disabled, enabledText, notEnabledText }) => (
+  <Tooltip
+    text={ enabled ? enabledText : notEnabledText }>
+    <div className={classNames(style.toggleSwitch)}>
+      <div
+        className={classNames(
+          enabled
```
amassarwi, line 14: actually here you don't need the classNames, it's useful when you have something like
`classNames(isOdd && style.odd, isGreen && style.green, ...` but in your case the condition will produce one class either enabled or disabled, so it can be simplified to not use classNames

amassarwi (approved):
ACK

vctt94 (approved):
Thanks for the cleanup! LGTM

PR #2886, opened 2020-11-09
Remove max wallets limit.

### LINKED ISSUE #2186: Create/restore wallet hidden if 3 wallets exist
If you create 3 wallets (regular or watch only) in Decrediton you will lose the ability to create more as the "create" and "restore" buttons will disappear.

For a temporary workaround you can move the directories out of /wallets/mainnet and then put them back when you are done creating additional wallets.

jholdstock:
The limit was imposed to protect the UI in case of a large number of wallets, but actually the UI handles large numbers of wallets pretty well.

Closes #2186


vctt94 (approved):
Yeah, that was a band-aid on older releases, but it was fixed. Thanks for removing it

PR #2887, opened 2020-11-09
Fix peer count grammar.

jholdstock:
`Peers Count: 8` is not correct - we should either go with `Peers: 8` or `Peer Count: 8`.

I have gone with `Peers: 8` because the word "count" seems redundant.


vctt94 (approved):
LGTM

PR #2888, opened 2020-11-09
Add log to privacy page

vctt94:
This PR adds log to privacy page

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -607,6 +607,12 @@ export const getDcrwalletLogs = () => () =>
       .catch((err) => reject(err))
   );
 
+export const getPrivacyLogs = () => () => new Promise((resolve, reject) => {
+  wallet.getPrivacyLogs()
+    .then((logs) => resolve(logs))
+    .catch(err => reject(err));
+});
+
 export const getDecreditonLogs = () => {
   wallet
     .getDecreditonLogs()
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -607,11 +607,11 @@ export const getDcrwalletLogs = () => () =>
       .catch((err) => reject(err))
   );
 
-export const getPrivacyLogs = () => () => new Promise((resolve, reject) => {
+export const getPrivacyLogs = () => () => new Promise((resolve, reject) =>
   wallet.getPrivacyLogs()
     .then((logs) => resolve(logs))
-    .catch(err => reject(err));
-});
+    .catch(err => reject(err))
+);
 
 export const getDecreditonLogs = () => {
   wallet
--- a/app/components/views/PrivacyPage/Privacy/Privacy.jsx
+++ b/app/components/views/PrivacyPage/Privacy/Privacy.jsx
@@ -1,8 +1,11 @@
 import PrivacyContent from "./PrivacyContent";
 import ConfigMixer from "./ConfigMixer/ConfigMixer";
 import { usePrivacy } from "./hooks";
+import { useMountEffect } from "hooks";
+import { useState } from "react";
+import ReactTimeout from "react-timeout";
 
-const Privacy = ({ isCreateAccountDisabled }) => {
+const Privacy = ({ isCreateAccountDisabled, setInterval }) => {
   const {
     stopAccountMixer,
     accountMixerRunning,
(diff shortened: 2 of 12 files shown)


app/actions/DaemonActions.js
```diff
@@ -607,6 +607,12 @@ export const getDcrwalletLogs = () => () =>
       .catch((err) => reject(err))
   );
 
+export const getPrivacyLogs = () => () => new Promise((resolve, reject) => {
```
amassarwi, line 610:
```suggestion
export const getPrivacyLogs = () => () => new Promise((resolve, reject) => 
```

amassarwi (approved):
LGTM, left one suggestion to remove extra `{}`

vctt94:
Thanks amass. Added your suggestion

app/components/views/PrivacyPage/Privacy/PrivacyContent.jsx
```diff
@@ -55,6 +56,14 @@ const PrivacyContent = ({
     {accountMixerError && (
       <div className={style.error}>{accountMixerError}</div>
     )}
+    { accountMixerRunning && (
+      <div className={style.logs}>
+        <div className={style.logsTitle}>
+          <T id="privacy.mixer.logs" m="Logs:" />
```
jholdstock, line 62: Maybe use a standard `<Subtitle>` for this?

jholdstock (commented):
If I generate some logs, then leave the privacy tab, when I come back to the privacy tab the log is empty. It takes some time (I would guess 2000ms) for the log to reappear. Can that be fixed?

app/components/views/PrivacyPage/Privacy/Privacy.module.css
```diff
@@ -35,4 +35,20 @@
 }
 .contentTitleButtonsArea {
   margin-left: auto;
-}
\ No newline at end of file
+}
+.logs{
```
jholdstock, line 39: Probably need some max height here, and vertical scroll when overflowing

jholdstock, line 39: Most of this css is already defined for the log panels in the Help tab. I guess we want the appearance for this new log panel to be the same, so should the existing css be reused rather than duplicating it here?

jholdstock, line 39: Most of this css is already defined for the log panels in the Help tab. I guess we want the appearance for this new log panel to be the same, so should the existing css be reused rather than duplicating it here?

amassarwi: this is a good example of a good opportunity to wrap shared styling in a component, with `.less` files you would import the styling in both places and this can make it very tough to maintain, but with css module we should move shared styling to a separated component with it;s own css module and then use the component in as many places as we like.

app/components/views/PrivacyPage/Privacy/PrivacyContent.jsx
vctt94: changed

app/components/views/PrivacyPage/Privacy/Privacy.module.css
vctt94: I added the logs into a textarea, so it has a similar behavior from logs. Thanks for the catch.

vctt94: I added a TODO to create a shared component for logs and using it instead. Will create an issue as well

jholdstock:
- Seeing an error if I start the mixer, immediately open a different tab, and then come back to the Privacy tab before anything has been added to the log panel.
```react-dom.development.js:89 Warning: `value` prop on `textarea` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.```

- Log is still removed when I turn the mixer off. As mentioned above, I think it should continue to be displayed.

File: app/components/views/PrivacyPage/Privacy/Privacy.module.css (hunk 3 above)
Comment (amassarwi) [reply app/components/views/PrivacyPage/Privacy/Privacy.module.css:39]: awesome, thanks @vctt94

File: app/components/views/PrivacyPage/Privacy/Privacy.jsx [hunk 4]
```diff
@@ -1,18 +1,39 @@
...
   } = usePrivacy();
 
+  const [logs, setLogs] = useState("");
+  useMountEffect(() => {
+    // get initial logs
+    onGetPrivacyLogs()
+      .then(privacyLogs => setLogs(privacyLogs.toString("utf-8")))
+      .catch(err => err);
+
+    setInterval(async () => {
```
victorgcramos, line 27: What about clearing the interval here on unmount?

app/main_dev/logging.js
```diff
@@ -106,12 +107,31 @@ export const AddToDcrlndLog = (destIO, data, debug) => {
   dcrlndLogs = AddToLog(destIO, dcrlndLogs, data, debug);
 };
 
+export const AddToPrivacyLog = (destIO, data, debug) => {
+  // if log contains any of those messages we consider it a privacy log.
+  const privacyLogsArray = [
+    "Dialed CSPPServer",
```
jholdstock, line 113: Maybe `RunAccountMixer` too? Or perhaps even just `AccountMixer`?

jholdstock (approved):
looks good!

alexlyp (approved):

PR #2890, opened 2020-11-09
Bump grpc to 1.24.3

alexlyp:


jholdstock (approved):
Works for me, but only tested on Linux.

vctt94 (approved):
LGTM. Tested on linux and on windows and found no problems.

PR #2893, opened 2020-11-10
Standardize mixing account names.

### LINKED ISSUE #2881: Remove references to "Change" account
The account names "unmixed" and "mixed" are descriptive and obvious - but we also refer to the unmixed account as the "change" account.

I understand the reasoning behind this, but I don't agree that it is necessary in Decrediton, and it can lead to confusion. When one thing has two different names, the uninitiated will assume it is actually two things. 

Let's simplify and just refer to it as "unmixed" everywhere.

jholdstock:
All of these are changed to "Unmixed account"...

- "Change account"
- "Unmix account"
- "Change acct"

All of these are changed to "Mixed account"...
- "Mixed acct"
- "Mixing account"
- "Mix account"

Closes #2881


vctt94 (approved):
Thanks. LGTM!

PR #2895, opened 2020-11-10
[Tickets] Show "sync" button on fee error automatically

### LINKED ISSUE #2858: [1.6.0-rc2] Show "sync" button on fee error automatically
Related to #2857 

After the ticket errored, in order to send a new fee tx I had to switch the ticket filter to "Fee error" so that the "Sync" button (that sends another fee tx) is shown.

It's probably a better idea to show this button any time a ticket fee error is detected. Another alternative would be figuring out a different place to show the the button, since the existing method relies on directing users to a somewhat "hidden" feature.

bgptr:
Closes #2858 
Now, if a ticket is errored the sync button is visible independently of the ticket filter.


vctt94 (approved):
LGTM!

PR #2901, opened 2020-11-11
Functional Wallet container

### LINKED ISSUE #2891: Decrediton making calls after shutdown requested
We need to clean requests that happens sporadically before closing decrediton.

amassarwi:
This diff fixes #2891, as we had intervals to fetch politeia's token inventory every minute & wallet GetPeerInfo call every ten seconds but we didn't cleanup the intervals on container's unmount.

This commit includes a bit more than that, as I took the opportunity to refactor the `Wallet` container to functional approach.

Changes:
 - Functional `containers/Wallet.jsx`.
 - Get rid of `WalletContainer` connector & introduce `useWallet` hook.
 - Partially get rid of `Theming` connector and replace it with `useTheming` hook.
 - Use `useEffect` to set the intervals & to clean them up on unmount.

---

Closes #2891
Part of #2438


thi4go (approved):
tACK and looking good to me mate, nice work towards #2438

amassarwi:
@vctt94 rebased mate

vctt94 (approved):
Thanks. LGTM!

@bgptr can you see if the issue you've seen is fixed by this PR?

app/containers/Wallet.jsx
```diff
@@ -0,0 +1,102 @@
...
+    const piInvInterval = setInterval(() => {
+      if (politeiaEnabled) {
+        compareInventory();
+      }
+    }, 60000);
+
+    // Get peer info every 10 seconds, so we can no if there are no available
+    // peers.
+    const peerInfoInterval = setInterval(() => {
+      if (politeiaEnabled) {
```
bgptr, line 55: Is `politeiaEnabled` have to be `true` here for getting peer info?

bgptr:
@amassarwi @vctt94 Looking great. I just left an inline question.

app/containers/Wallet.jsx
vctt94: No it shouldn't good catch!

vctt94: @amassarwi

amassarwi: that's was the original condition, will get rid of it

PR #2909, opened 2020-11-11
Add checkbox warning

### LINKED ISSUE #2877: [1.6.0-rc2] sending funds is only possible from mixed account
The user is not able to access any funds that are spread over his different accounts. This will ultimately lead the user beeing forced to multiple mix and transact in order to manage his own accounts, makes the usability of transactions with the wallet depending on cssp epochs unuseable for a lot of use cases. User should have the opportunity to operate mixing, not constantly required to do so.

Please let the user decide from which account they want to send funds from. I can understand the way it is implemented strict right now, but at least give a section in the settings to enable the selection of accounts again. Grey out account selection for the other users, as it obviously would 'not be required' when sending funds only from mixed account per default.

### LINKED ISSUE #2892: [1.6.0 rc2] Perpetually mixing and unhelpful error
I have the mixing animation and message going for several days now, the balance has been 0 in the unmixed account and no transactions for >24 hrs.

Not sure if related but when I eventually turned off the autobuyer it gave a not helpful error.

image

vctt94:
This PR adds a checkbox for allowing sending transaction from other accounts than mixed account. closes #2877


app/components/views/PrivacyPage/Privacy/Privacy.jsx
```diff
@@ -14,10 +14,27 @@ const Privacy = ({ isCreateAccountDisabled, setInterval }) => {
...
     onStartMixerAttempt,
-    onGetPrivacyLogs
+    onGetPrivacyLogs,
+    allowSendFromUnmixed,
+    toggleAllowSendFromUnmixed
   } = usePrivacy();
 
   const [logs, setLogs] = useState("");
+  const [showingSendUnmixModal, showModal] = useState(false);
+  const onToggleSendFromUnmixed = () => {
```
amassarwi, line 24: useCallback here might be useful

app/components/views/PrivacyPage/Privacy/Privacy.jsx
```diff
@@ -14,10 +14,27 @@ const Privacy = ({ isCreateAccountDisabled, setInterval }) => {
...
+    if (showingSendUnmixModal) {
+      toggleAllowSendFromUnmixed();
+      showModal(false);
+    }
+  };
+  const onChangeCheckbox = () => {
```
amassarwi, line 30: same here

amassarwi (approved):
LGTM added one suggestion 👍

app/components/views/PrivacyPage/Privacy/Privacy.jsx
vctt94: it will be a callback with multiple function methods on its array dependency, which will be re-rendered each update. That's why I didn't add

app/components/views/PrivacyPage/Privacy/Privacy.jsx
vctt94: same

victorgcramos (changes requested):
Good job! Working as expected. Just a few minor visual issues that I noticed:

- modal background should be a little darker
- components and files with SeedCopyConfirmModal naming

Screen Shot 2020-11-11 at 5 16 56 PM

app/components/views/PrivacyPage/Privacy/Privacy.jsx
amassarwi: hrrmmm i'm not sure if i understand, useCallback will prevent extra re-renders in children components (the one receives `onToggleSendFromUnmixed`) it will guarantee to keep the same pointer to functions unless one of the deps array updates, in that case it will re-evaluate the useCallback value which will result in new pointer and children components wil re-render, without useCallback each re-render in parent will cause re-evaluation of the function which means new pointer which means children will re-render as well.

app/components/views/PrivacyPage/Privacy/Privacy.jsx
amassarwi: see above comment

app/components/views/PrivacyPage/Privacy/Privacy.jsx
vctt94: Functions are comparing by reference, so if I add a function at the array dependency, it will not have much effect. Here: https://hackernoon.com/grasp-by-value-and-by-reference-in-javascript-7ed75efa1293

So I see no reason for adding it. We will not increase performance and the method will be more verbose.

amassarwi: oh i see, didn't notice these are functions.

app/components/modals/SendFromUnmixedAccountModal/SendFromUnmixedAccountContent.jsx
```diff
@@ -0,0 +1,65 @@
...
+  show,
+  onCancelModal,
+  onSubmit,
+  copyConfirmationPhrase,
+  typedConfirmationPhrase,
+  onTypedConfirmationPhraseChanged
+}) => (
+  <DefaultModal className={style.confirmSendFromUnmixed} {...{ show }}>
+    <div className={style.titleConfirmSendFromUnmixed}>
+      <T id="SendFromUnmixed.titleWarning" m="Seed Clipboard Copy Warning" />
```
jholdstock, line 19:
```suggestion
      <T id="SendFromUnmixed.titleWarning" m="Sending from Unmixed Accounts" />
```

app/components/modals/SendFromUnmixedAccountModal/SendFromUnmixedAccountContent.jsx
```diff
@@ -0,0 +1,65 @@
...
+    </div>
+    <div className={style.confirmSeedCopyToolbar}>
+      <DangerButton
+        className={style.confirmConfirmButton}
+        onClick={onSubmit}
+        disabled={
+          typedConfirmationPhrase.toLowerCase() !==
+          copyConfirmationPhrase.toLowerCase()
+        }>
+        <T id="SendFromUnmixed.btnConfirm" m="Allow send from Unmixed" />
```
jholdstock, line 54:
```suggestion
        <T id="SendFromUnmixed.btnConfirm" m="Enable sending from unmixed accounts" />
```

app/components/views/PrivacyPage/Privacy/PrivacyContent.jsx
```diff
@@ -60,6 +66,22 @@ const PrivacyContent = ({
...
+    <SendFromUnmixedAccountModal
+      show={showingSendUnmixModal}
+      onSubmit={onToggleSendFromUnmixed}
+      onCancelModal={() => showModal(false)}
+    />
+    <Checkbox
+      label={
+        <T
+          id="Privacy.checkbox"
+          m="Send From Unmixed accounts"
```
jholdstock, line 78:
```suggestion
          m="Send from unmixed accounts"
```

app/components/views/PrivacyPage/Privacy/PrivacyContent.jsx
```diff
@@ -60,6 +66,22 @@ const PrivacyContent = ({
...
     */}
+    <SendFromUnmixedAccountModal
+      show={showingSendUnmixModal}
+      onSubmit={onToggleSendFromUnmixed}
+      onCancelModal={() => showModal(false)}
+    />
+    <Checkbox
+      label={
+        <T
+          id="Privacy.checkbox"
```
jholdstock, line 77: I think this is the only string ID I have seen starting with a capital letter. We can give it a more descriptive name too, in case we add another checkbox later.

```suggestion
          id="privacy.sendFromUnmixedCheckbox"
```

app/i18n/docs/en/InfoModals/SendFromUnmixed.md
```diff
@@ -0,0 +1,3 @@
+## Allow Send from other unmixed accounts
```
jholdstock, line 1: I think we can remove this heading, because the modal already has one. This one isn't rendered properly anyway, it just appears as a bold line at the start of the paragraph.

app/i18n/docs/en/InfoModals/SendFromUnmixed.md
```diff
@@ -0,0 +1,3 @@
...
+
+Currently, you have privacy settings enabled.  As part of this, we have turned off the ability to create external spends from non-mixed accounts.  Any funds sent from the non-mixed account will not be private and will be able to be tracked.  If you would like to turn off this setting please type in 'I understand the risks' below.  You may re-engage this setting at any time.
```
jholdstock, line 3:
```suggestion
Sending from unmixed accounts has been disabled by the privacy settings of this wallet. This is for your protection.

Sending from unmixed accounts is not private and could possibly be traced back to you.  If you would like to turn off this setting please type 'I understand the risks' below.  You may re-engage this setting at any time.
```

jholdstock (changes requested):
- I just loaded up a wallet which already has privacy enabled and the checkbox was not checked. Perhaps it is using the wrong default value when no value is found in the config file?

- If I press "Cancel" in the modal, the checkbox value changes anyway

alexlyp (approved):

jholdstock (approved):

PR #2908, opened 2020-11-11
Improve disabled `plus` button bg color in send funds

amassarwi:
After:

image

Closes #2883

PR #2911, opened 2020-11-11
Add `cursor: not-allowed` to `plus` button in send funds

amassarwi:
This commit adds a tiny css nit which was missed in #2908 - showing `not-allowed` cursor to give a better ux for
a disabled button.

Screenshot:

image


vctt94 (approved):
LGTM

PR #2929, opened 2020-11-12
Fix: Transfer tx showing wrong receiver account

### LINKED ISSUE #2919: transfer tx can show wrong account
If making a transfer tx between accounts, decrediton can show it as 

image 

even though it was a transfer for other account and it just had change to default

amassarwi:
This diff fixes an issue with transfer fee tx showing wrong receiver account, as these txs may have multiple
outputs, this commit picks the account which corresponds to the first not change output.

---

Closes #2919

### DIFF
--- a/app/selectors.js
+++ b/app/selectors.js
@@ -604,7 +604,7 @@ export const transactionNormalizer = createSelector(
         const address = credit.getAddress();
         const creditedAccount = credit.getAccount();
         const currentCreditedAccountName = getAccountName(creditedAccount);
-        // If we find crediton account differnt than default, then we pick 
+        // If we find crediton account differnt than default, then we pick
         // it as receiver
         if (currentCreditedAccountName !== "default" || !creditedAccountName) {
           creditedAccountName = currentCreditedAccountName;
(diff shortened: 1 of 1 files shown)


app/selectors.js
```diff
@@ -593,9 +603,14 @@ export const transactionNormalizer = createSelector(
         const amount = credit.getAmount();
         const address = credit.getAddress();
         const creditedAccount = credit.getAccount();
-        creditedAccountName = getAccountName(creditedAccount);
+        const currentCreditedAccountName = getAccountName(creditedAccount);
+        // If we find crediton account differnt than default, then we pick
+        // it as receiver
+        if (currentCreditedAccountName !== "default" || !creditedAccountName) {
```
vctt94, line 609: I believe this shouldn't be hard coded for default. Changing the account name, from default -> "whatever" would break it, or using a different account

vctt94 (changes requested):

app/selectors.js
amassarwi: updated

vctt94 (approved):
LGTM!

xaur:
Depending on how "change output" is determined, this may or may not break with randomized UTXO selection added in https://github.com/decred/dcrwallet/pull/1914

PR #2931, opened 2020-11-13
Revert "TrezorActions: Allow updating firmware. (#2816)"

alexlyp:
This reverts commit bf54bf8bfdf5214f6aa61a5fd121b50de2a95004.

This commit was causing build errors that were not easily fixed so 
we're revert until the cause can be tracked down and fixed.


matheusd (approved):
tested yarn build && yarn start

PR #2932, opened 2020-11-13
TrezorActions: Allow updating firmware v2.

JoeGruffins:
We are unable to update firmware from TrezorActions.js. This is believed
    to be due to the size of the firmware and the limitations of this
    electron child process. We send the firmware to the main process to push
    to the device. The main process uses a temporary connection to trezor in
    order to accomplish this.
    
    Send path to ipc.
    
    Better handle update firmware failures.

    Move connect to app/node_modules.
    
    For some reason importing connect in main_dev/ipc.js was not working
    when building for release. Moving connect fixes it...

### DIFF
--- a/app/actions/TrezorActions.js
+++ b/app/actions/TrezorActions.js
@@ -1,6 +1,5 @@
 import * as wallet from "wallet";
 import * as selectors from "selectors";
-import fs from "fs";
 import { hexToBytes, str2utf8hex } from "helpers";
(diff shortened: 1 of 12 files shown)


alexlyp (approved):
Looks like building is working again.

matheusd (approved):
Tested updating the firmware on a model one and `yarn build && yarn start`

PR #2934, opened 2020-11-13
[shared] Add Log component

### LINKED ISSUE #2899: Create logs shared component
Now that we are showing logs on privacy page, we have 2 css which looks pretty much the same for showing logs and there is a lot of code repetition  on the logs page. 

We can have a shared component instead, used in both places with its own css modules.

amassarwi:
This diff adds a new component `Log` to _shared_ components.
 - Create new Log component.
 - Fix broken `daemonRemote` redux flag, which caused a bug where `dcrd` logs where displayed in remote daemon mode.
 - Replace duplicated logs in help page with new component.
 - Use log component in privacy section.
-  Hide logs section in privacy page by default and show it only if mixer is on, before this commit
logs were empty while mixer is off and occupied big section on the privacy page. 
 - Delete old `privacy.less` file. 

--- 

Closes #2899


app/components/views/HelpPage/LogsTab/Page.jsx
```diff
@@ -26,95 +24,42 @@ const Logs = ({
...
-            <T id="help.logs.dcrd" m="dcrd" />
-          </div>
-          <div className={styles.logs}>
-            <textarea rows="30" value={dcrdLogs} disabled />
-          </div>
-        </div>
-      )
-    ) : (
-      <div />
+    {!isDaemonRemote && (
```
jholdstock, line 28: I dont think this is working properly, I am seeing an empty dcrd log even when I am connected to a remote dcrd.

amassarwi: :eyes:

amassarwi: checking

amassarwi: I'm seeing the same on master - will fix that in this PR.

amassarwi: fixed 🧘‍♂️

jholdstock:
dcrd remote daemon looks good now. I notice one more thing though. 

When the mixer slider is switched on, the logs panel appears and works correctly, but when the mixer is switched off, the logs disappear. I expect them to stay on the screen after the mixer is off so they still can be read.

amassarwi:
@jholdstock This was intended i thought no need to show the logs is mixer is off, I agree, this is problematic when you turn off the mixer and you still want to see the logs, will revert that.

amassarwi:
@jholdstock updated to always show logs, but be default it will be collapsed and on toggle click it will get expanded.

jholdstock (approved):
All working well now :+1:

jholdstock (changes requested):
Oops, sorry, I need to remove my approval. Just noticed quite a severe issue which seems to be caused by this PR, my accounts have disappeared...

Screenshot from 2020-11-15 09-39-38

amassarwi:
@jholdstock Updated mate.

> https://github.com/decred/decrediton/pull/2934/files#diff-63bd92c374eb5875f8ca16487eaa01985ceef61478e36e2bd32b14ac0666440aL1-L3 this caused the broken select
we had global select styling in privacy.less file which affected the Select used to select account in places you mentioned - i'm very happy we are moving away from .less and global styling conflicts & mishmash, this kind of styling are impossible to maintain and pretty unpredictable - i moved the styling to ReactSelectGlobal.css which is used for that purpose now!

jholdstock (approved):
Looks good!

vctt94 (approved):
Nice improvement. Thanks!

PR #2935, opened 2020-11-13
[LN] Improve connection procedure

matheusd:
This makes the connection procedure to the underlying dcrlnd node services more
reliable by trying again when an EUNIMPLEMENTED error is returned.

Due to the gRPC server being restarted after an unlock, depending on the exact
timings, we might still connect to the wallet unlocker service so trying to
use the general services would fail.

While this isn't usually an issue today, it will be needed when switching to the grpc-js library in the future.

### DIFF
--- a/app/actions/LNActions.js
+++ b/app/actions/LNActions.js
@@ -284,18 +284,46 @@ const connectToLNWallet = (
     return { client };
   }
 
-  const lnClient = await ln.getLightningClient(
-    address,
-    port,
-    certPath,
-    macaroonPath
-  );
-  const wtClient = await ln.getWatchtowerClient(
-    address,
-    port,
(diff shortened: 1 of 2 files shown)


jholdstock:
Testing this with `grpc-js` I am still getting connection dropped errors:

```
Failed to unlock LN wallet: Error: 14 UNAVAILABLE: Connection dropped
```

matheusd:
Please try again with the latest push and this branch of dcrlnd: https://github.com/matheusd/dcrlnd/tree/sleep-walletsvc-stop

jholdstock (approved):
Running with the sleep on dcrlnd fixes grpc-js.

alexlyp (approved):

PR #2937, opened 2020-11-14
[startup] Filter incomplete setup wallets in wallet selection

### LINKED ISSUE #2920: Cannot complete wallet setup
- Open decrediton in SPV mode
- Click Create a new wallet
- Whilst on the first screen with the seed displayed, close Decrediton
- Reopen Decrediton
- Wallet with "Setup Incomplete" is displayed. Clicking on it shows an RPC error "Wallet has not been loaded"

This also happens when not on SPV mode, but the error message is not displayed

amass01:
This diff fixes #2920 by filtering out incomplete setup wallet in available wallets list.

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -250,7 +250,8 @@ export const getAvailableWallets = () => (dispatch, getState) =>
 
     get()
       .then(({ availableWallets, previousWallet }) => {
-        dispatch({ availableWallets, previousWallet, type: AVAILABLE_WALLETS });
+        const setupCompletedWallets = availableWallets.filter((w) => w.finished);
+        dispatch({ availableWallets: setupCompletedWallets, previousWallet, type: AVAILABLE_WALLETS });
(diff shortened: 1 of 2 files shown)


vctt94 (changes requested):
I don't think the wallet which are incompleted should be filtered. It should allow to complete its setup, if wanted.

amassarwi:
@vctt94 Right, i need to cleanup incomplete wallet on shutdown tho, as we do don't have access to the seed after a shutdown and we can't validate it if user restarted decrediton and tried to restore his wallet.

Sounds doable to you ?

vctt94:
Incomplete setup shows when the directories for the wallet are created, but it doesn't have a wallet.db file. 

So for finishing the setup it needs to create a new wallet, which will generate a new seed.

amassarwi:
I see, will do

alexlyp:
Going to complete this in an alternative PR.  I think we need to let people complete the set up instead of just not showing them the incomplete ones.

PR #2938, opened 2020-11-14
[VSP] show vsp tooltip on tickets `use Legacy` checkbox only if it is not checked

### LINKED ISSUE #2933: wrong tooltip message on tickets use Legacy checkbox
When the legacy checkbox is checked, if hovering it for swtiching to the new vspd it will show this message:

image

It should show only when the checkbox is not checked

bgptr:
Closes #2933


jholdstock (approved):
Works for me and code looks good too

vctt94 (approved):
LGTM

PR #2941, opened 2020-11-15
Update monospace font to Source Code Pro

jholdstock:
Two things here, updating the monospace font to Source Code Pro, and also fixing the use of monospace in a couple of places.

I've asked @linnutee on Matrix to clarify if we should be using SCP in decrediton rather than Inconsolata. PR marked as a draft until he confirms.


jholdstock:
Confirmed SCP is the correct monospace font for decrediton. It can be seen in the xd designs provided in #2965

Ready for review

alexlyp (approved):

PR #2942, opened 2020-11-15
[privacy] Improve mixer start error handling

### LINKED ISSUE #2878: [1.6.0-rc2] mixing autobuyer button giving "Account balance too small" error on incorrect passphrase
When I click to turn on autobuyer I get a message about account balance too small if I enter incorrect passphrase

amassarwi:
This diff fixes two issues I found while investigating #2878 (The error shown in the issue body is expected as @RichardRed0x  didn't have any dcr in unmixed account and we check balance before checking provided passphrase):

- When there was an error while starting the mixer the toggle displayed on mode, even tho it failed
- The displayed error was displayed even after a successful retry

Before:
image

After:
image

---

Closes #2878

### DIFF
--- a/app/actions/AccountMixerActions.js
+++ b/app/actions/AccountMixerActions.js
@@ -205,7 +205,8 @@ export const getCoinjoinOutputspByAcct = () => (dispatch, getState) =>
     wallet
       .getCoinjoinOutputspByAcct(walletService)
       .then((response) => {
-        const coinjoinSumByAcctResp = response.wrappers_ && response.wrappers_[1];
+        const coinjoinSumByAcctResp =
+          response.wrappers_ && response.wrappers_[1];
         const coinjoinSumByAcct = balances.reduce(
           (allAccts, { accountNumber }) => {
             // if account number is equals imported account, we ignore it.
--- a/app/components/views/PrivacyPage/Privacy/Privacy.module.css
+++ b/app/components/views/PrivacyPage/Privacy/Privacy.module.css
@@ -1,6 +1,3 @@
-.error {
-  color: var(--error-text-color);
-}
 .isRow {
   display: flex;
   flex-direction: row;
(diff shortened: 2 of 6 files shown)


jholdstock (commented):
This works well and the code looks good, I just have one minor nitpick.

When I enter an incorrect pasphrase, the slider activates and goes blue for a moment before the snackbar error is displayed. Can this be prevented?

Peek 2020-11-16 09-19

amassarwi:
Regarding this:
> When I enter an incorrect pasphrase, the slider activates and goes blue for a moment before the snackbar error is displayed. Can this be prevented?

As discussed on matrix, it's the related to the stream we use:
```
runMixerAsync()
      .then((resp) => {
        const { mixerStreamer, error } = resp;
        // we can throw errors, like when the account has a small balance,
        // so this check is necessary.
        if (error) {
          dispatch({ error: error + "", type: RUNACCOUNTMIXER_FAILED });
          return;
        }
        mixerStreamer.on("data", () => resolve());
        mixerStreamer.on("error", (error) => {
          // if context was cancelled we can ignore it, as it probably means
          // mixer was stopped.
          if (!String(error).includes("Cancelled")) {
            dispatch({ error: error + "", type: RUNACCOUNTMIXER_FAILED });
          }
        });
        mixerStreamer.on("end", (data) => {
          // not supposed to get here, but if it does, we log to see.
          console.log(data);
        });
        dispatch({ type: RUNACCOUNTMIXER_SUCCESS, mixerStreamer });
      })
      .catch((error) =>
        dispatch({ error: error + "", type: RUNACCOUNTMIXER_FAILED })
      );
```

we dispatch `RUNACCOUNTMIXER_SUCCESS` once, and then in the `error` handler we receive an error and dispatch `RUNACCOUNTMIXER_FAILED`

we decided to improve that after 1.6

jholdstock (approved):
Great, we can fix it later. Approved.

vctt94 (commented):
I like this changes, but I think it would be good keep showing the error in case a user misses the notifcation.

vctt94 (approved):
LGTM!

PR #2943, opened 2020-11-15
fix: Handle tickets without `ticket` tx and fallback to `spender`

### LINKED ISSUE #2769: VSP-fee wallets aren't loading
By "VSP-fee wallets" I mean wallets that are used to receive the ticket fees from the legacy VSP mechanism. Trying to load a wallet with such transactions triggers a "cannot read property getHash of undefined"

amassarwi:
Closes #2769


vctt94 (approved):
LGTM

PR #2947, opened 2020-11-16
[Purchase ticket] fix missing mixedAccountBranch selector

amassarwi:
This diff fixes an issue caused by recent change which deleted the `getMixedAccountBranch` selector and used `0` as mixed account branch.  This commit re-adds the selector and uses it when calling the wallet.


alexlyp (approved):
tACK

PR #2958, opened 2020-11-17
[sidebar] Declutter nav menu

### LINKED ISSUE #2939: Declutter nav menu
Suggestion from @linnutee 

https://xd.adobe.com/view/e51f9efc-bb29-48b2-9158-d41870429ecf-b04c/
> The navigation has gotten overloaded. Since there’s LN and in some cases Trezor added, that sums up to 10 items. I would propose to drop Help and Settings from the main listing and place those into the bottom area as only an icon. Replaced settings icon with the “gear wheel” metaphor as that’s more known. Alternatively the subpages under Help could be placed under Settings to have only 1 one icon there? Bit less loaded this way.

> I’ve made a combined component for the peer count and block sync. This is accompanied by the block icon, if synced has a checkmark next to it, if not has X and during rescan can display the block loop animation https://lottiefiles.com/20515-block-loop … Peercount can be right below this. SPV I’m not sure what’s the reasoning on having there, as well is really odd man out when it comes to the component/placement. If anyone can explain that, I’d have more insight how or where to work in the indicator. Doesn’t necessarily need to be even the menu imo. Alternatively, all three items (Blocksync, peercount and SPV/Non-spv can be a click-through component in the bottom left (ie. only one is shown at a time and others are shown on clicking through them as sort of carousel … can prototype an example).

amassarwi:
This diff implements the `SideBar` design changes suggested on #2939.

- Move settings page to be a new tab under old HelpPage & then renamed it to SettingsPage.
- Tweak RescanButton with new synced icon and syncing animation and create it's css module.
- Add SpvIcon component to show both on & off SPV mode icons and use it both expanded and collapsed menu components
- Add SettingsIconLink and use it in both collapsed and expanded menu components.
- Add two theme vars for menu active item background as requested in #2939.
- Cleanup old help page files and assets.

Todos:

- [x] Add syncing animation without alpha to make fir with all different background colors.
- [x] Fix tests.
- [x] Delete all existing active menu item animations as we would like to avoid having mixed of animated and
not animated menu items.

Screenshots:

image

image

---

Closes #2939


app/components/buttons/RescanButton/RescanButton.jsx
```diff
@@ -8,14 +10,16 @@ Rescanning may help resolve some balance errors.
 Note: This scans the entire blockchain for transactions,
 but does not re-download it.`;
 
+
```
amassarwi, line 13: not needed

vctt94 (commented):
Good job on this PR amass!

I have some small nits, but it looks pretty good overall. 

 - When clicking on a sidebar menu, it is possible noticing the menu caret doing the motion, while the hover effect just appears on the menu.

 - the mixer spinner is always showing the same message, even if I am purchasing tickets or turned on the auto buyer:

image

 - I think this sun icon on the bottom could have a tooltip:
 At first I was thinking it was for turning the theme light.

image

  - Some issues when the sidebar is on the bottom: 

image

alexlyp (approved):

PR #2961, opened 2020-11-18
Various string updates

jholdstock:
- Update Trezor page title and description
- Capitalize Trezor
- "LN" => "Lightning" for LN page title 
- Shorten wrong passphrase error
- Fix capitalization on Privacy zero balance error


matheusd (commented):
Don't change the non-default translation files. They are going to get overwritten by the next update from transifex (which is needed due to the de/pl language work being done right now)

jholdstock:
All non-English files have been reverted.

matheusd (approved):

PR #2487, opened 2020-05-17
Receive view improvements

guilhermemntt:
Closes https://github.com/decred/decrediton/issues/2450.

This PR ignores the share button, since Electron does not support this native functionality (see https://github.com/decred/decrediton/issues/2450#issuecomment-626497410). By the way, the amount input has been kept, so feel free to close it, due the fact the implemented PR is very different from the provided specs.

PR #2962, opened 2020-11-18
Remove LN binaries

jholdstock:
These shouldnt exist in the decrediton repo, they should be added to the `bin/` directory as per dcrd/dcrwallet

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,19 +88,20 @@ To ease node version management, install all top-level tools (node/npm/yarn) usi
 ### Decred binaries
 
 Development using the master version of decrediton usually requires using a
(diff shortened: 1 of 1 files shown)


matheusd (approved):

vctt94 (approved):

xaur:
These binaries added 75 MiB to the repo and were first introduced on Jun 15 commit 49708584 that rebased #2487 on latest master. The PR had 3 reviewers and was merged Jun 29. Then the binaries went unnoticed for another 5 months.

This reminds me of excessive data downloads in #2166 and https://github.com/decred/politeiagui/issues/2199 .

I think we have a problem somewhere.

PR #2967, opened 2020-11-18
[tx] avoid race condition in tx send view

### LINKED ISSUE #2950: Sending DCR in Transactions Tab is consistently incorrect 1.6RC3
image
image

See images. I input 200 DCR and it thinks I want to send 20. I've tested this a few times and found it to be oddly consistent with getting the amount I want to send incorrect.

bgptr:
Closes #2950 
I could reproduce issue #2950 in win10 in VirtualBox. I found that sometimes responses for `constructTransaction` requests don't arrive in the same order as they requested.
My solution is to save the last request and cancel it before the next one if it is necessary.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -572,13 +572,13 @@ export const constructTransactionAttempt = (
     request.setChangeDestination(outputDest);
   }
 
-  let { constructTxCall } = getState().control;
-  if (constructTxCall) {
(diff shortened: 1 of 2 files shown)


vctt94 (commented):
Ack.

Tested this PR with @bgptr inbox, but I am not able to reproduce this issue, although I think the code looks fine.

alexlyp (approved):

Issue #2971, opened 2020-11-19
(1.6 RC3) Request: Add vspd Stakepool Fee % When Purchasing Ticket


exitus1:
The Legacy VSP system displayed the fee right in the wallet: 
![image](https://user-images.githubusercontent.com/46197055/99623487-269dd800-2a24-11eb-8919-3740cb537215.png)

Currently there is no fee mentioned or shown when purchasing with vspd.

jholdstock:
`/vspinfo` includes the fee percentage so this should be doable https://github.com/decred/vspd/blob/master/docs/api.md#get-vsp-info

Decrediton should check the fee regularly as it is liable to change.

guilhermemntt:
On it. Well observed. I'll try to find a nice place to position the same icon, but I'm accepting suggestions too.

exitus1:
> 
> 
> On it. Well observed. I'll try to find a nice place to position the same icon, but I'm accepting suggestions too.

I'm not sure if this is possible, but maybe have the fee % next to the VSP's name?

![image](https://user-images.githubusercontent.com/46197055/99750118-b51f6180-2ad7-11eb-9e5a-eaadb460c73c.png)

guilhermemntt:
Yes, maybe something like a tooltip.

jholdstock:
@guilhermemntt  My comment above can be ignored - the fee percentage is actually available from dcrwebapi on `/?c=vsp`, so there is no need to hit each VSP API directly.

PR #2985, opened 2020-11-24
Add selected vsp fee

guilhermemntt:
This PR adds an icon displaying selected VSP fee and tooltips when hovering over VSP options:

image

Closes https://github.com/decred/decrediton/issues/2971.


vctt94 (commented):
Thanks for this PR, I think a tooltip is a good idea!

Can you make the time for showing the tooltip faster? there is a gap between hovering and showing it. 

Also, the tooltip only shows if hovering the text. Is it possible make it show anywhere the select option?

guilhermemntt:
I've just implemented the requested changes @vctt94. Thanks for the review!

alexlyp (approved):

PR #2987, opened 2020-11-24
Privacy UI/UX improvements

### LINKED ISSUE #2965: Privacy UI/UX improvements
Posting an updated version of the design concept for using privacy. 

Bg:
The goal for users is to create anonymity to their coins, as easily as possible. This should happen in a single view, esp. if we have a dedicated view for Privacy. If I understood correctly, to do this, we should only need a way to turn on the mixer, and a transaction to self component, to send funds to unmixed account as a bare bones solution. How these components make up the interaction flow is what to solve.

Ver 1: https://xd.adobe.com/view/58d3e580-9348-4ee6-b132-2adb97a8ac90-0d80/
Mixer is activated by the large toggle as it is now. Unmixed and Mixed Account Balances are shown for feedback and additional context to explain whats happening. Send-to-self contains a locked setting "To: Unmixed" account. Settings are tucked away in info icon/modal. Dev access to change these?

Cons from basic testing: the positioning logic of the items could be better based on some of the feedback. If the user needs to send funds to Unmixed first, this creates an odd backward movement, going from bottom right back to top left. It was noted that a toggle that requires password entry as a follow-up feels odd. The toggle gives no feedback if it can be started or not (in case unmixed has no funds).

Ver 2: https://xd.adobe.com/view/260c88d6-8b96-4cf3-abcb-cb803eaf631e-735f/
To make this more like a step by step action, and address the positioning issue, i've created to isolated components for each interaction. Both have buttons rather than having a toggle for the mixer. 

If the Unmixed account has too low funds, the button should be disabled. If the user clicks the disabled button, a snackbar warning is displayed. Along side that we can use the animated alert icon, which can pop over the right corner of the unmixed account icon. https://lottiefiles.com/20274-alert-open (color red).

If the mixer is running, we can also use a pre-existing animation of this decentralized loop for the time being: https://lottiefiles.com/20387-decentralized-loop (colors: #2970FF #70CBFF)

Please post any feedback here, also whether this could be worked into 1.6 or should be further iterated.

bgptr:
Closes #2965

Screenshot_20201124_162016
Screenshot_20201124_162027
Screenshot_20201124_162113
Screenshot_20201124_162141
Screenshot_20201124_161825


alexlyp:
Looks really good.  Huge improvement.  

Only nit I can see from testing:  updating the mixed/unmixed account balances when transactions sent across.  I have to click to another menu item and back before they update.

app/components/modals/MixerSettingsModal/MixerSettingsModal.jsx
```diff
@@ -0,0 +1,29 @@
+import { KeyBlueButton } from "buttons";
+import { FormattedMessage as T } from "react-intl";
+import Modal from "../Modal";
+import style from "../Modals.module.css";
+import { classNames } from "pi-ui";
+import { PrivacyForm } from "shared";
+
+const MixerSettingsModal = ({ onCancelModal, show }) => {
+  return (
```
amassarwi, line 9: extra return

app/components/shared/SendTransaction/Form.jsx
```diff
@@ -23,50 +22,55 @@ const SendPage = ({
...
             </div>
           </div>
         </div>
-      </div>
+      )}
     </div>
     <div className={styles.sendButtonArea}>
       {insuficientFunds && (
-        <div className="error">
+        <div className={classNames("error", styles.error && styles.error)}>
```
amassarwi, line 64: why do we still need the `error` string class here ?

app/components/views/PrivacyPage/Privacy/PrivacyContent.jsx
```diff
@@ -1,15 +1,20 @@
...
 import {
   InfoDocModalButton,
   MixerPassphraseModalSwitch,
-  MixerSwitch
+  MixerSettingsIconButton,
+  DangerButton
 } from "buttons";
 import { classNames, Checkbox } from "pi-ui";
 import { SendFromUnmixedAccountModal } from "modals";
 import style from "./Privacy.module.css";
```
amassarwi, line 13: delete empty line

app/components/inputs/ReceiveAccountsSelect.jsx
```diff
@@ -29,7 +41,11 @@ function ReceiveAccountsSelect({ onChange, className, showAccountsButton }) {
         onChange: onChangeAccount,
         accountsType: "visible",
         filterAccounts: [mixedAccount],
-        account
+        account:
+          account && account != nextAddressAccount.value
```
vctt94, line 45: I think a comment here would be good

amassarwi (commented):
Looks good, left some minor comments

davecgh:
I would suggest "Allow sending from unmixed account" for the checkbox text instead of the current "Send from unmixed accounts".  The existing text reads to me like you're instructing the software to "send from unmixed account" instead of specifying that the user wants to allow the possibility of doing it.

vctt94 (approved):
LGTM. Left one small comment, but overall a really good job.

PR #2998, opened 2020-11-25
Mixer: header should use actual account names

### LINKED ISSUE #2995: Mixer header should use actual account names
Account names are currently hard coded. 

https://github.com/decred/decrediton/blob/9535f500c30b2df6466a9022a43d8d91521f1b07/app/components/views/PrivacyPage/PrivacyPage.jsx#L14-L25

These should be dynamic because they won't necessarily always be named "mixed" and "unmixed". Also, the " Account" part of the name should not be part of the hightlighted blue box. I have created an example below.

Screenshot from 2020-11-25 09-45-15

bgptr:
closes #2995

### DIFF
--- a/app/components/views/PrivacyPage/PrivacyPage.jsx
+++ b/app/components/views/PrivacyPage/PrivacyPage.jsx
@@ -7,32 +7,37 @@ import { usePrivacyPage } from "./hooks";
 import style from "./Privacy/Privacy.module.css";
 
 const PrivacyPageHeader = React.memo(
-  ({ mixedAccountName, changeAccountName }) => (
-    <StandaloneHeader
-      iconClassName="security"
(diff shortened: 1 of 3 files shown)


jholdstock (changes requested):
This still isnt correct because the account names are still hard coded. They could be named something other than "mixed" and "unmixed"

jholdstock (approved):
Very nice. 

When account names are set:
Screenshot from 2020-11-25 13-52-51

when they are not set:
Screenshot from 2020-11-25 13-53-18

alexlyp (approved):

amassarwi (approved):

jholdstock (approved):
Latest changes all look good

PR #2999, opened 2020-11-25
Improve mixer animations

jholdstock:
- Replace stills with .svg files (previously .gif)
- Smooth the transition from still to animated by removing transparent borders from images (also removes the need to offset the images by -2px)
- Give gifs a background color rather than transparency. Necessary to prevent jaggies.


bgptr:
Very nice!

vctt94 (approved):
Looking good! Thanks jhold

amassarwi (approved):

alexlyp (approved):

PR #3001, opened 2020-11-25
[privacy] fix: fetch spendable account balances only once

amassarwi:
This diff fixes too many getbalance requests being sent when visiting privacy page by using
`useMountEffect` hook to call the `getMixerAcctsSpendableBalances` action only once - on mount.


jholdstock (approved):
lgtm

alexlyp (approved):

PR #3003, opened 2020-11-25
[transaction] clean tx info on unmount

### LINKED ISSUE #2990: transactions info is not being cleaned
How to reproduce:

 - go to transactions page
 - sucesufully construct a transaction, without sending it
 - now it will possible see the tx info on transaction details
 - go to other view and back to transactions
 
now the info will still show, even though the data was cleaned:

image

amassarwi:
Closes #2990


jholdstock (approved):
Works for me!

alexlyp (approved):

Issue #2964, opened 2020-11-18
Transaction and Accounts icons


linnutee:
Worked out some new icons regarding the privacy and ln features being added. These are the ones for Accounts and Transactions. 

Mixed Account uses shades derived from the privacy/incognito icon combined into the account icon. 
Unmixed Account uses a footstep symbol as a reference that these coins may be traced. 
LN Account uses the lightning symbol. 

For transactions there's two additional ones: 
Privacy related transactions also using the shades derived from privacy/incognito. 
All transactions which are sent within the wallet, from one account to another by the user show the grey user icon.

---

I'd also bring up a rather old idea of implementing either a set of slightly different icons to randomly assign to normal accounts, or some implementing a generative approach (ie. think of identicon generators) for those. Since we got a few fundamentally different accounts already, it would be more useful now to create visual differentiation for normal accounts also. Simplest take would be changing the colors of the default account icon details. If anyone is up for building that lmk.

Preview: https://xd.adobe.com/view/a3448576-c6f8-4ea5-bde5-f5fd07624bc4-f617/
Package: [decrediton - icons - 181120.zip](https://github.com/decred/decrediton/files/5561544/decrediton.-.icons.-.181120.zip)

amassarwi:
can do after #2958

amassarwi:
#3012 adds the transaction icons, account icons will be done in a separated PR.

PR #3012, opened 2020-11-27
[transaction] `self` & `mixed` icons

### LINKED ISSUE #3019: [privacy] Mixed tx not showing `is mix` indicator
As noted by @vctt94 some mixed tx(s) not showing mixed indicator.

image

This one to be more specific: https://testnet.decred.org/tx/d94b5d25e6b94aedc50ea2a29b17ce80d77d27d18cc4c9945edefce904e11d46

amassarwi:
This diff replaces `is mixed` text indication with new icon as specified in the specs for mixed transactions.
In addition, it defines new type - `self` transaction & adds it's icon.

While parsing wallet transactions we mark a tx as `self` transaction when:
```
// If we find in the transaction credits list a self credited account which isn't a change output
// & tx has one or more wallet inputs & no non-wallet outputs we consider
// the transaction as self transaction
``` 

Specs:

image

---

Part of #2964 
Closes #3019


vctt94 (commented):
Thanks for this PR amass, looking good! 

I think it would be good if we add tooltips on the icons, because it can be confusing now with so many icons. 

Also, when clicking on the tx, it has the old icon, can we change that?

jholdstock:
Tooltip is definitely a good idea!

vctt94:
Also, it looks like one of my mixed txs is showing as a `sent` tx: 

image

This one to be more specific: https://testnet.decred.org/tx/d94b5d25e6b94aedc50ea2a29b17ce80d77d27d18cc4c9945edefce904e11d46

amassarwi:
Thanks for the review guys, can do tooltip and add the icon in the details page.

Regarding the `sent` tx i saw that and was able to reproduce on master (this tx won't have current `is mix` text indicator) it's related to the `isMix` flag logic, maybe we should fix that in a different PR ?

amassarwi:
Will fix https://github.com/decred/decrediton/issues/3019 and then finalized this one

vctt94 (approved):
LGTM. Although the transaction mentioned is still showing as a `sent` tx, but I believe we can have an issue for that and fix in a further PR

PR #3013, opened 2020-11-28
fix: 'toString' of undefined console error

### LINKED ISSUE #3011: Console error: `TypeError: Cannot read property 'toString' of undefined`
image

amassarwi:
This diff handle rare case where treasury unspent balance info not available which caused `getTreasuryBalance` action in `ClientActions.js` to throw: `TypeError: Cannot read property 'toString' of undefined`.
As a fix I added a tiny check to make sure that the treasury info is available before trying to parse it.    

*Note*: I wasn't able to consistently reproduce the error, I had to refresh decrediton several times with `ctrl+r` in different pages/stages and jump between testnet/mainnet for it to show up.

Closes #3011


vctt94 (commented):
Hmm, not sure on how `treasuryInfo` can be null. Maybe a race?

alexlyp (approved):

PR #3021, opened 2020-11-30
[PI] fix never-ending `Enable PI Integration` loader

### LINKED ISSUE #2975: Enable PI Integration loading never stops
When I set up a new wallet, go to governance, and click "Enable Politeia Integration" the PI loading animation never stops spinning (waited 5 minutes) until I click on a different tab or page and then come back.  When I switch pages and come back it comes up immediately.  I'm using v.1.6.0-rc3 on Windows but I think I've seen it on earlier versions also.  I don't see any errors after the button was clicked in the console window or decrediton log.  I can repeat this by unchecking "Politeia" on the settings page, saving, going back to the governance page, and clicking "Enable Politeia Integration" again.  It's not critical since it's rare and easy to work around but also not a great impression for new users.

bgptr:
closes #2975

### DIFF
--- a/app/components/views/GovernancePage/Proposals/ProposalsList/ProposalsList.jsx
+++ b/app/components/views/GovernancePage/Proposals/ProposalsList/ProposalsList.jsx
@@ -16,7 +16,8 @@ const ProposalsList = ({ finishedVote, tab }) => {
     noMoreProposals,
     proposals,
     state,
-    send
+    send,
+    proposalsAttempt
   } = useProposalsList(filterTab);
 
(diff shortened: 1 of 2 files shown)


jholdstock (approved):
Recreated the issue and I can confirm this PR fixes it.

vctt94 (commented):
I am not really a fan of this solution. Adding a loadingView into the success state doesn't seem right to me

vctt94 (approved):
Thanks for fixing it! LGTM

PR #3022, opened 2020-11-30
[Logs] missing dcrlnd logs

bgptr:
I noticed that LN logs in the Logs tab are only visible during the LN start process. After that, they disappear. This diff fixes it.


amassarwi (approved):
lgtm

jholdstock:
Works for me

vctt94 (approved):
LGTM

PR #3023, opened 2020-12-01
Distinguish sends/receives in tx icon tooltips

jholdstock:
The icon for Send is red, for Receive is green, but both had a tooltip which just said "Transaction". 

Now they have "Sent" or "Received".

Peek 2020-12-01 10-37


amassarwi (approved):
ACK

vctt94 (approved):
LGTM

Issue #2907, opened 2020-11-11
Auto-detect mixed account when restoring wallet


jholdstock:
Currently there is a toggle when restoring from seed "Privacy? Yes/no".

After talking with @vctt94 we think this should not be necessary. Decrediton can check accounts for mixed utxos and automatically detect if the wallet has used privacy in the past, and then prompt the user to select their mixed and unmixed accounts.

amassarwi:
On it

PR #3029, opened 2020-12-01
[privacy] fix: Restore mixed wallet with SPV on

amassarwi:
While working on #2907 found that restoring mixed wallet doesn't work with SPV on, this diff fixes that.

### DIFF
--- a/app/components/views/GetStartedPage/hooks.js
+++ b/app/components/views/GetStartedPage/hooks.js
@@ -27,6 +27,7 @@ import SettingMixedAccount from "./SetMixedAcctPage/SetMixedAcctPage";
 // and styling defined in Loading.less and need to handled when loading.less
 // is migrated, and classes should be defined then in ./GetStarted.module.css
 // css animation classes
+const blockChainLoading = "blockchain-syncing";
 const daemonWaiting = "daemon-waiting";
 const discoveringAddresses = "discovering-addresses";
 const scanningBlocks = "scanning-blocks";
--- a/app/components/views/GetStartedPage/hooks.js
+++ b/app/components/views/GetStartedPage/hooks.js
@@ -27,7 +27,6 @@ import SettingMixedAccount from "./SetMixedAcctPage/SetMixedAcctPage";
 // and styling defined in Loading.less and need to handled when loading.less
 // is migrated, and classes should be defined then in ./GetStarted.module.css
 // css animation classes
-const blockChainLoading = "blockchain-syncing";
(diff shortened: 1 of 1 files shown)


app/components/views/GetStartedPage/hooks.js
```diff
@@ -208,9 +208,18 @@ export const useGetStarted = () => {
         if (synced === true) {
           send({ type: "GO_TO_HOME_VIEW" });
         }
-        if (context.isSPV) {
+        if (isSPV) {
           return startSPVSync(passPhrase)
-            .then(() => send({ type: "GO_TO_HOME_VIEW" }))
+            .then(() => {
+              if (isPrivacy) {
+                // if recoverying a privacy wallet, we go to settingMixedAccount
```
jholdstock, line 215:
```suggestion
                // if recovering a privacy wallet, we go to settingMixedAccount
```

app/components/views/GetStartedPage/hooks.js
```diff
@@ -208,9 +208,18 @@ export const useGetStarted = () => {
...
+                // state, so the user can set a mixed account based on their
+                // coinjoin outputs.
+                // This state should only be achievable if recoverying wallet.
```
jholdstock, line 218:
```suggestion
                // This state should only be achievable if recovering wallet.
```

jholdstock:
Looks good other than comment typos

amassarwi:
@jholdstock This comment was deleted in https://github.com/decred/decrediton/pull/3033/files#diff-7703b3a83a5aa0b1e709012b1b720d618adcb22031c163096ac2d99e1af80925L236-L244 along with the `isPrivacy` check, so we can keep it as is in this PR (also to avoid a rebase conflict 🙏)

jholdstock (approved):

app/components/views/GetStartedPage/hooks.js
vctt94, line 218: This state shouldn't be only achievable when recoverying a wallet, as the user may pass it and may want to do it later, but I am approving this PR, as this is true for this release, but we should allow it if one passes it, by mistake or whatever other reason.

vctt94 (approved):
Left a comment, but I am approving this PR.

app/components/views/GetStartedPage/hooks.js
amassarwi: You are right and this was completely deleted in #3033 as it will run on each startup now!

PR #3030, opened 2020-12-01
[transaction] Distinguish `Ticket fee` tx icon tooltip

### LINKED ISSUE #3024: Tooltip for ticket fee transactions
Ticket fee transactions seem to have their own icon - yellow. However they have a generic tooltip message "Transaction".

Peek 2020-12-01 10-42

Can we properly identify these as ticket fees and indicate that in the tooltip?

amassarwi:
Closes #3024


app/components/shared/TxHistory/RegularTxRow.jsx
```diff
@@ -5,17 +5,21 @@ import {
...
   MIXED,
-  SELFTRANSFER
+  SELFTRANSFER,
+  TRANSFER
 } from "constants";
 import styles from "./TxHistory.module.css";
 import { classNames, Tooltip } from "pi-ui";
 
 const iconTooltipByType = (type) => {
+  console.log({ type });
```
jholdstock, line 15: I guess this should be removed

jholdstock (commented):
This works for me, but i'm not entirely sure about the code. Is `TRANSFER` type **always** a ticket fee? If it is, perhaps it should be renamed to `TICKET_FEE` so it is more descriptive.

amassarwi:
@jholdstock great observation - after adding self tx & mixed tx all transfer transactions were actually ticket fee -
so i defined a new const TICKET_FEE and replaced all usage!

jholdstock (approved):
lgtm!

vctt94 (commented):
So I am not sure this is entire correct, as not always a transfer tx (sending from one account to the same account, in this case?) will always be a ticket tx.

It would be more correct checking if it is a split tx, which in this case, it will always be a ticket tx

amassarwi:
@vctt94 hrmmm i think currently that would be considered a selfTx no? will double check that.

vctt94 (approved):
Actually, you are right @amassarwi , my bad. 

After resolving the conflicts, it LGTM

PR #3032, opened 2020-12-01
[get-started] style: improve messaging on restore wallet form

### LINKED ISSUE #3020: Improve Restore Wallet page messaging
Got some UX reports from an average user, not much familiar with the crypto universe:

Screen Shot 2020-11-30 at 11 26 19 AM

1. The *wallet* on  "Restore Existing **Wallet**" is different from the one on "**Wallet** name", which caused an user to ask me if this Wallet Name should be the same as the wallet you are trying to recover.

2. The same user asked me what "Watch Only", "Trezor" and "Privacy" mean.

    a. Although those terms are self-descriptive, I think we should display some tooltip telling the user if the config can be changed further, what does the config do, etc.
    b. We can tell the user that they are all disabled by default. Maybe collapse all those options into an "Advanced Config" accordion.

victorgcramos:
This diff adds tooltips for Watch Only and Trezor, since Privacy will be removed by #3020, and adds an "Advanced Options" label.

Closes #3020 

Screen Shot 2020-12-01 at 6 52 14 PM
Screen Shot 2020-12-01 at 6 52 04 PM

### DIFF
--- a/app/components/shared/Collapse.jsx
+++ b/app/components/shared/Collapse.jsx
@@ -1,0 +1,26 @@
+import { useState } from "react";
+import styles from "./shared.module.css";
+import { classNames } from "pi-ui";
+import PropTypes from "prop-types";
+
+const Collapse = ({ isInfo, header, content }) => {
+  const [show, setShow] = useState(false);
+  return (
+    <div className={classNames(styles.collapse, isInfo && styles.info)}>
+      <div onClick={() => setShow(!show)}>{header}</div>
+      <div
+        className={classNames(styles.readMoreIcon, show && styles.active)}
+        onClick={() => setShow(!show)}
+      />
+      {show && <div className={styles.content}>{content}</div>}
+    </div>
+  );
+};
+
+Collapse.propTypes = {
+  header: PropTypes.node.isRequired,
+  content: PropTypes.node.isRequired,
+  isInfo: PropTypes.bool
+};
+
+export default Collapse;
--- a/app/components/shared/index.js
+++ b/app/components/shared/index.js
@@ -24,4 +24,5 @@ export { default as TabsHeader } from "./TabsHeader/TabsHeader";
 export { default as PrivacyForm } from "./PrivacyForm/PrivacyForm";
 export { default as Log } from "./Log/Log";
 export { default as SendTransaction } from "./SendTransaction/SendTransaction";
+export { default as Collapse } from "./Collapse";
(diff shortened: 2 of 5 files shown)


jholdstock:
I had thought that the "Advanced Options" would be a collapsible panel, which is hidden by default. Just adding a heading doesnt really add much value IMO.

Now that I am seeing it, I'm not sure if tooltip is correct for this. It isnt obvious that the toggle labels can be hovered over. 

Perhaps we could use a light italic text, under the toggle, similar to the LN and settings tabs:

Screenshot from 2020-12-03 14-03-25
Screenshot from 2020-12-03 14-03-36

victorgcramos:
Nice suggestion! I really like it!

victorgcramos:
Screen Shot 2020-12-03 at 1 18 35 PM

@jholdstock what about now?

vctt94 (commented):
I think that's a good improvement. Thanks for this PR.

About watching only wallet, it is true that you can not spend any decred, but you can not do any actions, which would need signing, as a watch only wallet does not know the wallet seed. 

So Also creating account, signing messages, etc. Not sure what would be a good message for that, but I'd like to make it clearer

jholdstock:
I think the current message for watch only is acceptable. We should not spam the user with information on this page. We can add a link to docs.decred.org once decred/dcrdocs#1137 is implemented

jholdstock (changes requested):
We should keep the watch-only message reasonably short so we don't spam the user with information on this page. For more information, we can add a link to docs.decred.org once decred/dcrdocs#1137 is implemented.

How about this:

> A watch-only wallet has limited functionality. It can only be used to view the balance and monitor transaction history. You won't be able to spend any DCR associated with this wallet.

And for the Trezor message:

> Trezor is a hardware wallet. For more information, visit **docs.decred.org**

Linking to https://docs.decred.org/wallets/decrediton/trezor/


Screenshot from 2020-12-07 10-33-14

I'd suggest indenting the "Master pub key" field so it is aligned with the "You'll not be able..." message, or perhaps even indented slighted more than that.

And also, can we limit the width of the "Advanced Options" panel? It seems a bit too wide at the moment. max-width around 600px looks good to my eye.

victorgcramos:
Preview:

Screen Shot 2020-12-07 at 11 27 52 AM

jholdstock (approved):
Great work :+1:

PR #3037, opened 2020-12-02
Add warning of ticket fee errors to close wallet action

### LINKED ISSUE #3025: [1.6.0-rc4] Add warning of ticket fee errors to close wallet action
If the wallet that has tried to purchase a ticket but ended up with a fee error for whatever reason, when trying to close Decrediton, a modal is displayed indicating there are still tickets with the fee unpaid.

However, when trying to close the wallet by way of Settings -> Close Wallet, this warning is **not** displayed.

Ideally, it should also be displayed there.

### LINKED ISSUE #3043: [1.6-rc4] Improve ticket fee error modal
Right now this is the modal we are showing when decrediton still has ticket with fee errors:

image

From @davecgh on matrix:
~~`perhaps: "You still have unpaid ticket fees." For the first part. The second one I'm not entirely sure what it means by "re-sync fees" to offer a better suggestion, but it would at the very least be "Click on this button to re-sync fees".
The title should also be "VSP Ticket Fee Error" imo.  It may or may not be more than one ticket.`~~

~~We also need to change the `confirm` button to `close decrediton`~~

image

Note that: if changing it to `close anyway`, we need to change our other modals, for keeping consistency

bgptr:
Closes #3025
Closes #3043

### DIFF
--- a/app/components/views/SettingsPage/SettingsPage.jsx
+++ b/app/components/views/SettingsPage/SettingsPage.jsx
@@ -57,9 +57,9 @@ const SettingsPageHeader = ({
         modalContent={
           isTicketAutoBuyerEnabled
             ? closeWalletWithAutobuyerModal(walletName)
-            : hasUnpaidFee 
-              ? closeWalletWithUnpaidFeeModal(walletName)
-              : closeWalletModalContent(walletName)
+            : hasUnpaidFee
+            ? closeWalletWithUnpaidFeeModal(walletName)
+            : closeWalletModalContent(walletName)
         }
         className={styles.closeModalButton}
         onSubmit={onCloseWallet}
@@ -69,7 +69,7 @@ const SettingsPageHeader = ({
 );
 
 const SettingsPage = () => {
-  const { onCloseWallet, isTicketAutoBuyerEnabled, walletName, hasUnpaidFee} = useSettings();
+  const { onCloseWallet, isTicketAutoBuyerEnabled, walletName, hasUnpaidFee } = useSettings();
 
   return (
     <TabbedPage
--- a/app/components/views/SettingsPage/SettingsPage.jsx
+++ b/app/components/views/SettingsPage/SettingsPage.jsx
@@ -25,10 +25,19 @@ const closeWalletWithAutobuyerModal = (walletName) => (
   />
 );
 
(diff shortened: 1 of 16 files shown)


amassarwi (approved):

app/components/views/SettingsPage/SettingsPage.jsx
```diff
@@ -25,10 +25,19 @@ const closeWalletWithAutobuyerModal = (walletName) => (
   />
 );
 
+const closeWalletWithUnpaidFeeModal = (walletName) => (
+  <T
+    id="settings.closeWalletModalWithUnpaidFeeModal"
+    m="Are you sure you want to close {walletName} and return to the launcher? You still have unpaid tickets fee. If you proceed and they are chosen to vote, they will be missed."
```
jholdstock, line 31: We probably need some more information here - what should the user do if they see this modal?

- Do they need to press a button somewhere?
- Do they just need to wait a bit longer? - 1 minute? Wait for a block to be mined?

vctt94 (commented):
As jhold already comment about the message here, I think maybe we can also close https://github.com/decred/decrediton/issues/3043 ? 

What do you guys think

app/components/views/SettingsPage/SettingsPage.jsx
alexlyp: Good point, Let's change it to the following.

"Please go back to the ticket status page and press the Sync VSP Tickets button to ensure that all of your ticket fees are paid."

app/components/views/SettingsPage/SettingsPage.jsx
```diff
@@ -17,19 +17,7 @@ const closeWalletModalContent = (walletName) => (
   />
 );
 
-const closeWalletWithAutobuyerModal = (walletName) => (
-  <T
-    id="settings.closeWalletModalWithAutobuyerModal"
```
jholdstock, line 22: I dont think this should have been removed. I am trying to close a wallet with the ticketbuyer running, and I see the action being logged
```
 action     {type: "SHOW_AUTOBUYER_RUNNING_MODAL"}
```
but nothing happens at all. I am not getting any popup modal, and decrediton does not close.

jholdstock (approved):
Tested functionality and its looking good

alexlyp (approved):

PR #3041, opened 2020-12-03
[Privacy] Add "Send all funds" button to the Privacy view

### LINKED ISSUE #3040: [1.6-RC4] Add "Send max" button in Privacy tab
The new privacy tab is awesome!
I would wish for a "send max" button as in the regular Transaction tab. I think most of the time people will just want to mix everything and having to type it in just doesn't feels right. It feels like this button is missing.
When people only want to mix a portion they still can.

image

bgptr:
Closes #3040


jholdstock (commented):
Working well for me!

amassarwi (approved):

vctt94 (approved):
Tested and works fine

Issue #2831, opened 2020-10-30
Performance of "Ticket Status" panel


jholdstock:
Opening the "Tickets" > "Ticket Status" panel on a wallet with hundreds of tickets is not a nice experience.

- Loading takes 1-2 minutes on a very fast modern computer
- When loading is completed, I am getting "No Tickets Found", even when this wallet has hundreds of tickets.
- The rest of decrediton becomes unusable while the list is loading. I would guess this processing is being done on the main thread and it should be offloaded to a background thread to prevent decrediton becoming unresponsive.
- Cancelling the listing works (see below), but the visuals can be improved:
  - The ticket motion (which in this case is acting as a spinner) continues, which gives the impression that loading is still happening. It would be better to show a static ticket, indication loading has stopped
  - The "Resume" button is still a cross - counterintuitive. would be better to show a seperate "Resume" button

![Peek 2020-10-30 08-36](https://user-images.githubusercontent.com/6762864/97678271-7162aa80-1a8b-11eb-85b6-52a7fc04c477.gif)

vctt94:
So right now this screen is showing tickets from the new vsp, so we show its fee status.

I believe we should also show live tickets from old vsps, as right now they can not be seen. History tickets are showing only tickets already voted.

The reason it gets really slow, is because decrediton fetches all transactions by stream and store them in a selector, but as we use infinite scroller, all of them are being rendered on history tickets. If you try to go to history tickets, you will probably see all your voted/missed/revoked tickets (which will also be pretty slow, as decrediton will be renderer all of them at once, because of the infinite scroll.)

I think a solution here would be to remove our infite scroller and gets transactions by batches from dcrwallet, so we can paginate them.

The problem with that, is that right now dcrwallet can not send transactions by batches, and can only do it by stream, iirc.

xaur:
UX-wise infinite scrolling is a better experience than clicking page number buttons. The user just needs some UI clues to be aware of the scrolling position.

It feels like the problem is an asynchrous programming issue and not the lack of batched API in dcrwallet. It should be possible start reading from the stream asynchronously without blocking the UI, and then add items as they come in. If there are lots of items you could batch them in the async reading routine before sending to the UI.

PR #3054, opened 2020-12-07
Add search limit for live tickets

vctt94:
A work torwards #2831.

I am not closing it, as I still think we can do more improvements on it

### DIFF
--- a/app/actions/TransactionActions.js
+++ b/app/actions/TransactionActions.js
@@ -514,7 +514,7 @@ export const getTransactions = (isStake) => async (dispatch, getState) => {
     stakeTransactions,
     regularTransactions
   } = getState().grpc;
-  let { noMoreTransactions, lastTransaction } = isStake
+  let { noMoreTransactions, lastTransaction, noMoreLiveTickets } = isStake
     ? getStakeTxsAux
     : getRegularTxsAux;
   const listDirection = isStake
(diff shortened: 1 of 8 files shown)


matheusd:
Ok, this is definitely an improvement in sorting. And just to confirm: we're not supposed to show expired/missed tickets in the tickets status pane?

vctt94:
No, right now it is only responsible for showing live and immature tickets, so we can show the fee status of the new vsp.

Missed/revoke tickets are still being showed on the `history tickets` tab. 

Not the best naming, but that's how it is being done right now

jholdstock (changes requested):
This works as intended, only the previous `(ticket expiry + coinbase maturity)` blocks are being scanned now. This is a good improvement, but still does not fix most of the issues in #2831.

Issue related to this PR:

- The way progress percentage is calculated hasnt been updated, so it is still calculated down to block 0. This means the list finishes loading at around 7-8%.

vctt94:
Right, but after it reaching the end of the search it will show like this, so I don't think this is a problem

image

jholdstock:
Regardless, a progress indicator which runs from 0-8% is not helpful. It should either run from from 0-100%, or it should be removed.

vctt94:
I am not really willing to change a lot this code piece at this point. 

I agree that it can be improved, but it would require a new selector and code changes which I think can be done for the next release

alexlyp:
Yah let's just remove/hide the progress indicator for now until we have a chance to improve it later.

jholdstock (changes requested):
So now the entire "Loading" page is gone, and the tab just remains blank until the "No tickets" message appears. 

With my comments above I was just referring to removing the Percentage. The ticket animation, the "Loading more tickets..." message, and the "Down to block XX..." message, can all be kept.
Screenshot from 2020-12-10 08-01-33

alexlyp (approved):

PR #2986, opened 2020-11-24
i18n/docs it translations .md files

karamble:
this adds the .md files that are not editable trough transifex in italian language. Needs review by teknico.

PR #3062, opened 2020-12-08
Review of the Italian translation of the docs

teknico:
This improves the files added in #2986.

All Markdown files have been checked against the English source and their rendering previewed.

In addition to semantic, grammar and spelling improvements, this changes the encoding to UTF-8, removes Windows CR characters and wraps lines (for better readability and diffing).


matheusd (commented):
The files in warning/ are also in transifex, so when pulling the translations from transifex they would override the changes you made in this PR.

If there are translation changes for those files, they need to go in transifex for now to avoid problems.

matheusd (approved):

PR #3064, opened 2020-12-08
[tickets] Ensure accounts are properly set for tickets

alexlyp:


app/wallet/control.js
```diff
@@ -179,12 +179,14 @@ export const purchaseTickets = (
...
         request.setMixedSplitAccount(mixedAccount);
+        request.setAccount(accountNum.value);
         request.setChangeAccount(changeAccount);
         request.setCsppServer(csppServer + ":" + csppPort);
         request.setMixedAccountBranch(mixedAcctBranch);
+      } else {
+        request.setAccount(accountNum.value);
+        request.setChangeAccount(accountNum.value);
       }
-      request.setAccount(accountNum.value);
```
vctt94, line 186: Do we need to move setAccount to if/else? I think it is fine moving only `setChangeAccount`

vctt94 (approved):
Tested here and worked. Left only one comment

vctt94 (approved):
tACK

Issue #3067, opened 2020-12-09
default vsp not showing name


vctt94:
If checking `always use this VSP`, leaving the page and going back it will show as select VSP. 

It should show the selected vsp name

![image](https://user-images.githubusercontent.com/15069783/101643713-2be2b700-3a13-11eb-9beb-f2414f8d13d3.png)

bgptr:
on this

PR #3071, opened 2020-12-10
[Tickets] Fix not showing vsp name

bgptr:
Closing #3067

Also noticed that when the VSP listing was disabled but previously clicked on the `Always use this VSP` checkbox, the VSP fee did not appear for the saved VSP host. This diff fixes this too.


jholdstock (approved):
Works for me

alexlyp (approved):

PR #3086, opened 2020-12-15
Bring in Translation for Chinese (Traditional)

smartwojak:
Currently, the locale option of Chinese only means for the Simplified Chinese variant. 

This PR brings in docs and strings of the Traditional Chinese variant, which is the most widely-used and offical language in places such as Hong Kong, Macau and Taiwan.


xaur:
Hey @smartwojak I see more changes since the initial PR. Is this ready to go now or do you plan more changes?

smartwojak:
It is ready to go.

matheusd (approved):

PR #3092, opened 2020-12-15
GetStarted: Hide top nav menu links in nested pages

### LINKED ISSUE #3079: Unclickable menu items on seed copy/entry
`About Decrediton`, `Settings` and `Logs` are displayed during the wallet creation pages - see top right in the screenshots. These were not displayed in 1.5.
`About Decrediton` is clickable, but the other two are not. I suggest these should be hidden, as they were previously.

seed-entered
wallet-seed

amassarwi:
This diff hides the top nav menu links `Settings` & `Logs` when navigating to a nested page in the
wallet creation or wallet restore flows.

Closes #3079


jholdstock:
Works well for me!

alexlyp (approved):

PR #3095, opened 2020-12-16
Add sync vsp ticket by hash button on transaction page

vctt94:
depends on https://github.com/decred/dcrwallet/pull/1951


alexlyp:
I believe we've decided against an individual ticket sync button.  Closing

PR #3105, opened 2020-12-22
Fix race between getStarted and error page

### LINKED ISSUE #3103: [launch] Resource temporarily unavailable not recognized
If starting decrediton with a daemon already running, an error needs to be thrown

vctt94:
closes #3103 .

When starting decrediton, we redirect from "/" to "/getStarted". If an error is thrown before this redirect finishes, the error page shown, but is changed right away. From logs:

image

This PR adds a timeout before redirecting to the error page for trying to solve it. Also I am making the fatalErrorPage a functional component


jholdstock:
Do you have some steps to recreate this issue? I've never seen it before

vctt94:
Just start decrediton in regular mode with dcrd already running.

It is not consistent, so I need to keep trying until I see the error page not showing.

jholdstock (approved):
Thanks, got it. Recreated the error without this PR, and with the PR it seems to be fixed. LGTM

amassarwi (approved):
lgtm

alexlyp (approved):

PR #3106, opened 2020-12-24
Fix address not updating after receveing amount

### LINKED ISSUE #3104: [receive tx] Address reuse
When receiving decred in an address, it shouldn't still be shown on the receive tab, for avoiding address reuse. Which is not the case right now

vctt94:
closes #3104

### DIFF
--- a/app/actions/TransactionActions.js
+++ b/app/actions/TransactionActions.js
@@ -20,6 +20,7 @@ import {
   VSP_FEE_PROCESS_PAID
 } from "constants";
 import { TICKET, VOTE, VOTED, REVOKED } from "constants/Decrediton";
+import { getNextAddressAttempt } from "./ControlActions";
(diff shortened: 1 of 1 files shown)


app/actions/TransactionActions.js
```diff
@@ -46,6 +47,22 @@ function checkAccountsToUpdate(txs, accountsToUpdate) {
   return accountsToUpdate;
 }
 
+// getNewAccountAddresses get accounts which received new inputs and get
+// new addresses for avoiding reuse.
+export const getNewAccountAddresses = (txs) => (dispatch) => {
+  const acctAddressUpdated = [];
+  txs.forEach((tx) => {
```
amassarwi, line 54: no need to `{`

amassarwi (approved):
lgtm

alexlyp (approved):

PR #3112, opened 2021-01-01
multi: Add active tickets qr code.

JoeGruffins:
This adds a button to tickets history that shows a qr with all active
tickets in the current view's txid. It is intended to be used with a
tracking app in order to notify the user of ticket status.


JoeGruffins:
The current modal is not adequate and I believe I will need to make a custom one.

app/index.js
```diff
@@ -432,6 +432,7 @@ const initialState = {
     getNodeInfoAttempt: false,
     towersList: []
   },
+
```
JoeGruffins, line 435: remove

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -112,6 +127,67 @@ const MyTickets = ({ toggleIsLegacy }) => {
...
+      }
+    };
+   updateAddrs()
+   .then(() => {
+     if (qrCommitmentAddrs.length == 0) {
+       qr = null;
+       return;
+     }
+     const qrdata = JSON.stringify(qrCommitmentAddrs);
+     QRCode.toDataURL(qrdata, (err, url) => {
```
JoeGruffins, line 180: don't ignore error

JoeGruffins:
After implementing this for the consumer, it looks like addresses will have to be change to the ticket txid.

JoeGruffins:
screenshot
I'm sure a UI person could make this better, but this looks ok for now to me.

JoeGruffins:
Somehow the ui has gotten worse. Asking @amassarwi to help save me...

amassarwi:
looking rn

package.json
```diff
@@ -290,6 +291,7 @@
     "react-intl-po": "^2.2.2",
     "react-markdown": "^3.3.0",
     "react-motion": "^0.5.2",
+    "react-paginate": "^7.0.0",
```
amassarwi, line 294: can we avoid this dep please ?

app/components/views/TicketsPage/MyTicketsTab/hooks.js
```diff
@@ -3,6 +3,9 @@ import * as sel from "selectors";
 import * as ca from "actions/ClientActions";
 import * as ta from "actions/TransactionActions";
 
+let qrPage = 0;
```
amassarwi, line 6: this should use `useState` instead of global var!

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -15,6 +17,13 @@ const labels = {
   [txTypes.REVOKED]: <T id="ticket.status.multiple.revoked" m="revoked" />
 };
 
+const qrHashes = [];
```
amassarwi, line 20: this logic should move to useTicketsList custom hook, which should expose minimal set of variables to the components to use.

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -112,6 +131,59 @@ const MyTickets = ({ toggleIsLegacy }) => {
     setSortOrderKey(type.value);
   };
 
+  const arraysEqual = (a, b) => {
```
amassarwi, line 134: all of this should move to the custom hook above `useTicketsList` and shouldn't be part of the component.

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -15,6 +17,13 @@ const labels = {
   [txTypes.REVOKED]: <T id="ticket.status.multiple.revoked" m="revoked" />
 };
 
+const qrHashes = [];
+const txPerQR = 15;
+// TODO: Use translated phrase or loading gif.
+const defaultLoading = "Loading...";
```
amassarwi, line 23: this should be a boolean(true/false) - component should decide which message/content to show when `loading=true`.

Also in react you don't use global variables to store data.
Instead you should use mainly `useState` or other react hook.

amassarwi: react hooks intro: https://reactjs.org/docs/hooks-intro.html
useState is in the first example.

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
amassarwi, line 134: we use `lodash` which has many optimized utility functions - https://docs-lodash.com/v4/is-equal/ can be used to compare arrays(and bunch of other stuff).

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -4,6 +4,8 @@ import { FormattedMessage as T } from "react-intl";
 import { useTicketsList } from "./hooks";
 import * as txTypes from "constants/Decrediton";
 
+const QRCode = require("qrcode");
```
amassarwi, line 7: This is the syntax to import a module in a nodejs app, we want to use ES6/ES7 syntax:

Note the import statement and the usage of promises instead of callback.
```
import QRCode from 'qrcode'
 
// With promises
QRCode.toDataURL('I am a pony!')
  .then(url => {
    console.log(url)
  })
  .catch(err => {
    console.error(err)
  })
 
// With async/await
const generateQR = async text => {
  try {
    console.log(await QRCode.toDataURL(text))
  } catch (err) {
    console.error(err)
  }
}
```

ref: https://www.npmjs.com/package/qrcode#es6es7

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -112,6 +131,59 @@ const MyTickets = ({ toggleIsLegacy }) => {
...
+      return;
+    }
+    qrHashes.length = 0;
+    qrs.length = 0;
+    setQRPage(0);
+    if (newQRHashes.length == 0) {
+      loadingQR = null;
+      return;
+    }
+    qrs.length = Math.ceil(newQRHashes.length / txPerQR);
```
amassarwi, line 169: don't need to set length on an array in JS.

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -112,6 +131,59 @@ const MyTickets = ({ toggleIsLegacy }) => {
...
+    qrHashes.push.apply(qrHashes, newQRHashes);
+    for (let i = 0; i < qrs.length; i++) {
+      const start = i * txPerQR;
+      const qrdata = JSON.stringify(newQRHashes.slice(start, start + txPerQR));
+      QRCode.toDataURL(qrdata, (err, url) => {
```
amassarwi, line 174: See comment above about using promise here instead of `(err, url) => {}` callback function.

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.jsx
```diff
@@ -112,6 +131,59 @@ const MyTickets = ({ toggleIsLegacy }) => {
...
+
+  // getQR asychronously updates the qr codes if current viewed ticket hashes
+  // have changed since last viewing.
+  const getQR = () => {
+    loadingQR = defaultLoading;
+    const filteredTiks = tickets.filter(activeFilter);
+    const newQRHashes = [];
+    for (let i = 0, len = filteredTiks.length; i < len; i++) {
+      newQRHashes.push(filteredTiks[i].txHash);
+    }
```
amassarwi, line 156: can use Array.map

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.module.css
```diff
@@ -54,6 +61,28 @@
   margin-left: 4px;
 }
 
+.menuButton {
```
amassarwi, line 64: this class might be not needed now

app/components/views/TicketsPage/MyTicketsTab/MyTicketsTab.module.css
```diff
@@ -54,6 +61,28 @@
...
+}
+
+.qrButton {
+  background-position: 10px 1px;
+  background-repeat: no-repeat;
+  background-size: 18px 18px;
+  width: 40px;
+  height: 20px;
+  border-radius: 4px;
+  padding: 0px;
```
amassarwi, line 76:
```suggestion
  padding: 0;
```

app/components/views/TicketsPage/MyTicketsTab/hooks.js
```diff
@@ -19,6 +23,58 @@ export const useTicketsList = () => {
   const changeTicketsFilter = (newFilter) =>
     dispatch(ta.changeTicketsFilter(newFilter));
 
+  const [QRs, setQRs] = useState([]);
+  const [qrHashes, setQRHashes] = useState([]);
+  const [QRsPage, setQRsPage] = useState(0);
+  const [loadingQRs, setLoadingQRs] = useState(true);
+
+  // XXX make app or file level constant.
+  const txPerQR = 2; // 15
```
amassarwi, line 32: i would make a file constant here `const TX_PER_QR = 15`

app/components/views/TicketsPage/MyTicketsTab/hooks.js
```diff
@@ -19,6 +23,58 @@ export const useTicketsList = () => {
...
+      );
+      qrsPromises.push(QRCode.toDataURL(qrdata));
+    }
+    Promise.all(qrsPromises)
+      .then((qrs) => {
+        setQRHashes(filteredHashes);
+        setQRs(qrs);
+        setLoadingQRs(false);
+      })
+      // XXX snackbar error instead plzx.
```
amassarwi, line 74: snacknack reminder

amassarwi:
@JoeGruffins needs a `yarn prettify`.

package.json
amassarwi: it's ok if we decide to merge it with this dep, bit definitely worth creating a separate issue for creating our own, 
it could be a `pi-ui` component.

JoeGruffins:
So, not sure if because of my rebase, but icon is a little off here, along with tooltip. Also, the pending ticket is not in active qrs, although it should be. Also, also, that ticket isn't pending, it has 25 confirms as of this writing.

messedup

JoeGruffins:
The icons are ok now, but the tooltips are messed up for the others as well. One just says "My". I guess the spacing is bad. I also think the unmined ticket problem is unrelated to this pr.

matheusd:
Seems funcionally ok, but the styling of the modal seems very off. Having the paginator on the bottom of the page, over a blurred, transparent background makes it almost invisible.

JoeGruffins:
I probably messed something up with a rebase. Thank you for reviewing, will fix.

JoeGruffins:
Yeah, a few things are out of wack, please hold.

JoeGruffins:
@matheusd Thank you for your patience!

I had indeed left out some commits somewhere. Everything should look as intended now. We settled on 10 txid per qr code, because 15 was just a little too huge.

matheusd (commented):
Design is still not great for me (notice scrollbars and paginator on top of the qr code itself)

image

JoeGruffins:
Thank you again @matheusd 

How about now?

matheusd (commented):
The UI still isn't 100% (there are some issues when the height of screen is too small and responsive issues) but I guess we can iterate on those.

Seems there's uncommitted changes to `yarn.lock` though (try doing a `yarn` and see if you get new changes to yarn.lock).

matheusd (approved):
Working fine, though as I mentioned the design can use some sprucing up

PR #3113, opened 2021-01-01
Tx export error handling

### LINKED ISSUE #3107: export txs can't recover from error
If an error happens when exporting transactions, for example, trying to export for a directory with no write access. No error shows and it keeps forever loading, which can not be recovered. Example:

image

bgptr:
Closes #3107


vctt94 (approved):
Thanks. Tested and works fine.

PR #3120, opened 2021-01-05
move wallet config initial values to constants

vctt94:
This PR moves how we start our wallet initial config, based on how we start our global config, for keeping consistency and simplicity.

It moves the constants initial value to our constant file and initialize them.


alexlyp (approved):
ACK

PR #3122, opened 2021-01-06
Redraw optimization

bgptr:
After some profiling, I found these issues:
- leaving the app for some time - or refresh it by F5 or close wallet and reopen it - the `GETPEERINFO_ATTEMPT/GETPEERINFO_SUCCESS` requests pile up (**update**: #2901 resolves this part of the issue)
- the `Privacy` and the `SendTransaction` components are redrawn after every `GETPEERINFO_ATTEMPT/GETPEERINFO_SUCCESS` request despite the redux state did not change at all.
- these two factors can increase CPU usage quite high after some time.

This PR fixes this.

### DIFF
--- a/app/containers/Wallet.js
+++ b/app/containers/Wallet.js
@@ -29,8 +29,9 @@ const pageAnimation = {
 
 @autobind
 class Wallet extends React.Component {
-  componentDidMount() {
-    const { compareInventory, politeiaEnabled, getPeerInfo } = this.props;
+  constructor(props) {
+    super(props);
+    const { compareInventory, politeiaEnabled, getPeerInfo } = props;
     // Compare politeias inventory and update proposal list if they are different
     // every 1 minute.
     this.fetchPoliteiaInventory = this.props.setInterval(() => {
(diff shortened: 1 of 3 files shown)


app/containers/Wallet.js
```diff
@@ -41,12 +40,18 @@ class Wallet extends React.Component {
...
     // peers.
-    this.props.setInterval(() => {
-      if (politeiaEnabled) {
-        getPeerInfo();
-      }
+    this.peerInfoIntervalRef = this.props.setInterval(() => {
+      getPeerInfo();
     }, 10000);
   }
+  componentWillUnmount() {
```
amassarwi, line 47: I already did this in https://github.com/decred/decrediton/pull/2901 after refactoring this container to be
functional component, I would recommend merging #2901 then merge this one for the other part which isn't covered in #2901

amassarwi (commented):
Thanks for this @bgptr 

I left a general comment regarding the intervals cleanup in the wallet container which was fixed already in #2901 which also refactors the wallet container to be functional, I would suggest merging #2901 and then this one for the other part.

app/containers/Wallet.js
amassarwi: What do you think guys @vctt94 @alexlyp ?
Would it make sense rebasing 2901 and merge it at this level ? 
Looks like the intervals cleanup is essential performance improvement.

bgptr:
@amassarwi Great, I've reverted the `Wallet.js` so it won't conflict with your PR.

app/containers/Wallet.js
vctt94: I think that's fine, can you rebase it, so I can review it?

amassarwi: Surething, on it now

vctt94 (approved):
Thanks, it does improve the perfomance on privacy page!

alexlyp (approved):

PR #3123, opened 2021-01-06
Improve UI performance

guilhermemntt:
As discussed in Decrediton room in Matrix, according to @alexlyp:

> (...) wondering if you guys could take a look at a performance issue i've noticed recently. was running tip decrediton and noticed a bit of CPU churn after starting the mixer (on the privacy page). I'm thinking it could be the gifs on that page, though i still see the CPU bump when toggling to a different page (transactions or tickets, etc)

and @vctt94:

> I noticed my CPU struggling more on latest master, as well

It seems that our recent changes mainly in the privacy page resonated in some lack of performance. After investigating, we realized high CPU usage due to ``react-router-transition`` expensive transitions and ``menuMixer.svg`` CSS animation.

Below, the performance of three use cases, all of them following the script: 10s on privacy page, 10s on accounts page, and 10s on privacy page again, disabling ``react-router-transition`` and while the mixer is started. Data was acquired using Dev Tools.

image

Now, using ``react-router-transition``.

image

- 1 (no gif nor svg);
- 2 (only gif in provacy page);
- 3 (mix svg and gif in privacy page).

In order to diminish CPU usage, the following resolutions have been taken in this PR:

- Remove ``react-router-transition``;
- Replace ``menuMixer.svg`` with a new GIF ``menuMixer.gif`` that renders in 5 frames per second, rather than 60 FPS;

menuMixer

- Optimize ``mixerArrows.gif`` and ``decentralizedLoopAnimation.gif`` to use 1/4 of their frames.

After that, CPU usage was measured again, comparing now mixer started (# 1) and stopped (# 2).

Captura de tela de 2021-01-06 19-59-49

The GIF's replacement and optimization can be discussed with @linnutee as well, as some quality has been lost in exchange for performance. Also, we still can achieve more performance using static images.

### DIFF
--- a/app/containers/App.js
+++ b/app/containers/App.js
@@ -3,7 +3,6 @@ import { defaultFormats } from "i18n/locales";
 // XXX useTheming already exists in app/hooks
 import { app, theming } from "connectors";
 import { Redirect, Route, Switch } from "react-router-dom";
-import { AnimatedSwitch } from "react-router-transition";
 import { StaticSwitch } from "shared";
 import GetStartedContainer from "./GetStarted";
 import WalletContainer from "./Wallet";
@@ -17,12 +16,6 @@ import "style/Layout.less";
 import { ipcRenderer } from "electron";
 import { hot } from "react-hot-loader/root";
 import { CantCloseModals } from "modals";
-
-const topLevelAnimation = {
-  atEnter: { opacity: 0 },
-  atLeave: { opacity: 0 },
-  atActive: { opacity: 1 }
-};
 
 // minimum size to reduce the sidebar in px.
 const MINIMUM_SIZE_TO_REDUCE_SIDEBAR = 1179;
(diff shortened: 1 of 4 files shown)


jholdstock:
- `menuMixer.gif` is too small, and IMO it is also a bit too slow.
- New arrows look great - no appreciable difference to before (viewed on a 21.5 inch 1080 display).
- The new `decentralizedLoopAnimation.gif` looks terrible IMO. It makes decrediton look buggy and struggling for performance, and its not appropriate for a brand new headline feature. Can we find some middle ground between the original gif and the new one?

vctt94 (commented):
thanks for this PR and your time looking into this.

I think the performance has have some improvements, but I agree with jhold, that the spinning icon looks buggy. 

I think we can reach a middle ground here, which looks good enough/ performance not so much compromised.

guilhermemntt:
Great, thank you for your review. I'll try a more conservative approach here.

guilhermemntt:
Due to https://github.com/decred/decrediton/pull/3231 and https://github.com/decred/decrediton/issues/3243, I'm restricting this PR to the removal of ``react-router-transition``.

alexlyp (approved):

Issue #3124, opened 2021-01-07
ui_animations setting does not affect tab transitions


guilhermemntt:
Currently, ``ui_animations: false`` does not affect ``react-motion`` transitions in tabs. Those transitions should occur only when ``ui_animations: true``.

guilhermemntt:
On it.

PR #3129, opened 2021-01-07
Disable tab animations

guilhermemntt:
This PR disables tab animations when ``ui_animations: false`` in the config file. It disables sidebar transitions as well.

Closes https://github.com/decred/decrediton/issues/3124.


alexlyp:
So I definitely like this PR, but can you add a UI Animations knob to the settings page so users can toggle this without going into the config.json?

alexlyp:
Looks like General > UI would be a good spot for it.

And I think it would also be valuable to add a ? to click (like what is next to Gap Limit) to explain what changing this will do.

alexlyp:
You can use the following to add to i18n/docs/InfoModals/uiAnimations.md

# UI Animations

When set to true, UI Animations will provide a fade-in/out when going from page to page.  And on tabbed pages, this will  cause tabs to slide into place.

It's recommended to set to false on machines where performance may be an issue.

guilhermemntt:
Thank you @alexlyp for the nice suggestion. I've just added a checkbox, but I'm not sure if it looks good. Anyway, some changes have also been made in ``app/index.js`` regarding the initial state.

image

alexlyp:
I think a dropdown to match makes sense.  True/False or On/Off is good

guilhermemntt:
Yes, I think it's much better now. I've used the options "Disabled" and "Enabled" to keep our pattern.

alexlyp (approved):
tACK

PR #3133, opened 2021-01-09
[governance] Remove politeia CSRF tokens

### LINKED ISSUE #3117: Remove politeia CSRF tokens
decred/politeia#1356 removed CSRF protection on public routes due to privacy issues with attaching a CSRF token to a request that casts votes.  This has been deployed to production as a patch.

The existing behavior of including CSRF tokens will not cause any errors, the tokens are now just ignored.  Decrediton can remove all code for handling politeia CSRF tokens since Decrediton only uses public routes.  It should be verified that neither cookie nor header CSRF tokens are included in requests that cast votes.

amassarwi:
This commit cleans logic related to storing and attaching CSRF tokens
when communicating with Politeia.

Closes #3117


jholdstock (approved):
lgtm and works well

alexlyp:
Is the CSRF removal live on mainnet production?  This was tested on both testnet and mainnet?

amassarwi:
Yep I tested both mainnet and testnet but not voting on proposals, 
let me try that

alexlyp (approved):

Issue #3131, opened 2021-01-08
Deprecated type checks


guilhermemntt:
We should use ``typeof`` operator rather than the deprecated type checks of ``Node.js`` ``util`` module.

![image](https://user-images.githubusercontent.com/39631429/104051216-90399780-51c6-11eb-8e35-af056e372c9d.png)

guilhermemntt:
On this one.

PR #3135, opened 2021-01-09
Remove util type checks

guilhermemntt:
Closes https://github.com/decred/decrediton/issues/3131.

### DIFF
--- a/app/actions/StatisticsActions.js
+++ b/app/actions/StatisticsActions.js
@@ -1,7 +1,7 @@
 import * as wallet from "wallet";
 import * as sel from "selectors";
 import fs from "fs";
-import { isNumber, isNullOrUndefined, isUndefined } from "util";
+import { isNumber, isNil, isUndefined } from "lodash";
 import { endOfDay, formatLocalISODate, isSameDate } from "helpers";
 import {
   REGULAR,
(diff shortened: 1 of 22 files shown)


app/actions/StatisticsActions.js
```diff
@@ -391,7 +390,7 @@ export const exportStatToCSV = (opts) => (dispatch, getState) => {
   // called once for each data line
   const progressFunction = (time, series) => {
     const values = allSeries.map((s) =>
-      !isUndefined(series[s.name]) ? s.valueFormatFunc(series[s.name]) : null
+      series[s.name] !== undefined ? s.valueFormatFunc(series[s.name]) : null
```
amassarwi, line 393: This isn't enough, you could also check if typeof x === "undefined" but I would rather user lodash `isUndefined` func for that.

Lodash may have it's own version of `isNumber, isNullOrUndefined` funcs as well, checkout their documentation.

amassarwi (changes requested):
Thanks for this PR, left an inline suggestion.

guilhermemntt:
Thanks @amassarwi, you're totally right. Just to be clearer: ``something === "undefined"`` throws error if ``something`` hasn't been defined yet. So, there was a trap in ``@deprecated`` annotation and its suggestion. Following your suggestion now, ``lodash`` has all these type check, including [isNil](https://lodash.com/docs/4.17.15) which is equivalent to ``isNullOrUndefined``.

amassarwi (approved):
Noice 👍

vctt94:
needs rebase

guilhermemntt:
Rebased.

alexlyp:
Quite a few conflicts that will need to be dealt with.

matheusd (approved):
Nice job. Thanks for this!

PR #3136, opened 2021-01-09
Enable switching accounts in coinjoin account selection page

### LINKED ISSUE #3130: coinjoin account selection can't change options
When selecting accounts on coinjoin selection page, if choosing one account, it makes not possible to change:

image

It should be allowed to change

amassarwi:
Closes #3130.


alexlyp (approved):
tACK

PR #3142, opened 2021-01-13
ipc: Check Trezor model when uploading firmware.

JoeGruffins:


matheusd (approved):
Tested on a model one on the dev and packaged versions on linux.

PR #3144, opened 2021-01-14
WalletLoaderActions: Clear trezor values on close.

JoeGruffins:
This ensures that session.connect is called when changing wallets, which
clears the password data stored on the Trezor, allowing switching trezor
wallets with password enabled.

### DIFF
--- a/app/actions/TrezorActions.js
+++ b/app/actions/TrezorActions.js
@@ -37,6 +37,7 @@ const BOOTLOADER_MODE = "bootloader";
 
 let setListeners = false;
 
+export const TRZ_WALLET_CLOSED = "TRZ_WALLET_CLOSED";
 export const TRZ_TREZOR_ENABLED = "TRZ_TREZOR_ENABLED";
 
 // enableTrezor attepts to start a connection with connect if none exist and
--- a/app/actions/TrezorActions.js
+++ b/app/actions/TrezorActions.js
@@ -65,6 +65,7 @@ export const initTransport = async (session, debug) => {
     env: "web",
     lazyLoad: false,
     popup: false,
+    transportReconnect: false,
     manifest: {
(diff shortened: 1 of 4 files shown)


app/actions/TrezorActions.js
```diff
@@ -65,6 +65,7 @@ export const initTransport = async (session, debug) => {
     env: "web",
     lazyLoad: false,
     popup: false,
+    transportReconnect: false,
```
JoeGruffins, line 68: Default should be false, but just making sure.

app/helpers/trezor.js
```diff
@@ -74,15 +74,14 @@ export const walletTxToBtcjsTx = async (
   });
 
   const outputs = tx.outputs.map(async (outp, i) => {
-    if (!outp.decodedScript.address) {
+    const addr = outp.decodedScript.address;
+    if (!addr) {
       // TODO: this will be true on OP_RETURNs. Support those.
```
JoeGruffins, line 79: refactoring

app/reducers/trezor.js
```diff
@@ -99,7 +99,8 @@ export default function trezor(state = {}, action) {
       return {
         ...state,
         device: null,
-        deviceLabel: null
+        deviceLabel: null,
+        connected: false
```
JoeGruffins, line 103: This was needed. Otherwise pulling the device will put us in a state where we are connected but have no device.

vctt94: might be good dropping a comment into the code?

vctt94: might be good dropping a comment into the code?

JoeGruffins: Ok.

matheusd (approved):

PR #3162, opened 2021-01-19
Call processManageTickets on every startup if having live tickets

vctt94:
This PR changes how we call processManaged tickets, to be called before processUnmanagedTickets and also to be called on every startup if the wallet have live tickets


alexlyp (approved):
tACK

PR #3165, opened 2021-01-20
[CSS] fix purchase tickets page misalignments

### LINKED ISSUE #3164: Minor CSS misalignments in "Purchase Tickets" tab
Some of them are very recurrent by the way, probably due to some overlapping rebases.

image

bgptr:
Closes #3164

### DIFF
--- a/app/components/buttons/ToggleSwitch/ToggleSwitch.module.css
+++ b/app/components/buttons/ToggleSwitch/ToggleSwitch.module.css
@@ -1,7 +1,7 @@
 .toggleSwitch {
   display: inline-block;
   vertical-align: middle;
-  margin: 0 0 0.5em 10px;
+  margin: 0 0 10px 1em;
 }
 
 .enabled,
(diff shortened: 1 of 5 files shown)


app/components/buttons/ToggleSwitch/ToggleSwitch.module.css
```diff
@@ -1,10 +1,7 @@
 .toggleSwitch {
   display: inline-block;
   vertical-align: middle;
-  margin-right: 0.5em;
-
-  margin-bottom: 10px;
-  width: 85px;
+  margin: 0 0 0.5em 10px;
```
vctt94, line 4: Any reasons for using em and px?

vctt94 (approved):
LGTM, left just a comment with a question for you

app/components/buttons/ToggleSwitch/ToggleSwitch.module.css
bgptr: Actually, it was a mistake. My original intention was to add more margin to the left and remove it from the right while keeping the bottom. I've fixed it now, thanks for the test.

PR #3173, opened 2021-01-22
Add initial value to new vsp tickets filter

vctt94:
This PR adds a initial value to the ticket status for new vsp tickets


alexlyp (approved):
tACK

PR #3175, opened 2021-01-23
Fix declaration for pi-ui and regen yarn.lock.

jrick:
This format matches the dependency declaration for another
Decred dependency, node-addon-loader.

Regenerating yarn.lock using this fix modifies the "resolved" line
for fetching the dependency's source, in a format that is understood
by the Nix package manager after converting the yarn.lock using
yarn2nix.  Previously, the HTML for the GitHub project was fetched
instead, which failed checksum matches and never would have been usable
in the build.

To avoid changing the version of pi-ui in the build, the package.json
references the Git sha1 of the previously resolved commit.  This can
be changed to a branch or version specification when it needs to be
updated later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -287,7 +287,7 @@
     "node-abi": "^2.15.0",
     "node-addon-loader": "decred/node-addon-loader#master",
     "nouislider": "^12.0.0",
-    "pi-ui": "https://github.com/decred/pi-ui",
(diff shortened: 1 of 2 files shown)


jrick:
I don't understand the CI test failures, so hopefully someone does :)

alexlyp:
Closing this since i think it's old and can be retried on master

PR #3176, opened 2021-01-23
Functional inputs

bgptr:
In this diff, I went through the following input components and refactored them to functional components using hooks and CSS modules, if not already done:
- Input: refactored to a functional component, `.less` file refactored to css module, 
- FeeInput: use hooks instead of a connector
- InlineField: `.less` file refactored to css module 
- NumericInput: `.less` file refactored to css module 
- NumTicketInput:`.less` file refactored to css module
- PassphraseModalField:  `.less` file refactored to css module
- PathInput:  `.less` file refactored to css module
- PathBrowseInput:  refactored to a functional component, `.less` file refactored to css module
- SettingsInput: refactored to a functional component
- AddressInput, BlocksInput, FloatInput, IntegerInput, SettingsTextInput, TextInput components have just renamed from '.js' to '.jsx'.

Also, it includes automated tests.

Part of #2438 and #2439


app/components/inputs/SettingsInput.js
```diff
@@ -1,6 +1,5 @@
 import Select from "react-select";
 import { injectIntl } from "react-intl";
-import "style/Input.less";
 
 @autobind
```
amassarwi, line 4: Do you think we could refactor this to be a functional component as part of this effort ?

app/components/buttons/PathButton.js
```diff
@@ -3,7 +3,11 @@ import { Tooltip } from "shared";
 
 const PathButton = ({ disabled, onClick }) => (
   <Tooltip text={<T id="startup.dataDir.tip" m={"Select a path"} />}>
-    <div className={"path-button"} {...{ disabled, onClick }} />
+    <button
+      aria-label="Select a path"
+      className={"path-button"}
```
amassarwi, line 8: where is this class coming from ? ideally we should introduce local css module here as well.

amassarwi (approved):
Nice work, code looks good!

Left couple of suggestions.

bgptr:
@amassarwi Thank you for your review. I've implemented your suggestions.

alexlyp (approved):

Issue #3145, opened 2021-01-14
check if address already updated


vctt94:
When receiving a tx, decrediton generates a new address, but it will generate a new address, when receiving the unmined tx and when mining it. 

We should check if the adress is already updated, so we only generate one new address.

guilhermemntt:
On it.

PR #3178, opened 2021-01-25
Check already generated address when receiving mined tx

guilhermemntt:
Depends on https://github.com/decred/dcrwallet/pull/1967 and closes https://github.com/decred/decrediton/issues/3145.

This PR ensures that the current address stored in decrediton - ``nextAddress`` - is renewed if it was already used in some previous transaction and, in this way, strengthens the user's privacy. This checking is performed through ``getReceivedByAddress`` gRPC call to dcrwallet every time a new transaction is received.


alexlyp:
Closing due to age.  And not sure relying on a new dcrwallet request is the best way to solve this issue.

Issue #3177, opened 2021-01-23
1.6 RC5 - Staking fatal error


mwade290:
## Background
- Mac OS Catalina 10.15.7
- Decrediton 1.6 RC5
- Primary Account contains funds
- Mixed Account has 0 funds
- Staking page Purchase Tickets does not give option to use Primary Account (only Mixed) - is this a bug?

## Steps
- Navigate to Accounts page
- Click Mixed Account
- Click the eye icon to hide
- Attempt to navigate to Staking page
- Application just has blue background and will not close, must be force closed by Mac OS and application restarted

## To fix
Mixed Account must not be hidden

guilhermemntt:
> Staking page Purchase Tickets does not give option to use Primary Account (only Mixed) - is this a bug?

No, as answered in Decred room in Matrix:

> It's expected to not be able to select any other account than mixed after privacy is enabled.

Apart from that, I'll take a look at this one.

PR #3179, opened 2021-01-25
Add account validation and warning tooltip in AccountsSelect

guilhermemntt:
Closes https://github.com/decred/decrediton/issues/3177.

Currently, it's possible to hide _mixed_ and _change (unmixed)_ accounts even when "Allow send from unmixed accounts" isn't checked (balance should be zero DCR, of course). This PR prevents users from hiding these accounts in this case. Also, if one checks "Allow send from unmixed accounts" and hide one of these accounts, then they will be unhidden automatically just after  "Allow send from unmixed accounts" has been unchecked.

This PR also checks for the ``account`` object definition before trying to access some attribute of it. Furthermore, a tooltip was added in ``AccountsSelect`` component to warn users when there's no available visible account to be selected. This is especially important when trying to buy tickets with privacy enabled, since that, in this case, it's impossible to buy a ticket using an account other than _mixed_, even if "Allow send from unmixed accounts" is checked.

image

### DIFF
--- a/app/actions/AccountMixerActions.js
+++ b/app/actions/AccountMixerActions.js
@@ -4,8 +4,8 @@ import * as sel from "selectors";
 import * as wallet from "wallet";
 import { getWalletCfg } from "config";
 import {
+  getAccountsAttempt,updateAccount,
   getAcctSpendableBalance,
-  getAccountsAttempt,
   getMixerAcctsSpendableBalances
 } from "./ClientActions";
 import {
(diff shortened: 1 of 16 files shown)


guilhermemntt:
Hide unmixed account leads to some other problems as well, I'm checking this out.

alexlyp:
Needs a rebase.

guilhermemntt:
Rebased.

alexlyp:
Closing PR.  Feel free to reopen or resubmit after rebasing against tip.

PR #3182, opened 2021-01-26
Disable autobuy when changing VSP modes

### LINKED ISSUE #3181: Disable autobuy when changing VSP modes
When a user has auto ticket enabled and switches from legacy <-> "new" mode. decrediton still has auto purchase enabled and running ?. (same other way around) 

This might lead to surprise purchases that become hard to track.

bgptr:
Closes  #3181
Also, I've added snackbar notifications for the new auto buyer's start and stop events similar to the legacy auto buyer's events. Thus it is a little bit easier to follow what is happening in the background.


degeri:
tested. LGTM

amassarwi (approved):
LGTM

alexlyp (approved):

PR #3192, opened 2021-01-29
Bump to Electron 9.4.0

bgptr:
I set the `allowRendererProcessReuse` flag to `false` that reverts to the old behavior. It is functional for now, but this could be just a temporary solution until Electron 11.


alexlyp (approved):
tACK

Issue #3188, opened 2021-01-27
"Insufficient funds" message flickering in "Send DCR" tab


guilhermemntt:
When trying to send DCR  with insufficient funds, "Insufficient funds" message keeps flickering and a new transaction is built indefinitely.

![Peek 2021-01-27 16-06](https://user-images.githubusercontent.com/39631429/106040930-e741d680-60b9-11eb-9c23-f8f4098aefbf.gif)

guilhermemntt:
On this one.

PR #3213, opened 2021-02-11
Fix flickering insufficient funds message in "Send" tab

guilhermemntt:
Closes https://github.com/decred/decrediton/issues/3188 by removing ``getNextAddressAttempt`` redundant call. This issue can be reproduced by trying to send from _mixed_ to _unmixed_ accounts when having insufficient funds to do that. As ``nextAddress`` was updated, ``useEffect`` hook in ``app/components/shared/SendTransaction/SendTransaction.jsx`` tended to run indefinitely and new outputs were generated based on that new address, so the transaction was constructed generating a new address again. There's no need of calling ``getNextAddressAttempt``, for it's already called every time a new destination account is selected. It means that ``nextAddress`` is always synced with the selected destination account (``nextAddressAccount``).

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -536,7 +536,7 @@ export const constructTransactionAttempt = (
   confirmations,
   outputs,
   all
-) => async (dispatch, getState) => {
+) => (dispatch, getState) => {
   const request = new ConstructTransactionRequest();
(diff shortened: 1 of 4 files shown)


guilhermemntt:
Rebased.

alexlyp (approved):

PR #3215, opened 2021-02-12
Add TransactionPage CSS module

guilhermemntt:
Implementation of https://github.com/decred/decrediton/issues/2439 in ``app/components/views/TransactionPage/`` components.


amassarwi (commented):
Code LGTM nice work, one thing I noticed we use the same css module to all three
 quite big components and I know we might do the same in other places but ideally we should split that css module to 3 modules one for each component, placed inside nested directory next the component's `.jsx` file.

guilhermemntt:
Thanks for the review @amassarwi. As you mentioned, that CSS module could be separated into two others placed with their respective components. Note that sometimes it's needed to access some CSS classes from the parent component through very specific selectors (``>``, ``:nth-child(n)``, etc.). Another possible option is to propagate some prop and apply a style conditionally (``myProp && styles.myStyle``), but this prop has to exist and store the proper condition value. We have some global CSS rules also, _but they are global_ and probably their application would be inappropriate. I think it's important to put this topic into discussion as we tend to deal with more complex components afterward.  Actually, this problem can be particularly perceived among the modals, since they're implemented as a HOC and its lower-order components. So, I haven't tried this yet, but 
[this](https://github.com/webpack-contrib/css-loader#scope) link and [this](https://github.com/css-modules/css-modules#exceptions) other can provide us an alternative (and more sophisticated) way to define CSS properties of lower components based on higher ones. Just wanted to share. I'll make some tests and then, if they succeed, we can implement it after the CSS modules refactoring ends.

amassarwi:
@guilhermemntt I think a common approach that we already use in `pi-ui` & `pigui` codebases is to exposes *className props, children can expose several className props to provide the parents/consumers applying custom styling using these props.

Combining the above approach with css modules will result in custom styling in the parents' css modules instead of multiple nested classes inside the children using parents' class wrappers to override default styling **in the children* `.less` file which is completely out of context.

Ideally this PR should go in before the `components/layout/*` refactoring i'm working on as it conflicts with `layout.less` file because of thaw i tried to explain above.

guilhermemntt:
Rebased.

amassarwi (approved):
ACK

alexlyp (approved):

PR #3230, opened 2021-02-16
[tickets] Add sanity check for legacy purchase tickets

alexlyp:
We've had various reports of legacy ticket purchases not having Pool Address and Pool Fees populated which resulted in malformed legacy VSP tickets.  This is a final sanity check that should catch any that are attempted.  While this won't fix what's causing the core problem, it will stop them from happening until it's identified.

### DIFF
--- a/app/wallet/control.js
+++ b/app/wallet/control.js
@@ -127,7 +127,8 @@ export const purchaseTickets = (
   signTx
 ) =>
   new Promise((ok, fail) => {
-    if (stakepool.PoolAddress == "" || stakepool.PoolFees == 0) {
+    if ((!stakepool.PoolAddress && stakepool.PoolAddress == "") ||
+    (!stakepool.PoolFees && stakepool.PoolFees == 0)) {
       return fail("Purchase ticket failed: Pool Address or Pool Fees can't be empty");
     }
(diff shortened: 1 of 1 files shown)


app/wallet/control.js
```diff
@@ -127,6 +127,10 @@ export const purchaseTickets = (
   signTx
 ) =>
   new Promise((ok, fail) => {
+    if (stakepool.PoolAddress == "" || stakepool.PoolFees == 0) {
+      fail("Purchase ticket failed: Pool Address or Pool Fees can't be empty");
```
amassarwi, line 131: you should be able to do `return fail(...` i think

amassarwi (approved):

app/wallet/control.js
```diff
@@ -127,6 +127,9 @@ export const purchaseTickets = (
   signTx
 ) =>
   new Promise((ok, fail) => {
+    if (stakepool.PoolAddress == "" || stakepool.PoolFees == 0) {
```
vctt94, line 130: I think would be better to have a check like this
```suggestion
    if (!stakepool.PoolAddress || !stakepool.PoolFees) {
```

To be sure it also caught undefined

vctt94 (approved):
LGTM

PR #3238, opened 2021-02-18
Improve error handling on wallet create and restore views

### LINKED ISSUE #3235: [Wallet restore] `Create Wallet` button is enabled when it shouldn't
On the wallet restore view if the inputs are filled properly, the `Create Wallet` button became enabled. 
It is good, but after this, the button should be disabled if one of the inputs is changed to an invalid state. 
The user must not continue if e.g. the passphrases don't match.

DeepinScreenshot_select-area_20210217162111

bgptr:
closes #3235

Now the `Create Wallet` button reacts to invalid inputs properly. 
After the button was enabled, it is toggled to disabled state when:
- passphrases cleared or,
- passphrases do not match or,
- seeds are not valid or,
- one or more seeds are deleted

Automated tests were expanded to cover this new behavior.

### DIFF
--- a/app/components/views/GetStartedPage/CreateWalletPage/CreateWalletPage.jsx
+++ b/app/components/views/GetStartedPage/CreateWalletPage/CreateWalletPage.jsx
@@ -70,12 +70,13 @@ const CreateWalletPage = ({ createWalletRef, onSendBack }) => {
   );
 
   const checkIsValid = useCallback(() => {
-    const { seed, passPhrase } = current.context;
+    const { seed, passPhrase, error } = current.context;
     // We validate our seed and passphrase at their specific components
     // So if they are set at the machine it means they have passed validation.
     if (!seed || !passPhrase) return setIsValid(false);
     if (seed.length === 0) return setIsValid(false);
     if (passPhrase.length === 0) return setIsValid(false);
+    if (error !== "") return setIsValid(false);
 
     return setIsValid(true);
   }, [setIsValid, current.context]);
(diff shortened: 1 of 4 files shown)


app/components/views/GetStartedPage/CreateWalletPage/CreateWalletPage.jsx
```diff
@@ -70,12 +70,13 @@ const CreateWalletPage = ({ createWalletRef, onSendBack }) => {
...
 
   const checkIsValid = useCallback(() => {
-    const { seed, passPhrase } = current.context;
+    const { seed, passPhrase, error } = current.context;
     // We validate our seed and passphrase at their specific components
     // So if they are set at the machine it means they have passed validation.
     if (!seed || !passPhrase) return setIsValid(false);
     if (seed.length === 0) return setIsValid(false);
     if (passPhrase.length === 0) return setIsValid(false);
+    if (error !== "") return setIsValid(false);
```
vctt94, line 79: I guess we can have 
```suggestion
    if (error) return setIsValid(false);
```

to avoid cases like error equals null, for example

vctt94 (commented):
Left a coment and thanks for the tests!

app/components/views/GetStartedPage/CreateWalletPage/CreateWalletPage.jsx
bgptr: you are right, thx!

alexlyp (approved):

PR #3245, opened 2021-02-19
TabsHeader: Fix wrong aligned tabs in finished propsoals list.

amassarwi:
This improves the alignment of TabHeader component which used is in the finished proposals list.

It Replaces `TabsHeader`'s `tabsWrapperClassName` prop with a new `className` prop which is used to override
the default absolute position css attributes.
With that I was able to position the tabs in `HomePage` and  `ProposalsFilter` components differently.


amassarwi:
@bgptr Are you aware of any third place where we use the TabsHeader component besides ProposalFilter & HomePage ? 
We will need to make sure that the tabs are positioned properly everywhere.

alexlyp (approved):
tACK

PR #3247, opened 2021-02-19
[stakinng] Add forgotten selector

alexlyp:
Forgot to add a selector in the previous Syncing Failed tickets PR which was causing Ticket Status tab to crash on click


vctt94 (approved):
tACK

Issue #3239, opened 2021-02-18
Use `husky` to run pre-commit command(s).


amassarwi:
We have added `Pretiffy` as formatting tool and the `yarn prettify` works well, but we as it for now
we can just use it manually but with `husky` it's possible to run `yarn pretiffy` and `yarn lint` automatically as pre-commit easily.

See example from pigui repo:
```
"husky": {
    "hooks": {
      "pre-commit": "lint-staged"
    }
  },
  "lint-staged": {
    "*.{js, jsx}": [
      "yarn prettify",
      "yarn lint --fix",
      "git add"
    ]
  },
```
https://github.com/decred/politeiagui/blob/master/package.json#L68-L79

This guarantees all files are formatted and linted on each commit.

P.S: working on it

PR #3248, opened 2021-02-19
Prettify codebase.

amassarwi:
All changes can be replicated with `yarn prettify`.
This needs to go in before integrating husky (#3239) which will ensure all committed code
is prettified on each commit.


alexlyp (approved):
ran yarn prettify as well and results matched

PR #3250, opened 2021-02-19
Add husky pre-commit hook to run yarn prettify & lint

amassarwi:
This installs husky v5 https://typicode.github.io/husky/#/ 

It also adds `postinstall` script to`package.json` to run `husky install` after yarn install automatically as mentioned in the 
installation guide linked above.

To test I think you need you need to:
 - `yarn install`.
 - it should run `husky install` automatically. 
 - add some empty lines to any `.jsx`, `.js`, `.css` or `.json` file.
 - try to commit and the empty lines to should be prettified automatically.


.husky/pre-commit
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+pretty-quick --staged
```
amassarwi, line 4: I have installed https://github.com/azz/pretty-quick to run prettify on staged files
as found on https://prettier.io/docs/en/precommit.html#option-2-pretty-quickhttpsgithubcomazzpretty-quick

PR #3252, opened 2021-02-21
[tickets] Show different message when less tickets bought

alexlyp:
Due to utxos requirements for ticket purchasing, sometimes when a user requests more tickets than they are able to construct it only purchases as many as it's able to.  This message alerts the user to this situation.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -280,6 +280,7 @@ export const publishTransactionAttempt = (tx) => (dispatch, getState) => {
 export const PURCHASETICKETS_ATTEMPT = "PURCHASETICKETS_ATTEMPT";
 export const PURCHASETICKETS_FAILED = "PURCHASETICKETS_FAILED";
 export const PURCHASETICKETS_SUCCESS = "PURCHASETICKETS_SUCCESS";
+export const PURCHASETICKETS_SUCCESS_LESS = "PURCHASETICKETS_SUCCESS_LESS";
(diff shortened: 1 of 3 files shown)


app/actions/ControlActions.js
```diff
@@ -385,8 +385,12 @@ export const newPurchaseTicketsAttempt = (
     }
     // save vsp for future checking if the wallet has all tickets synced.
     dispatch(updateUsedVSPs(vsp));
-
-    dispatch({ purchaseTicketsResponse, type: PURCHASETICKETS_SUCCESS });
+    const numBought = purchaseTicketsResponse.getTicketHashesList().length;
+    if (numBought < numTickets) {
+      dispatch({ purchaseTicketsResponse, numAttempted: numTickets, type: PURCHASETICKETS_SUCCESS });
```
vctt94, line 390: can dispatch `PURCHASETICKETS_SUCCESS_LESS` directly.

app/reducers/snackbar.js
```diff
@@ -654,8 +659,14 @@ export default function snackbar(state = {}, action) {
         case PURCHASETICKETS_SUCCESS:
           values = {
             numTickets: action.purchaseTicketsResponse.getTicketHashesList()
-              .length
+              .length,
+            numAttempted: action.numAttempted
           };
+          if (values.numAttempted && values.numTickets < values.numAttempted)  {
+            message = messages[
+              "PURCHASETICKETS_SUCCESS_LESS"
```
vctt94, line 667: no need with above comment

matheusd (approved):

PR #3256, opened 2021-02-22
Fix ticket showing as solo after purcahase

vctt94:
This PR fixes ticket showing as solo after purchase


alexlyp (approved):
nice works great!

PR #3267, opened 2021-02-25
Reverse txhash to avoid missing Transaction errors

alexlyp:
There were pretty constant `wallet.TransactionSummary(XXXX): item does not exist` errors being logged, but no noticeable problems in the UI.  Apparently the ticket spender hash that was being requested here was reversed so the tx was thought to be missing.


amassarwi (approved):
👍

matheusd (approved):

PR #3278, opened 2021-03-01
Functional components/layout/* with css modules

amassarwi:
~~waiting on #3215 as it conflicts with `layout.less`.~~

This refactors `components/layout/*` components to functional approach and splits `layout.less` to **many** local, component specific, encapsulated css modules. 

It also adds few className props to custom style parts of `TabbedPage` & `TabbedPageTab`, before this commit we used to override classes in too many different `less` files but with this approach wrapper will provide specific className defined in it's css module instead of overriding children css classes in wrapper's `.less` file.

This also deletes couple of redux connectors & `MyTickets.less` which wasn't used anymore.

Part of #2438 & #2439.


app/hooks/useModalVisible.js
```diff
@@ -0,0 +1,12 @@
+import { useSelector } from "react-redux";
+import * as sel from "selectors";
+
+const useModalVisible = () => {
```
bgptr, line 4: This custom hook is not used anywhere.

bgptr (commented):
LGTM! Left a tiny comment. Nice work!

app/hooks/useModalVisible.js
amassarwi: It's used in `BlurableContainer.jsx`:

https://github.com/decred/decrediton/pull/3278/files#diff-24b84bdbeaf013c28b29eec0c133f6dae393ad8231a1caccdb62928b18d70229R6

bgptr: Yes, but it imports its own custom hook from `./hooks` and not from `hooks`

amassarwi: hrmmmm

amassarwi: You are 💯 right, thanks for this nice catch!

matheusd (approved):
Tested fine

alexlyp (approved):

PR #3291, opened 2021-03-06
Ditch Documentation.less & locale redux connector.

amassarwi:
Part of #2438 & #2439.


app/components/views/PrivacyPage/Privacy/ConfigMixer/ConfigMixer.jsx
```diff
@@ -1,14 +1,13 @@
 import { FormattedMessage as T } from "react-intl";
 import { Documentation, Subtitle } from "shared";
 import { classNames } from "pi-ui";
-import style from "../Privacy.module.css";
+import styles from "./ConfigMixer.module.css";
```
amassarwi, line 4: **NOTE:**
We have this in many other places, where we import parent's css module to style nested/children components.
I guess the reason is that this how `.less` were defined, in css modules tho all these components need to get there own css module to achieve maximum encapsulation.

alexlyp (approved):

PR #3319, opened 2021-03-18
Reorganize overview balances

### LINKED ISSUE #3271: Overview Balances
@davecgh idea from matrix:
I think it might make sense to just show to cateogries on the overview which consist of Available (spendable) and Locked (unspendable).  The reason why the funds are locked might be because unconfirmed, immature, locked by tickets, has a time lock, etc.

So available can remain as is.  Locked would now be the combination of anything "unspendable" currently. 

Then we can add a tooltip to show the various categories when Locked is hovered.  

In this case unconfirmed can be removed.

bgptr:
closes #3271

### DIFF
--- a/app/components/views/HomePage/HomePage.module.css
+++ b/app/components/views/HomePage/HomePage.module.css
@@ -19,7 +19,7 @@
 }
 
 .overviewSpendableLockedWrapper {
-  padding-top: 96px;
+  padding-top: 146px;
 }
 
 .overviewTransactionsTicket {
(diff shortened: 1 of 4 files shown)


alexlyp:
Mostly looks good.  Though I think you should ditch the Voting Authority Balance listing under the Locked Tooltip.  Voting Authority is more related to stake info than real balances imo.  Locked by tickets is the accurate amount to be tracked for tickets that are currently waiting to vote.

bgptr:
@alexlyp  ok, I ditched the Voting Authority Balance listing under the Locked Tooltip. 
Thanks for the review.

alexlyp (approved):

PR #3322, opened 2021-03-19
[deps] Upgrade to webpack 5

matheusd:
~**Rebased on top of #3318**~

This upgrades the project to the most recent major version of webpack (5).

Notable changes:

- Updated webpack dependencies and plugins
- Changed static assets to use the new "assets" type of loader
- Removed unnecessary entries
- Simplified config
- Static assets are now going through webpack (instead of local file access and using the copy plugin for packaged versions)


amassarwi (commented):
Having the same comms issues reported by @alexlyp on matrix,
otherwise LGTM.

alexlyp (approved):

Issue #3321, opened 2021-03-19
[bug] Modals: cut off left side shadow.


amassarwi:
![image](https://user-images.githubusercontent.com/10324528/111827336-6f8e6c00-88f2-11eb-8d46-e3d1eada27d6.png)

bgptr:
on this

amassarwi:
@alexlyp addressed in #3323 and can be closed.

PR #3323, opened 2021-03-19
List UTXO’s modal

### LINKED ISSUE #3281: List UTXO’s view / modal update

Basically we left the link at the same place as before. Just changed the way of displaying the list. Now it is within a modal.
image

**An implementation right now:**
image

**An updated design:**
image

**An example with scroll bar and maximum practical height for the modal, in case there's more than fits the view:**
image

bgptr:
Closes #3281 and #3321


alexlyp:
looks mostly good.  Only issue I notice is in dark mode the font colors look a little off and the top bar (that has UTXO and Value) is not colored correctly.

alexlyp:
I also think it's worth matching the minimum spacing on the top and bottom of the modal to match the sides.  Seems a little tight on the top and bottom, 20 px seems like a good amount of margin/padding there

bgptr:
@alexlyp Thanks for the review. I've improved the colors in dark mode, but I am not sure about your margin/padding comment. Could you attach a screenshot?

alexlyp:
Can't unfortunately.  But basically I'm see the top and bottom of the modal come too close to the very edge of the electron window.  Feel like that space should be wider.

bgptr:
@alexlyp Now I see what you mean. I've improved the spacing.

alexlyp:
Ah much better thanks for the update @bgptr

alexlyp (approved):

Issue #2517, opened 2020-06-05
Change react-select dependency


vctt94:
Right now our react-select dependency is at version `1.0.0`, while their repo is at `3.1.0`.

Talking at matrix and @amassarwi suggested we change it to use pi-ui's select input.

So after https://github.com/decred/pi-ui/issues/264 is closed, we can make this change, as it may end with some breaking change and duplicated work, if done before it.

amassarwi:
Can be prefixed with [pi-ui]

PR #3326, opened 2021-03-21
Updated `Transactions` views layout

### LINKED ISSUE #3280: Send view update
Recently we've been working on some updates for the Send view, so this is what we ended up with:

**How it looks in implementation right now:**
image

**How an updated design looks like:**
image

Well, listing the things that has been changed:

**Common**
- Changed Details section. Now there are no 2 separate sections (white and gray), only common white one.

- Reordered fields according to users’ logical chain of actions. Previously there were From-To-Amount fields, now we changed to From-Amount-To order.

- Redesigned small buttons. Now they all are icons over white/blue square buttons (also added shadows if the button is white)
 
- Changed the behavior after clicking on “List UTXOs” link over the up right corner of the section. Now it will open a modal (instead of a page) with the list of all unspent utxos. (see #3281 )

- We also moved Send button inside the section (actually the common pattern is when the button is below sections, but it’s an exception here). That has been done for the situations when users add more than 1 recipient address, so the button has its static place and doesn’t move down with the bottom line of the section.         

**“Account” Field**
-  Now balance preview is under an account name + adeded “Balance” title.

- Now there is one more small button (Send All Funds From Selected Account) next to “Send Funds To Another Account” one. It was next to Amount field before.

- Within the dropdown show balances under account names , not in one row like before.
image

**“Amount” Field**
- Amount field differs a bit from all other fields. We needed to create a field when it’s possible to enter 2 sized numbers, as that’s how we mostly show amount. It means that after writing a dot the size of numbers has to be automatically reduced from 20px to 13px.
image

- Also added  some new information regarding Account Balance, % of Account Balance. It shows how many % of existing balance in user’s account this user is going to send.

**“Send To” Field**
- The field was changed to the TextInput component from the component library.
- Added “Paste” button into the “Send” field. On a click, an address from the clipboard will be pasted. It means that users don’t need to make more than one action to paste a copied address.
image

When an address pasted, Paste button disappears.
image

- The button that adds new recipients addresses (the one with ‘+’ icon) basically remained the same as before if we talk about its functionality. We just changed the visual side, now it’s blue.
After clicking on it, new empty “Amount” and “Send To” fields for one more recipient appears below the 1st one and the serial number of a recipient appears above every Send To field, including the 1st one. Also one more thing is that such fields have Delete (cross) button instead of Add (plus) one. If you click on the Cross one, you will delete this new recipient.

On the example below you can see filled addresses of the 1st and 2nd recipients and an empty field for the 3rd one.
image

Previously it was like this (I mean in implementation right now)
image

@linnutee tell me if I forgot something.

### LINKED ISSUE #3282: Receive view update

Reorganized space in the section a bit. Changed the view of fields (now we use the common component from piui library) and put them in a row.
Removed the titles of Copy and View QR buttons.

**Right now in implementation:**
image

**An updated design:**
image

### LINKED ISSUE #3283: Export view update

Changed the view of fields, now we use the common component from piui library. Also more detailed information that previously were in the dropdown is now under information icon above right up corner (so removed dropdown feature in information at all)

**In implementation right now:**
image

**New design:**
image

The modal with detailed information appears, as soon as you clicked on this info icon.
image

### LINKED ISSUE #3334: [bug] tabbed page header on smaller screens
On smaller screens, the description in the tabbed page header should be hidden.
DeepinScreenshot_select-area_20210324161020

bgptr:
Closes #3282.
Closes #3283. 
Closes #3334. 
Closes #3280.

~~This diff should be merged after #3278. After it, I will handle the merge conflicts around `Layout.less`.~~
Notes:
- I skinned the account select input as it is in the design, but I did not replace it with a  `pi-ui` component. I think it should be handled in a separate PR globally. After this PR, we should take care of #2517.
- I did not implement the multiple inner styles of the amount input. There is an already opened issue about it:  #1511.
- icon of the "Send funds to another account" button was replaced by the account icon as @mariapleshkova and @linnutee suggested on matrix.


amassarwi:
> I skinned the account select input as it is in the design, but I did not replace it with a pi-ui component. I think it should be handled in a separate PR globally. 

Good point mate, worth creating an issue, it would be great if we replace all input components with generic input
imported from `pi-ui`.

bgptr:
@amassarwi I found that there is already an open issue about this. I linked it in the description.

alexlyp:
Overall looks really good.  Only thing that looks a little wacky is the "Paste" text in the button looks off to my eye.  Like the text is too wide or something.

app/components/views/TransactionsPage/ExportTab/hooks.js
```diff
@@ -1,14 +1,17 @@
 import { useSelector, useDispatch } from "react-redux";
 import * as sel from "selectors";
 import * as sta from "actions/StatisticsActions";
+import { useIntl } from "react-intl";
 
 export function useExportTab() {
   const exportingData = useSelector(sel.exportingData);
   const dispatch = useDispatch();
   const exportStatToCSV = (opts) => dispatch(sta.exportStatToCSV(opts));
+  const intl = useIntl();
```
amassarwi, line 10: 👍

amassarwi (approved):
code LGTM!

bgptr:
Thank you guys for the review.
@alexlyp I used `pi-ui` `Button` component and I did not notice it uses Arial by default. Good catch, thx! I've fixed it.


#3336 broke the layout a little bit. I tried to hunt down every misalignment, but if a fresh eye could check this PR again, it would be great.

amassarwi:
Testing...

amassarwi (commented):
Looks great!!!!

Hit the following warnings in console - the first one seems to happen on master as well:

image

bgptr:
@amassarwi Thanks for the test. I fixed the warnings.

alexlyp (approved):

PR #3335, opened 2021-03-25
App container: fix ipcRenderer listeners not updating bug.

amassarwi:
This fixes a bug in the App container when can close check listener didn't update when provided canClose updated.
This also re-arranges the App container (see second commit) and moves all state related logic to the `useApp` custom hook.


alexlyp (approved):
tACK

PR #3336, opened 2021-03-25
Align decrediton's global `box-sizing` property with `pi-ui`.

amassarwi:
This gets rid of a styling hack which was used to resolve a global styling conflict between `pi-ui` and
`decrediton`.

Before this commit:
```
/* TODO: delete box-sizing: content-box;
// I added this ugly `:not` workaround to be able to skip pi-ui's classes
// which are prefixed with 'styles_'.
// Ideally we should use `* { box-sizing: inherit; } and html { box-sizing: border-box }`(already done in pi-ui)
// we shouldn't override it in a global level - we should take care of this while migrating this file to a css module.*/
*:not([class^="styles_"]),
*:not([class^="styles_"])::before,
*:not([class^="styles_"])::after {
  box-sizing: content-box;
}

[class^="styles_"] > * {
  box-sizing: inherit;
}
```

As mentioned in the code comment above; before this diff decrediton's needed `box-sizing: content-box;` which conflicted with `pi-ui`'s which defines `* { box-sizing: inherit; } and html { box-sizing: border-box }`.

After ditching `box-sizing: content-box;` I needed to tweak width/height property of many parts of decrediton.

### DIFF
--- a/app/components/SideBar/MenuBottom/MenuBottom.module.css
+++ b/app/components/SideBar/MenuBottom/MenuBottom.module.css
@@ -6,8 +6,8 @@
 }
 
 .shortSeparator {
-  height: 7px;
-  margin-bottom: 15px;
+  height: 15px;
+  margin-bottom: 12px;
   background-image: var(--menu-arrow-up);
   background-position: 50% 8px;
   background-repeat: no-repeat;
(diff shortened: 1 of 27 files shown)


matheusd (approved):

alexlyp:
Not sure if it was introduced in this PR, but seeing a misalignment in the Extended PubKey area in Account Details

alexlyp:
Yah confirmed that it was introduced in this PR.

amassarwi:
@alexlyp good catch, updating it.

amassarwi:
@alexlyp updated 👍

alexlyp (approved):
tACK

PR #3346, opened 2021-03-29
[trezor] Switch to iframe implementation

matheusd:
~**Rebased on top of #3322**~

This switches the trezor integration implementation to use the iframe approach as recommended by the sample trezor/electron demo.

The main benefit for this approach is being able to move back the firmware update action from the main process back to the renderer process, in turn removing most dependencies from the main process code.

In order to allow using the iframe approach, a new html file with accompanying js code needs to be generated and included in Decrediton's package (`iframe.html`, `trezor-iframe.js`). These files are generated into `app/dist-trezor` by a new webpack stage (`yarn build-trezor`, configured in `webpack.config.trezor.js`). The trezor-connect library is then configured to load and use this new code.

The `yarn build` target was modified to automatically run the required trezor build stage in production builds. **However**, for the moment, dev builds (`yarn dev`) do **not** automatically run it.

The above means that devs working with trezor will need to run `yarn build-trezor`. Devs that don't run the previous command will see a warning when running `yarn dev`, but the wallet should still work the same. I chose not to include `yarn build-trezor` in `yarn dev` in order to speed up dev builds, since the warning is harmless during development and the iframe code shouldn't change frequently anyway. I can include it back if other people feel strongly enough about it.

As of this PR, all dependencies have been moved from `app/package.json` back to the root `package.json` file, so there are no duplicate or incompatible modules across `node_modules/` and `app/node_modules` anymore.


matheusd:
I intended to open this as a draft pr until #3322 is merged, but ended up misclicking. No need for a deep review until that PR is merged.

alexlyp (approved):

alexlyp:
@JoeGruffins can you test this as well?

JoeGruffins (approved):
Looks great!

PR #3349, opened 2021-03-30
[redesign] Purchase tickets views

### LINKED ISSUE #3286: Purchase tickets view update

**How it's already implemented:**
image

**New design:**
image

**_Staking Overview section_**
- Changed the alignment (from left to center)
- Changed dropdown arrow orientation. Now it points down when the section is collapsed, and up (and getting blue) when expanded

-  Changed a bit the expanded view: now there are 3 items in each row

**How implemented:**
image

**New design:**
image

**_Purchase Tickets section_**
- Use Legacy VSP checkbox is now inside the section and closer to VSP field, right under it
- Changed input fields to the appropriate components from piui library
- Generally tidied up the whole section
- VSP Fee was moved to the “checkout” area above “Total” and “Remaining” amount, a fee icon was changed to the title “VSP Fee”. It should always be displayed and the fee is 0% by default.
- Changed Revoke button style from сontainedButton - Blue to the outlineButton one (white with blue stroke), so now when Purchase button is active they differ from each other

**How implemented:**
image

**New design:**
image

After selecting VSP, new checkbox “Always use this vsp” appears and if the fee for this VSP is more than 0%, the number of % has to be colored in Orange (#ED6D47), so users can easily notice it.
image

When Use Legacy VSP checkbox is checked, an additional gray area appears under the section + info icon above right up corner also appears (the same as it is implemented right now) 
image

**_Automatic Ticket Purchases section_**
- Now it has only a toggle button and the Settings icon on the right. All those fields as VSP, From, Balance can be reached when we click on this Settings icon.
When we click on it, a modal with these fields appears.
- If you click on the toggle button to turn automatic purchase on without clicking on the Settings icon before, the modal (that I mentioned above) appears, because this automatic purchase can’t be turned on without information from those fields.

**How implemented:**
image

**New design:**
image

**The Settings modal**
image

### LINKED ISSUE #3344: Purchase ticket form minor misalignments.
Screenshot 2021-03-28 at 20 57 02

### LINKED ISSUE #3384: Purchase Tickets info panel does not update after ticket purchase
After purchasing a ticket, it seems the ticket purchase info (in the purchase tickets page) does not update to account for the new maximum number of possible purchases.

In the screenshot below, I've just purchased some tickets and thus my balance has lowered but the "Amount" field hasn't updated to show the warning that I can't purchase this new number of tickets and the "Purchase" button is still enabled (notice how my balance is 354 DCR while 5 tickets would cost 438 DCR).

Also, the "DCR" label for the "Total" field is spilling over into the "Remaining" field.

image

bgptr:
Closes #3286
Closes #3344
Closes #3384

In this diff, I've implemented the new design of Purchase tickets views. Above that:
- add shared `PurchaseTicketsForm` ,`AutoBuyerSettingsModal` `TicketAutoBuyerForm` components
- introduced the `PiUiButton` component, which is a wrapper for the `pi-ui`'s `Button` component with decrediton styles.
- update and expand tests


matheusd (approved):
Looks very nice! Confirmed it fixed the issue in #3384

bgptr:
@alexlyp rebased

alexlyp (approved):

PR #3351, opened 2021-03-30
Fix webpack css modules & pure css configs.

amassarwi:
Our webpack 4.* config didn't work as is becuase of `css-loader` moving `localIdentName` option to be under `modules` instead of directly under options - see https://stackoverflow.com/a/57903310/8148239:

webpack 4:
```
 {
            loader: "css-loader",
            options: {
              sourceMap: true,
              importLoaders: 1,
              localIdentName: "[local]__[hash:base64:5]",
              modules: true
            }
          }
```

webpack 5:
```
"style-loader",
          {
            loader: "css-loader",
            options: {
              sourceMap: true,
              importLoaders: 1,
              modules: {
                localIdentName: "[local]__[hash:base64:5]"
              }
            }
          }
```

### DIFF
--- a/webpack.config.development.js
+++ b/webpack.config.development.js
@@ -41,8 +41,7 @@ export default merge(baseConfig, {
               sourceMap: true,
               importLoaders: 1,
               modules: {
-                // Adds original className as prefix to the hashed css module
-                // class names.
+                // Prepend the original class name in dev mode to ease debugging.
                 localIdentName: "[local]__[hash:base64:5]"
               }
             }
--- a/webpack.config.development.js
+++ b/webpack.config.development.js
@@ -29,9 +29,9 @@ export default merge(baseConfig, {
   },
 
   module: {
+    // CSS files injected directly in the DOM.
     rules: [
       {
-        // CSS files injected directly in the DOM.
         test: /\.css$/,
         use: [
           "style-loader",
@@ -47,7 +47,24 @@ export default merge(baseConfig, {
               }
             }
           }
-        ]
+        ],
+        include: /\.module\.css$/
+      },
+      {
+        test: [/\.css$/],
+        use: [
+          {
+            loader: "style-loader"
+          },
+          {
+            loader: "css-loader",
+            options: {
+              sourceMap: true,
+              importLoaders: 1
+            }
+          }
+        ],
+        exclude: /\.module\.css$/
       }
     ]
   },
--- a/webpack.config.development.js
+++ b/webpack.config.development.js
@@ -30,8 +30,39 @@ export default merge(baseConfig, {
 
   module: {
     rules: [
-      // CSS and Less files, injected directly in the DOM.
-      { test: /\.css$/, use: [ "style-loader", "css-loader" ] },
+      {
+        test: /\.css$/,
+        use: [
+          "style-loader",
+          {
+            loader: "css-loader",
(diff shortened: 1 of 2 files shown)


matheusd:
What's the point of configuring `identNames` or otherwise changing to this config (vs the current one)? The only difference I could see is that in development mode there's a prefix with the class name, while in production mode the result ends up the same.

amassarwi:
@matheusd Hrmm I had the exact same thought first, but then I noticed that we actually differentiate between css modules and pure css files - note the `exclude` & `include` fields.
https://github.com/decred/decrediton/pull/3351/files#diff-eec67ebf4011df965ec6c8a0a5b05ac578aa7015c6738998381f6b078af6c7ceR44

https://github.com/decred/decrediton/pull/3351/files#diff-eec67ebf4011df965ec6c8a0a5b05ac578aa7015c6738998381f6b078af6c7ceR59

amassarwi:
Hrmm with the current config the devlopment classes look like the following:
image

Also, I'm pretty sure with your minimal config both `*.module.css` & `&.css` files considered as modules and their
classes are being hashes which will break global styling files ie `main.css`.

matheusd:
> Also, I'm pretty sure with the minimal config both *.module.css & &.css files considered as modules and their
classes being hashes which will break global styling files ie main.css.

It doesn't break the sources. The `modules` options by default only applies to files named `*.module.*`, there's no need for different rules for modules and non-module css files (see [webpack docs](https://webpack.js.org/loaders/css-loader/#modules)). 

You can verify this is true by packaging the current master: the style.css file contains the unchanged class names for standard css files:

image

And the body element (for example) correctly contains the css rules from `main.css` and even has the correct source map (without having to enable the `sourceMap` option):

image

To get the class prefix name in development mode, it's _sufficient_ to add _just_ `localIdentName` to the development config, without all the other changes:

```diff
diff --git a/webpack.config.development.js b/webpack.config.development.js
index bb4e075d..f01377d5 100644
--- a/webpack.config.development.js
+++ b/webpack.config.development.js
@@ -31,7 +31,18 @@ export default merge(baseConfig, {
   module: {
     rules: [
       // CSS and Less files, injected directly in the DOM.
-      { test: /\.css$/, use: [ "style-loader", "css-loader" ] },
+      { test: /\.css$/, use: [
+        "style-loader",
+        {
+          loader: "css-loader",
+          options: {
+            modules: {
+              // Prepend the original class name in dev mode to ease debugging.
+              localIdentName: "[local]__[hash:base64:5]"
+            }
+          }
+        }
+      ] }
     ]
   },
   resolve: {
```

amassarwi:
@matheusd I see, thanks for clarifying will update the devlopment config.

amassarwi:
@matheusd updated

amassarwi:
hrmmm now it's broken

amassarwi:
The minimal dev config suggestion didn't work as it seems to apply the custom rule on all files - `*.module.css` & `*.css`.

I was able to make it work by explicitly specifying the pure css files rule as well for the devlopment env:
https://github.com/decred/decrediton/pull/3351/files#diff-11b2debd2ccb75181542fb336aed9674b4aef05560f69b1508f7c985225b9d39

matheusd (approved):

alexlyp (approved):

PR #3357, opened 2021-04-01
[deps] Update development related dependencies

matheusd:
~**Rebased on top of #3346**~

This updates most packages to use their latest version. The only packages _not_ updated are related to major UI elements (react, i18n), which might need more extensive code changes, so those are left for a future PR.

Notable changes:

- Removed our custom `node-addon-loader` in favor of the webpack standard `node-loader`
- Uncoupled the electron webpack config from the base webpack config (the electron config overrides most options anyway)
- Updated the electron webpack config so that `yarn build && yarn start` correctly work under windows
- Changed from using `fbjs-scripts checkEngine` to use the standard package.json `engines` entry, updated to minimum required versions of `node` and `yarn`
- Ran `yarn-deduplicate` to remove duplicate entries in `yarn.lock` 
- Removed deps from `package.json` that were either not used anymore or are implicitly required by other deps
- Allowed all `chrome-extension://` requests in development mode to fix extension-related warnings
- Regenerated protobuf files (due to dependency upgrades)
- Upgraded winston to latest major version (finally gets rid of the related dev startup warning)


alexlyp:
`yarn dev` seems to work fine, but seeing `Unchecked runtime.lastError: "sync" is not available in this instance of Chrome` in the console logs.

alexlyp:
Tested and works on all 3 dev, build && start and pacakge.

alexlyp (approved):

matheusd:
> yarn dev seems to work fine, but seeing Unchecked runtime.lastError: "sync" is not available in this instance of Chrome in the console logs.

That error is due to react-dev-redux being broken for newer electron versions: https://github.com/zalmoxisus/redux-devtools-extension/issues/767

PR #3360, opened 2021-04-04
Change "Immature Stake Generation" label to "Immature Staking Returns"

bgptr:
This diff implements @davecgh idea from matrix:
I wonder if the "Immature Stake Gen" label should be changed to "Immature Vote Returns" or similar.
I suspect "stake gen" is not at all obvious to 99.9999% of people who aren't devs.


alexlyp (approved):

xaur:
@bgptr I suggest to only use term "Rewards" to not amplify the already popular overfocus on "profits" or "returns" (some even call it "dividends" or "passive income").

Similar to https://github.com/decred/dcrstakepool/issues/582

PR #3345, opened 2021-03-29
[redesign][governance] proposal details view.

amassarwi:
Main changes:

- Split `ProposalsDetials.module.css` helpers styling into components own css modules.
-  Add vote end event & blocks left number.
- Define proposal status & vote status as constants.
- Add Join component to display subtitle separated fields.
- Add global helper css classes.
- Ditch `OverviewField` component.
- Replace `EligibleTickets` component with new `VoteSection`.
- Delete `TxHsitory/EligibleRow.jsx` and it's styling.
- Move Vote Pref. component to `VoteSection` according to new specs.
- Improve viewedProposal state obj to avoid duplicated fields.
- Declarative Proposals Details Page helpers.
- Replace shared classes with new components (ie `VoteButton` & `NoTicketsMsg`)
-  Finish Vote Preference section & introduce `VotedCheckmark`.
- Rename ChooseOptions component to VotePreference.
- Rename ChooseVoteOption to VotePreferenceWrapper.
- Move CastVoteModalButton to a separate dir with css module.
- Tweak proposal body & politeia proposal button/link.
- Re-arrange proposal card.
- Tweak subtitle.
- Add proposal card back button.
- Add title tooltip.

**Note:** before reviewing the code I'd suggest comparing `components/viewed/ProposalDetailsPage` dir and it's structure on master and on this branch.

Screenshots:

Voting proposal & user voted:
image

Abandoned:
Screenshot 2021-04-05 at 3 11 40

Rejected:
Screenshot 2021-04-05 at 3 11 53

Closes #3288.

PR #3370, opened 2021-04-08
[governance] parse & store proposal's body on redux level.

amassarwi:
This is a small followup after #3345 moving proposal body parsing logic from component
to governance redux actions layer. 

After this commit `ProposalDetailsPage` gets the proposal's body as part objects returned from the custom hook.

### DIFF
--- a/app/helpers/politeia.js
+++ b/app/helpers/politeia.js
@@ -1,0 +1,9 @@
+// politeiaMarkdownIndexMd returns markdown text from the payload of a politeia
+// proposal file that corresponds to its index.md). This was extracted from the
+// helpers.js file of politeia. Assumes the payload has been converted from
+// base64 into bytes.
+export const politeiaMarkdownIndexMd = (payload) => {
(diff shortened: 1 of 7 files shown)


app/helpers/politeia.js
```diff
@@ -0,0 +1,9 @@
+// politeiaMarkdownIndexMd returns markdown text from the payload of a politeia
+// proposal file that corresponds to its index.md). This was extracted from the
+// helpers.js file of politeia. Assumes the payload has been converted from
+// base64 into bytes.
+export const politeiaMarkdownIndexMd = (payload) => {
+  const text = decodeURIComponent(escape(payload));
+  console.log(text);
```
matheusd, line 7: Extraneous log

amassarwi: oh, good catch!

amassarwi:
@matheusd updated

matheusd (approved):

PR #3377, opened 2021-04-09
Dismiss blur effect instantly after ticket purchase modal closing

### LINKED ISSUE #3368: [staking] Blur on screen when purchasing ticket
When purchasing a ticket, the view gets bluered, while showing the modal, but after typing the passphrase and submiting, the modal gets dismissed, but the bluered keeps showing:

image

bgptr:
Closes #3368

I'm not very happy with this solution, but I could not figure out what causes the sync delay. The redux store updates fine, but the selector reads the value of `modalVisible` wrong at first, and after some delay, it is updated with the right value. 
I did not notice this blur lingering issue with other modal though.

### DIFF
--- a/app/reducers/control.js
+++ b/app/reducers/control.js
@@ -305,6 +305,11 @@ export default function control(state = {}, action) {
         purchaseTicketsError: null,
         purchaseTicketsRequestAttempt: true,
         numTicketsToBuy: action.numTicketsToBuy,
+        // set modalVisible to false to hide blur effect instantly.
+        // Without this, it is dismissed with delay. Some hint for later
+        // investigation: the redux store updates fine, but the selector reads
+        // the value of `modalVisible` wrong at first, and after some delay,
+        // it is updated with the right value.
         modalVisible: false
       };
     case PURCHASETICKETS_FAILED:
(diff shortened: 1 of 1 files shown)


amassarwi:
As discussed on matrix:

not a bad workaround, other better option might be using `useLayoutEffect` instead of `useEffect` to apply/disable the blur effect.
which might fix the delay we are seeing:
https://daveceddia.com/useeffect-vs-uselayouteffect/ 

> When to useLayoutEffect
>
> The right time to useLayoutEffect instead? You’ll know it when you see it. Literally ;)If your component is flickering when state is updated – as in, it renders in a partially-ready state first and then immediately re-renders in its final state – that’s a good clue that it’s time >to swap in useLayoutEffect.This’ll be the case when your update is a 2-step (or multi-step) process. Do you want to “batch” a couple updates together before redrawing the screen? Try useLayoutEffect.I think of useLayoutEffect as the way to squeeze in a little extra work before React updates the DOM. “Hey, you’re making some changes already – could you throw this one in there too? Awesome.”

see politeiagui's useScrollToTop custom hook as an example:
https://github.com/decred/politeiagui/blob/a84c4299e73cbdd6d6bb74b90cd15f981bf5f8f0/src/hooks/utils/useScrollToTop.js

amassarwi (approved):
`useLayoutEffect` doesn't seem to improve the situation, so I'm fine with for the moment.

alexlyp (approved):

PR #3381, opened 2021-04-12
Show the correct amount of incoming DEX transactions

### LINKED ISSUE #3354: [dex] Incoming DEX transactions show as 0.000 DCR
When receiving an incoming atomic swap (or filled transaction on DEX) the amount received shows 0.000 DCR.  

When viewing the transaction details the output looks correct, but the input is also showing as 0.0000 instead of the amount it should be.

bgptr:
I found that the output in the redeeming transactions on DEX is generated as internal. In Decrediton during the tx normalizing the internal outputs are considered as change outputs. 

In this diff, if a received transaction has only one output and it is internal and its value equal to the amount prop of the decoded transaction data, it is assumed as an atomic swap redeem transaction and use its internal output's value as the amount.

The zero inputs (mentioned in #3354) on the transaction details page is a separate bug and it affects other types of transactions too. I handled this issue in this PR too.

Closes #3354.

### DIFF
--- a/app/components/views/TransactionPage/TransactionContent/TransactionContent.jsx
+++ b/app/components/views/TransactionPage/TransactionContent/TransactionContent.jsx
@@ -28,7 +28,7 @@ function mapNonWalletOutput(output) {
 
 function mapNonWalletInput(input) {
   const address = `${input.prevTxId}:${input.outputIndex}`;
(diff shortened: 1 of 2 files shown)


alexlyp (approved):

alexlyp:
nice work!

PR #3398, opened 2021-04-21
[pi-ui] Integrate TextInput component

bgptr:
This is a follow-up diff of #3326 and replaces text inputs with the pi-ui `TextInput` component.
`TextInput` component needed some additional functionality, which has been implemented in decred/pi-ui#326
~~Until it will be merged, I edited the package.json.~~ decred/pi-ui#326  and decred/pi-ui#338 have been merged and I upgraded the pi-ui package


app/components/views/GetStartedPage/AdvancedStartup/RemoteDaemonForm.jsx
```diff
@@ -50,6 +50,7 @@ const RemoteDaemonForm = ({
       </Label>
       <Input>
         <TextInput
+          id="prc-user"
```
matheusd, line 53: **rpc**-user

app/components/views/GetStartedPage/AdvancedStartup/RemoteDaemonForm.jsx
```diff
@@ -65,6 +66,7 @@ const RemoteDaemonForm = ({
       <Input>
         <PasswordInput
           required
+          id="prc-pass"
```
matheusd, line 69: **rpc**-pass

package.json
```diff
@@ -264,7 +278,7 @@
     "minimist": "^1.2.3",
     "mv": "^2.1.1",
     "node-polyfill-webpack-plugin": "^1.1.0",
-    "pi-ui": "https://github.com/decred/pi-ui",
+    "pi-ui": "https://github.com/bgptr/pi-ui#improve-textinput",
```
matheusd, line 281: Upstream PR was merged, so ready to change back.

matheusd (commented):
Getting an error when adding new output addresses in the send tab ("An effect function must not return anything besides a function...").

Adding then removing a new output address triggers a "func.apply is not a function".

Getting an error in the transaction history tab ("Prop id is marked as required...")

I noticed the font size of placeholders is slightly higher with this PR. Is this intended?

package.json
bgptr: @matheusd Thanks for the review, good catches.
The error in the sent tab was caused unfortunately by the pi-ui TextInput component. It was my bad, opened a new PR with the fix. Until it will be merged, I updated the package.json with the new fixing branch name.
The "Prop id is marked as required..." message was a warning. I fixed it and the similar warnings too.
The font size of placeholders is slightly higher indeed (16px) and it was dictated by the design.

bgptr: @matheusd I examined the placeholders again and you are right, something was weird about them. I figured out that the font family was inherited from `pi-ui`-s `TextInput` component and I missed overwriting it. 
I fixed them and a few misalignments that I found with fresh eyes.
Thanks.

matheusd (changes requested):
Stilll getting an error when trying to rename an account ("prop id required") but otherwise the other issues were fixed.

bgptr:
@matheusd I think I've hunted down all of the remaining warnings. Thanks for the testing.

matheusd (approved):
Working fine now!

amassarwi:
@bgptr need a tiny rebase

bgptr:
Rebased. In the `ExternalButton` component (introduced in the lastly merged #3497) the wallet module was imported directly and not from the `wallet-preload-shim` which was introduced in #3486. This is preventing the master to build at this moment, but this diff going to fix it.

PR #3405, opened 2021-04-22
fix css bugs

### LINKED ISSUE #3403: Some bugs in 1.6.2 release
Posting a list of bugs that I noticed during testing the 1.6.2 release:

1. Buttons alignment is a bit off in the launcher view. We decided to use the right-alignment in the wallet, so let's change it here as well. Also, please, swap the buttons, it should be Cancel-Continue if reading from left to right.
image
image

2. Arrow within dropdowns has to be turned up when clicked
image
image

3. Same issue with buttons alignment(move and swap):
image

4. Swap the buttons (for all pop ups that have buttons):
image
image
image
image
image

5. Add a spacing here:
image

6. Move the button to the right:
image

Also @linnutee you mentioned something about icons and gifs on loader. I didn't notice anything weird, so could you explain here? Thanks.

bgptr:
Closes #3403

### DIFF
--- a/app/style/ReactSelectGlobal.css
+++ b/app/style/ReactSelectGlobal.css
@@ -226,7 +226,6 @@
   height: 0;
   width: 0;
   position: relative;
-  transition: all var(--input-transition-time) var(--input-timing-function);
 }
 
 .Select-control:hover:not(:disabled) {
@@ -243,6 +242,7 @@
 .is-open .Select-control:hover:not(:disabled) .Select-arrow,
 .Select-arrow-zone:hover > .Select-arrow {
   border-top-color: var(--stroke-color-focused);
+  transform: rotate(180deg) translateY(4px);
 }
 
 .Select--multi .Select-multi-value-wrapper {
--- a/app/style/ReactSelectGlobal.css
+++ b/app/style/ReactSelectGlobal.css
@@ -51,7 +51,6 @@
   overflow: hidden;
   position: relative;
   width: 100%;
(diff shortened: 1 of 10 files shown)


app/style/ReactSelectGlobal.css
```diff
@@ -226,7 +226,6 @@
   height: 0;
   width: 0;
   position: relative;
-  transition: all var(--input-transition-time) var(--input-timing-function);
```
amassarwi, line 229: are these used in other places 🤔 ?

amassarwi (approved):
LGTM + 1 question.

app/style/ReactSelectGlobal.css
bgptr: @amassarwi Thanks for the review. 
This transition only affects the arrow icon. I've removed the line for the sake of uniformity. Other similar arrows aren't animated in the app.

amassarwi: i mean the theme variables: `input-transition-time` & `input-timing-function` - i'm not sure if they are used/needed in other places, if not we could ditch them from the theme files.

bgptr: Ahh, I see :) They are used in this file, but ditching them hasn't changed the behavior of the select control.

amassarwi: Noice, thanks 🥇

alexlyp (approved):

PR #3411, opened 2021-04-23
[dexc] Fix paths to libdexc.so

alexlyp:
This has become necessary due to macOS not properly resolving the path for the packaged version.
Also removes a testnet dex listening address at 127.0.0.2 since macOS doesn't appear to let you bind to those addresses automatically.

Confirmed on linux (yarn dev/start/package)
Confirmed on mac (yarn dev/start/package)
Confirmed on win (yarn start/package)


matheusd (approved):
tested yarn dev/build+start/package on linux and windows

PR #3421, opened 2021-04-26
Fix malformed tooltips

### LINKED ISSUE #3417: [transactions] Tooltip showing mixer is running on send is malformed
When the mixer is running we don't allow users to send transactions.  A tooltip is shown when the button is disabled due to this situation.  Currently the tooltip is malformed (tall and skinny) making the information hard to read.

bgptr:
Closes #3417


alexlyp:
Looks good, just needs the same for when a mixed purchase is running tooltip on Autobuyer switch and the Mixer button.

Also looks like there are 2 tooltips being shown for the Autobuyer switch in this case.  Should only show the Purchase/Mixer running message.

alexlyp (approved):

PR #3422, opened 2021-04-26
[mixer] Revert incorrect api.proto changes

### LINKED ISSUE #3414: [master] Nothing happens while mixer is on.
Running latest against latest `dcrd`/`dcrwallet` and when starting the mixer I just see the indicator that it's running
but I don't see any logs and it seems like the mixer isn't really working in the background.

alexlyp:
Closes #3414

This was causing the incoming req to dcrwallet to be malformed which led to the mixer not firing as expected.


alexlyp:
Closes #3414

matheusd (approved):

PR #3424, opened 2021-04-26
remove unlock wallet

### LINKED ISSUE #3420: [accounts]  Newly created accounts need to SetAcccountPassphrase
Currently, newly created accounts aren't locked.  They should be locked directly after creation with the private passphrase that is given for the create account request itself.

vctt94:
closes #3420.

The unlock is not necessary as we sent the passphrase through the grpc request so dcrwallet can make the unlock.


matheusd (commented):
This does't really fix the issue. One way you can try it out is to create a new account, then sign a message with an address from that account: you'll get an "account not encrypted" error.

matheusd:
This also affects LN wallet creation when creating a new account is chosen (vs using an existing account).

vctt94:
You are right. This should fix it now

matheusd (commented):
Manually creating new account and new LN account are working, but it seems DEX and Mixer accounts are still not encrypted after creation.

Mixer and DEX account creation are using `wallet.getNextAccount` directly instead of the one from control actions

vctt94:
RIght, I change them to use the getNextAccountAttempt, instead. Thanks

matheusd (approved):
The four account creation triggers (new account, LN account, dex account and mixer account) are now working as intendend. 

Thanks!

PR #3429, opened 2021-04-26
[dex] Adding label and styling to get fee page

alexlyp:


matheusd (approved):

app/components/views/DexPage/RegisterPage/RegisterPage.module.css
```diff
@@ -1,9 +1,35 @@
...
+.dexAddress input::placeholder {
+  font-size: 16px;
+}
+
 .error {
   color: var(--error-text-color);
   margin-top: 8px;
 }
+
+label {
```
amassarwi, line 20: this is too global and will be applied on **all** `<label>` tags cross the app, it won't hashed as it targets an html tag and not a class name.

amassarwi (changes requested):
Looks good except one suggestion.

app/components/views/DexPage/RegisterPage/RegisterPage.module.css
alexlyp: kk will change

PR #3432, opened 2021-04-27
update dex to 0.2.0-rc1, wait for webserver shutdown

chappjc:
Build libdexc with dcrdex v0.2.0-beta.

Also make `(*CoreAdapter).shutdown` wait for the webserver in addition to `Core`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/decred/decrediton
 go 1.16
 
 require (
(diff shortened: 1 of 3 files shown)


chappjc:
Here's some text I was considering for the enable DEX page, or even the start DEX page that is showed each time the user starts: https://gist.github.com/chappjc/85a7b9d27771cb104a8b723139464b6c

alexlyp (approved):

PR #3434, opened 2021-04-27
fix coinjoin not showing

vctt94:
I noticed when having coinjoin outputs, and no mixed account set on config, the screen turns blank, being not possible to leave it. 

This PR fixes it.


matheusd (commented):
I'm still getting a `onSendContinue is not a function` error when restoring a cspp wallet with this PR (without the PR it was just a blank screen).

Reloading the UI correctly loads the wallet, so it seems to be something in the state machine of the getstarted page.

matheusd (approved):
Working as intended on restores now.

Issue #3423, opened 2021-04-26
[accounts] Current account locking issues to fix


alexlyp:
Here are the current account locking issues that have been observed:
- Revoke Tickets is still currently unlocking/locking the whole wallet.  Though we need to discuss Revoke Tickets in general since it's kind of a weird case.
- VSP Fees are not being paid on ticket purchase and "account with unique passphrase is locked" errors are repeatedly shown by the vspClient on repeated attempts to check the status of the ticket.
- Sync Failed VSP Fees is unlocking/locking the whole wallet and also doesn't appear to be processing any fees.

PR #3435, opened 2021-04-27
[locking] Various account locking issues patched

alexlyp:
Closes a few of the issues notes in #3423

No longer unlocks whole wallet for Revoke Tickets.

Due to voting address needing to be unlocked for tickets to communicate with the VSPs.  Typically the voting account is currently the default account, but there are certain circumstances if using the Auto-buyer that the voting account is the purchase account.

### DIFF
--- a/app/actions/AccountMixerActions.js
+++ b/app/actions/AccountMixerActions.js
@@ -105,8 +105,7 @@ export const runAccountMixer = ({
       const mixerStreamer = await dispatch(
         unlockAcctAndExecFn(
           passphrase,
-          changeAccount,
-          null,
+          new Array(changeAccount),
           () =>
             runAccountMixerRequest(sel.accountMixerService(getState()), {
               mixedAccount,
--- a/app/actions/AccountMixerActions.js
+++ b/app/actions/AccountMixerActions.js
@@ -105,7 +105,7 @@ export const runAccountMixer = ({
       const mixerStreamer = await dispatch(
         unlockAcctAndExecFn(
           passphrase,
-          new Array(changeAccount),
+          [changeAccount],
           () =>
             runAccountMixerRequest(sel.accountMixerService(getState()), {
               mixedAccount,
--- a/app/actions/AccountMixerActions.js
+++ b/app/actions/AccountMixerActions.js
@@ -106,6 +106,7 @@ export const runAccountMixer = ({
         unlockAcctAndExecFn(
           passphrase,
           changeAccount,
+          null,
           () =>
             runAccountMixerRequest(sel.accountMixerService(getState()), {
               mixedAccount,
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -460,11 +460,20 @@ export const revokeTicketsAttempt = (passphrase) => async (
 ) => {
   dispatch({ type: REVOKETICKETS_ATTEMPT });
   const walletService = sel.walletService(getState());
-  const accountNum = 0;
+  const accounts = sel.balances(getState());
+  const accountUnlocks = [];
+  accounts.map((acct) => {
+    // just skip if imported account.
+    if (acct.accountNumber === Math.pow(2, 31) - 1) {
+      return;
+    }
+    accountUnlocks.push(acct.accountNumber);
+  });
+  console.log(accountUnlocks);
   try {
     const revokeTicketsResponse = await dispatch(
-      unlockAcctAndExecFn(passphrase, [accountNum], () =>
-        wallet.revokeTickets(walletService, passphrase)
+      unlockAcctAndExecFn(passphrase, accountUnlocks, () =>
+        wallet.revokeTickets(walletService)
       )
     );
     dispatch({ revokeTicketsResponse, type: REVOKETICKETS_SUCCESS });
@@ -1101,7 +1110,7 @@ export const unlockAcctAndExecFn = (
 
   // unlock wallet
   try {
-    accts.map(async (acctNumber) => {
+    await accts.map(async (acctNumber) => {
       console.log(" unlocking", acctNumber);
       await wallet.unlockAccount(
         sel.walletService(getState()),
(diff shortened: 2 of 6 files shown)


app/actions/ControlActions.js
```diff
@@ -1045,6 +1051,7 @@ export const LOCKACCOUNT_SUCCESS = "LOCKACCOUNT_SUCCESS";
 export const unlockAcctAndExecFn = (
   passphrase,
   acctNumber,
+  secondaryAcctUnlock,
```
vctt94, line 1058: I think we can make acctNumber an array, and unlock all sent. This way would avoid needing a third param, in case. What do you think?

alexlyp: yah i think i like the sounds of that, will change

app/actions/AccountMixerActions.js
```diff
@@ -102,8 +102,7 @@ export const runAccountMixer = ({
       const mixerStreamer = await dispatch(
         unlockAcctAndExecFn(
           passphrase,
-          changeAccount,
-          null,
+          new Array(changeAccount),
```
vctt94, line 105: This is creating an array with number of `changeAccount` positions.

Ex: if change account equals 5, it will create an array with 5 positions and arr[0] will be undefined.

Use brakets `[ ]` instead of `new Array`

app/actions/ControlActions.js
```diff
@@ -1067,35 +1087,51 @@ export const LOCKACCOUNT_SUCCESS = "LOCKACCOUNT_SUCCESS";
...
-      acctNumber
+    await Promise.all(
+      accts.map(async (acctNumber) => {
+        await wallet.unlockAccount(walletService, passphrase, acctNumber);
+      })
     );
     dispatch({ type: UNLOCKACCOUNT_SUCCESS });
   } catch (error) {
-    // no need to lock as unlock errored.
+    if (String(error).indexOf("invalid passphrase") < 0) {
```
matheusd, line 1125: I don't think this check is needed: you probably want to re-lock irrespective of the error that happened.

This can't (easily) happen right now, but consider if you're unlocking multiple accounts and one of them has a different passphrase: you'll end up not re-locking the accounts due to this check. This situation could happen in the future if we allow individual passphrases or if the user manually interacted with dcrwallet to set individual pwds.

app/actions/ControlActions.js
```diff
@@ -348,8 +348,10 @@ export const purchaseTicketsAttempt = (
         );
       }
       accountNum = account.encrypted ? account.value : null;
+      // Add default account unlock if account isn't default
```
matheusd, line 351: I suggest adding the reason for adding the default account here.

Something to the effect of "... because the voting address will be derived from the default account".

app/actions/ControlActions.js
```diff
@@ -394,9 +396,10 @@ export const newPurchaseTicketsAttempt = (
       csppPort: sel.getCsppPort(getState()),
       mixedAcctBranch: sel.getMixedAccountBranch(getState())
     };
-
+    // Add default account unlock if account isn't default
```
matheusd, line 399: Same as other inline comment

app/actions/ControlActions.js
```diff
@@ -451,17 +454,34 @@ export const REVOKETICKETS_ATTEMPT = "REVOKETICKETS_ATTEMPT";
...
-      }, 4000);
-      dispatch({ revokeTicketsResponse, type: REVOKETICKETS_SUCCESS });
-    })
-    .catch((error) => dispatch({ error, type: REVOKETICKETS_FAILED }));
+  const walletService = sel.walletService(getState());
+  const accounts = sel.balances(getState());
+  const accountUnlocks = [];
+  accounts.map((acct) => {
+    // just skip if imported account.
+    if (acct.accountNumber === Math.pow(2, 31) - 1) {
```
matheusd, line 467: Use `>=` instead of `===` because we could eventually have more than a single imported account.

Also, would be better to set this as a constant (either locally here or on `constants/decred.js`) so this is less magical.

app/actions/ControlActions.js
```diff
@@ -451,17 +454,34 @@ export const REVOKETICKETS_ATTEMPT = "REVOKETICKETS_ATTEMPT";
...
-      setTimeout(() => {
-        dispatch(getStakeInfoAttempt());
-      }, 4000);
-      dispatch({ revokeTicketsResponse, type: REVOKETICKETS_SUCCESS });
-    })
-    .catch((error) => dispatch({ error, type: REVOKETICKETS_FAILED }));
+  const walletService = sel.walletService(getState());
+  const accounts = sel.balances(getState());
+  const accountUnlocks = [];
+  accounts.map((acct) => {
```
matheusd, line 465: Switch from `map` to `forEach` since you're not using the output of the map (this avoids creating an unnecessary result object that ends up discarded)

matheusd (commented):
I see there are 3 instances in the code where all accounts are being unlocked now and they have some specific logic we might want to modify later on.

I suggest abstracting the 3 instances in a single `unlockAllAcctsndExecFn` (or whatever) so that's easier to keep track of.

app/actions/ControlActions.js
```diff
@@ -1067,35 +1082,144 @@ export const LOCKACCOUNT_SUCCESS = "LOCKACCOUNT_SUCCESS";
...
+      accts.map(async (acctNumber) => {
+        await wallet.unlockAccount(walletService, passphrase, acctNumber);
+      })
+    );
+    dispatch({ type: UNLOCKACCOUNT_SUCCESS });
+  } catch (error) {
+    await Promise.all(
+      // Need to try and lock all since 1 may have unlocked but not another?
+      accts.map(async (acctNumber) => {
+        if (dexAccount && acctNumber !== dexAccount.accountNumber) {
```
matheusd, line 1123: This is failing if the wallet doesn't have DEX setup. Needs to be `!dexAccount || acctNumber !== dexAccount.accountNumber`.

alexlyp: doh! good catch

matheusd (approved):

PR #3438, opened 2021-04-28
[dex] Allow selection of Dex account

### LINKED ISSUE #3419: [dex] Allow selection of Dex account
Instead of forcing people to create a new dex account, they should be able to select an existing account to become the dex account.

bgptr:
This diff allows users to select an existing account to become the dex account besides creating a new account.

Closes #3419

### DIFF
--- a/app/actions/DexActions.js
+++ b/app/actions/DexActions.js
@@ -442,3 +442,22 @@ export const createDexAccount = (passphrase, accountName) => async (
     dispatch({ error, type: CREATEDEXACCOUNT_FAILED });
   }
 };
+
+export const SELECT_DEXACCOUNT_ATTEMPT = "SELECT_DEXACCOUNT_ATTEMPT";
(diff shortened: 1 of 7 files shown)


matheusd (commented):
I'm clicking on "select dex account" but it's not advancing to the next step of dex setup

matheusd (approved):
Tested again after removing all dex config entries from the wallet config and restarting from sratch and it worked

alexlyp:
Overall works well!  

After discussing with @matheusd on matrix, due to the possibility of buys going back into the mixed account we need to exclude the mixed account from being in the dropdown for the dex account selection.  dcrdex will support privacy at some point and then we'll probably just have the dex accounts to be set to mixed (for selling) and unmixed (for buying).

bgptr:
@alexlyp updated

alexlyp (approved):

PR #3440, opened 2021-04-29
SetupWallet: Check selectedWallet.value for isTrezor.

### LINKED ISSUE #3431: Restoring watching only or trezor wallet triggers account reencryption
When restoring a watching only or trezor wallet, Decrediton still asks for the passphrase to perform the migration to per-account encryption

JoeGruffins:
This fixes the issue for me, but I think this is the second time this has happened. I don't know why there are these two sets of values, but I think I will use another pr to get to the bottom of it and make sure the code base only has to deal with one of these and this doesn't happen again.

closes #3431

### DIFF
--- a/app/components/views/GetStartedPage/SetupWallet/hooks.js
+++ b/app/components/views/GetStartedPage/SetupWallet/hooks.js
@@ -66,7 +66,10 @@ export const useWalletSetup = (settingUpWalletRef) => {
   }, [send]);
 
   const getStateComponent = useCallback(async () => {
-    const { error, isWatchingOnly, isTrezor } = current.context;
+    const ctx = current.context;
+    const { selectedWallet } = ctx;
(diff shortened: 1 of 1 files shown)


matheusd (approved):
Tested restoring watching only and trezor wallets

xaur:
@alexlyp at the risk of a false report, I noticed one oddity worth mentioning. Diff between tag `release-v1.6.3` and `master`'s commit d9a52f74 (last patch ported to 1.6.3) shows that this PR #3440 is the only one not included in v1.6.3:

https://github.com/decred/decrediton/compare/release-v1.6.3..d9a52f74

alexlyp:
@xaur yah dunno, must have been overlooked when creating the release-v1.6+ branch.  I've added it now and it will be included in the next patch release.

Issue #3433, opened 2021-04-27
[startup] Overview takes a few seconds to load


alexlyp:
On a basic startup I notice that the launcher screen goes mostly blank (except for the settings and logs button on top) for like 5-10 seconds before the Overview page loads in.  From watching the console, i don't see any actions happen other than a @@router/LOCATION_CHANGE to (*/home") right before the page loads. So guessing it's waiting on something before dispatching (goToHome) action

matheusd:
Just double checking: this is during initial app startup of during wallet startup (after selecting the wallet)?

I see a flash of a blank page on wallet startup, but it's pretty fast (definitely **not** in 5-10 seconds range). I also see a slightly longer flash when the wallet has tickets and will go to the "Process Managed Tickets" page (before the overview).

I did these tests on a `yarn build && yarn start` run.

alexlyp:
Yah I believe I'm referring to that "slightly longer flash" but I see the same flash with a wallet with no tickets.  I believe it's delaying with something in the get started machine

alexlyp:
I see "nothing" happening then the location change action hit and then the overview shows up.

amassarwi:
hrm it calls the wallet to check for coinjoin txs, i think it might be related

PR #3449, opened 2021-05-04
[startup] Add loader to Setup Wallet state

bgptr:
On wallet startup during Setup Wallet state, some tasks need time to finish: #3433
This diff doesn't try to optimize the background tasks but shows a loading bar instead of the blank page.


alexlyp:
Overall looks great. 

Was thinking if it's possible to also hide the Time estimation and Log lines once it goes into the Setup Wallet step.

bgptr:
@alexlyp Sure. Good catch. Updated.

alexlyp (approved):

PR #3453, opened 2021-05-06
[tickets] Track vpsd tickets and prevent account locking

matheusd:
This adds a call to fetch data about vspd tickets being tracked by the wallet and uses that data to prevent accounts being locked while the wallet needs them to perform vpsd-related work.

This gets rid of errors coming from the wallet about locked accounts processing managed tickets.

It also adds an internal service that runs while a wallet is opened and monitors when the accounts no longer need to be kept unlock and triggers a lock.

Exceptions are added for mixed, unmixed and ticket buying accounts while those services are running.

There are still some debug lines left to ease testing.

### DIFF
--- a/app/actions/ClientActions.js
+++ b/app/actions/ClientActions.js
@@ -7,7 +7,8 @@ import isUndefined from "lodash/fp/isUndefined";
 import {
   getNextAddressAttempt,
   getPeerInfo,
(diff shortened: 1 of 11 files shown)


alexlyp (approved):
works great tACK!

need to remove logging,

PR #3454, opened 2021-05-06
[locks]  Fix incorrect passphrase allowing to continue

alexlyp:
Incorrect passphrase during account security migration allowed users to proceed without completing the migration.


amassarwi (approved):
LGTM

PR #3457, opened 2021-05-07
[tickets] Only ProcessManaged at startup if needed

matheusd:
This adds a new config flag that can be used (when toggled to false) to skip the "Process Managed Tickets" view.

The `needs_vspd_process_managed` flag starts out as true by default and is also set as true whenever a ticket is purchased or the autobuyer is turned on.

During ticket monitoring, if no tickets are found to be tracked by the VSP and the autobuyer is off and the PMT view wasn't previously skipped, then the flag is set to false, to signal that processing is no longer needed.

The skip check is there to avoid skipping once then never seeing the PMT screen again.


alexlyp (approved):
tACK, just needs console msgs removed and it'll be all set

PR #3461, opened 2021-05-11
[dex] Add dcrdex asset dep and add script

alexlyp:
Also remove old instructions from README


matheusd (approved):
Correctly installing site after a `yarn` call

PR #3469, opened 2021-05-17
[dex] Fix default windows bitoin directory

alexlyp:
Roaming is the used directory for bitcoin instead of Local.


bgptr (approved):
Tested and works fine.

vctt94 (approved):
tACK

PR #3476, opened 2021-05-21
[tickets] Unlock unmixed account for autobuyer

alexlyp:
Previously, only the selected account was unlocked for the autobuyer.  For privacy autobuyers, we need to unlock the unlocked/change account too.


vctt94 (approved):
Ticket buyer worked fine on mixed and non mixed wallet. 

I've seen a problem with balance to mantain, on non privacy wallet, but I believe we can have a different PR for this.

PR #3485, opened 2021-05-25
[multi] Move code to wallet package

matheusd:
~**Requires #3397**~

This moves code that requires node and electron APIs from various places in the UI code stack to the `wallet` package. 

This makes it possible to (eventually) lift the entire wallet package to be offered via the preload script, such that node integration can be disabled for the main UI code.

The commits were done so that they can be individually reviewed, to simplify verifying all changes are mostly just relocating code.

Significant changes:

- Switch many places to use the `invoke/handle` paradigm of electron IPC
- Move all invoke and ipcRenderer functions to the wallet pkg
- Standardize all uses of the `wallet` pkg to use the `wallet` namespace and to import the wallet pkg in the same way (`import * as wallet from wallet`)
- Move some constants from `main_dev` files to `constants` so that UI code doesn't try to load `main_dev` files


alexlyp (approved):
tACK, tested most functionality on a testnet wallet and everything seemed to work as expected.

PR #3486, opened 2021-05-25
[multi] Disable Node Integration in the UI code

matheusd:
**Requires #3485**

This switches out all references to the `wallet` package in UI code to use a wallet API exposed by the preload scripts. This finally allows disabling node integration for the UI code.

This builds upon work done by previous PRs that moved all node and electron related APIs from being sprinkled around UI code to reside entirely in the wallet package. 

Disabling node integration for the UI code is an important security feature and a recommended practice in modern electron app development. However, since historically we never made an effort to clearly separate node-api-using code from generic UI code, we had a lot of interconnected dependencies that needed to be untangled before being able to finally take this step.

Therefore, while the changes in this PR are somewhat small and most of the work to get to this point was done by previous PRs, reviewers are encouraged to test out the major functions of the app, since switching off node integration completely removes those functions from UI code.

Major changes in this PR:

- Switch from `axios` to the standard `fetch` call for REST APIs (needed to be able to switch different implementations depending on the execution context of the code - whether main, renderer or preload process)
- Tweak webpack config based on the different execution contexts
- Expose various APIs (wallet, fs, blake, etc) via the preload script shim
- Switch all UI code to use the wallet via the preload script shim
- Finally disable node integration for the UI code


matheusd:
One important downside to note is that due to wallet code moving to the preload layer, changes during development to that code take a bit longer to recompile and you're forced to fully reload the page so that the preload script is reloaded.

bgptr (approved):
LGTM

alexlyp (approved):

PR #3496, opened 2021-06-07
multi: Enable toggling passphrase on device.

### LINKED ISSUE #3355: [trezor] Ask for passphrase in device for model T
For model T devices, we can ask for the passphrase in the device itself instead of asking it in the host (i.e. decrediton), which is more secure.

JoeGruffins:
Add a new button to trezor config options that toggles on and off
     password on device.

closes #3355


matheusd (approved):
Working fine on emulator.

PR #3500, opened 2021-06-08
[electron] Enable WebSecurity

matheusd:
~**Rebased on top of #3492**~

This re-disables the remote module on windows and enables the `webSecurity` feature.

`webSecurity` is used to tighten up external requests by enabling same-origin policy and disallowing mixed content. 

In order to enable this in development mode, external requests are modified in to return a `Allow-Origin: http://localhost:3000` header so that the main wallet page loaded via the HMR server is allowed to perform them. Additionally, we fix the `OPTIONS` request performed as a prefetch step to politeia servers when performing a `POST` request to always return a 200 response.

PR #3503, opened 2021-06-09
[webpack] Refactor configs

matheusd:
~**Rebased on top of #3500**~

This refactors the webpack configs to improve their organization and remove the need for babel when running the various webpack scripts.

Summary of changes:

- Move existing webpack config files to the `webpack/` dir so they don't crowd the root project dir anymore
- Normalize all config files to use standard node.js syntax (no es6 style imports)
- Drop babel from being loaded when running webpack scripts
- Switch `server.js` to `server.mjs` and drop babel from it
- Add a dev mode webpack config for the preload script
- Cleanup related scripts in `package.json`


alexlyp (approved):
tACK
yarn dev, build/start and package all working on linux

PR #3505, opened 2021-06-10
Disabled Dex menu with tooltip in SPV mode

bgptr:
This diff implements @jholdstock's idea on the matrix: "Maybe decrediton in SPV mode should just gray out the Dex tab, with a tooltip to explain that SPV needs to be switched off. Hiding it completely has caught a lot of people out)"


alexlyp:
Overall looks pretty good! 

Couple things that stand out: While in Dark tonality, it's hard to read DEX in the sidebar.  Light looks great.

I think it reads better as "DEX not available while using SPV.  Please go to settings and disable SPV to access the DEX."

bgptr:
@alexlyp Thanks for the review. I pushed the fixes.

alexlyp (approved):
looks great!

PR #3506, opened 2021-06-11
[tickets] Revoke Ticket in SPV

alexlyp:
~Requires https://github.com/decred/dcrwallet/pull/2061~

This adds the ability to revoke tickets while in SPV mode.

The Revoke button is currently shown to all non-pending tickets while in SPV mode.  There is a disclaimer shown when confirming the private passphrase.


matheusd (commented):
Created the revocation tx correctly, but I noticed a couple of wrong UX behaviors (that seem to be in master as well, so I'm fine if we just create issues and fix them in other PRs).

~An unmined transaction shows up as confirmed with a very large confirmation count for me, when looking at the tx details.~ (duh, just saw the issue you've already opened)

Also, the revocation tx does not show the "abandon tx" button.

alexlyp:
Yah @matheusd looks like some issues on master in #3507.  I believe the not showing the abandon button is the same situation.

matheusd (approved):
Needs a rebase and to wait for the dcrwallet pr to be merged, but worked fine both for an actual expired ticket and one that hadn`t expired yet.

alexlyp:
Ok has been rebased and ready for final testing against master dcrwallet @matheusd

matheusd:
I'm getting an error now "ticket hash is required to revoke ticket"

matheusd (approved):
Working as intended now.

PR #3511, opened 2021-06-16
PreCreateWallet: Replace more name chars.

### LINKED ISSUE #3359: [bug] specific special characters as wallet names cause decrediton errors
Naming a wallet 

`#something` or `$something` seems to cause decrediton to bug out and refuse to load any wallets. 

Also odd that it creates two wallets when `#` is used.
```
\#test
#test
```

image

Need to review the usage of special chars in wallet names.

JoeGruffins:
Disallow some problematic characters in wallet names.

closes #3359

This works in the same way it did before, just adds a few more characters to the replace and doesn't allow for leading spaces. space, +, -, _, and : look to be fine on linux, but Windows may be more restrictive. Will test there if noone else can. Parenthesis and brackets may also be fine and used by some people. Allow?

### DIFF
--- a/app/components/views/GetStartedPage/PreCreateWallet/PreCreateWallet.jsx
+++ b/app/components/views/GetStartedPage/PreCreateWallet/PreCreateWallet.jsx
@@ -48,8 +48,12 @@ const PreCreateWallet = ({
     (newWalletName) => {
       setHasFailedAttemptName(true);
       let nameAvailable = true;
-      // replace all special path symbols
-      newWalletName = newWalletName.replace(/[/\\.;:~]/g, "");
+      const replaceNameChars = /[`!@#$%^&*()=[\]{};'"\\|,.<>/?~]/;
+      // Replace all special path symbols except for space, _, -, :, and +.
+      newWalletName = newWalletName.replace(replaceNameChars, "");
+      // Remove leading spaces.
+      if (newWalletName.length > 0 && newWalletName[0] === " ")
+        newWalletName = newWalletName.slice(1);
       for (let i = 0; i < availableWallets.length; i++) {
         if (newWalletName == availableWallets[i].value.wallet) {
           nameAvailable = false;
--- a/app/components/views/GetStartedPage/PreCreateWallet/PreCreateWallet.jsx
+++ b/app/components/views/GetStartedPage/PreCreateWallet/PreCreateWallet.jsx
@@ -48,8 +48,16 @@ const PreCreateWallet = ({
     (newWalletName) => {
       setHasFailedAttemptName(true);
       let nameAvailable = true;
-      const replaceNameChars = /[`!@#$%^&*()=[\]{};'"\\|,.<>/?~]/;
-      // Replace all special path symbols except for space, _, -, :, and +.
+      // Users should be able to (at least in principle) use any wallet name
+      // they so choose. However, certain special chars need to be explicitly
+      // filtered due to us directly using the wallet name as part of the path
+      // to the wallet's files. The list of chars that need to be filtered out
+      // are:
+      //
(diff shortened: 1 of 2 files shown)


JoeGruffins:
It might also be better to change to only allow a certain set, which may be more complete. I think some windows systems in different languages have other special characters.

JoeGruffins:
@degeri Any opinion on what to allow or disallow?

degeri:
Cant we go for a whitelist ? 

a-z,A-Z,0-9

matheusd:
> Cant we go for a whitelist ?
>
> a-z,A-Z,0-9

This would be too restrictive for non-US speakers (e.g. can't name a wallet ¨Poupança").

degeri:
> > Cant we go for a whitelist ?
> > a-z,A-Z,0-9
> 
> This would be too restrictive for non-US speakers (e.g. can't name a wallet ¨Poupança").

Yes it would be. The problem is that if we extend to many languages we will hit alot of problems like R->L chars, Long chars ด้้้้้็็็็็้้้้้็็็็็้้้้้้้้็็็็็้้้้้็็็็็้้้้้้้้็็็็็้้้้้็็็็็้้้้้้้้็็็็็้้้้้็็็็ etc. Not sure how we handle that currently tbh. 

Maybe we should test with  https://github.com/minimaxir/big-list-of-naughty-strings/blob/master/blns.json  and see which cause errors/inconsistencies .

degeri:
After much discussion @matheusd and me came to the conclusion that denying any unicode char (Unless it specifically causes and issue) will be bad UX for users. 

Even the bug due to `#` was due to a dependency escaping the char. https://github.com/npm/ini/blob/master/ini.js#L159

matheusd:
To expand a bit on degeri's comment:

Users should be able to (at least in principle) use any wallet name they so choose. However, certain special chars need to be explicitly filtered due to us directly using the wallet name as part of the path to the wallet's files. The list of chars that I believe need to be filtered out are:

- Filesystem related chars: `/\.:`
- Escaped when stored in `dcrwallet.conf` [ini files](https://github.com/npm/ini/blob/master/ini.js#L159): `;#[]`
- [Specially handled](https://github.com/decred/dcrwallet/blob/master/config.go#L194-L197) by dcrwallet: `$%~`

JoeGruffins:
Blacklisting suggested chars. Thank you.

app/components/views/GetStartedPage/PreCreateWallet/PreCreateWallet.jsx
```diff
@@ -48,8 +48,20 @@ const PreCreateWallet = ({
...
+      // are:
+      //
+      // Filesystem related chars: /\.:
+      // Escaped when stored in dcrwallet.conf ini files: ;#[]
+      // Specially handled by dcrwallet: $%~
+      const replaceNameChars = /[/\\.:;#[\]$%~]/;
+      newWalletName = newWalletName.replace(replaceNameChars, "");
+      // Remove leading spaces.
+      if (newWalletName.length > 0 && newWalletName[0] === " ")
+        newWalletName = newWalletName.slice(1);
```
matheusd, line 64: You could have multiple spaces, so best to just use `trim()`

JoeGruffins: How do you mean. It runs this code every key press, so trim disallows any spaces at all. I could put trim in somewhere else I suppose?

JoeGruffins: Should we also disallow spaces? I just thought a leading space was silly and confusing, but something like "My Wallet" should be fine.

matheusd (approved):

PR #2869, opened 2020-11-06
multi: Allow trezor ticket purchasing on testnet.

JoeGruffins:
Fix updating firmware with model T by sending binary rather than hex.

In signTransactionAttemptTrezor give spending from stake scripts
inputs specific script types "SPENDSSRTX" and "SPENDSSGEN" rather than
"SPENDADDRESS" as trezor needs this information to process the scripts.

Add purchaseTicketsV3 which purchases tickets using a watching only
trezor wallet from a vsp with api v3. Errors if not on testnet.

Add separate payVSPFee function that will pay a tickets fee if not
already paid, and throws if already paid.

Correct purchase ticket button for watching only wallets. It no longer
asks for a password. Connect purchasing through trezor when isTrezor is
true.

Add vsp v3 endpoints "feeaddress" and "payfee" to allowed external
requests.

Change wallet/control.js to not require an unlocked wallet when signTx
is false.

PR #3512, opened 2021-06-16
trezorActions: Add votingXprivSeed.

JoeGruffins:
Add a function that can query trezor for deterministic bytes to be used
    as the seed for a voting xpriv.

depends on #2869

### DIFF
--- a/app/actions/TrezorActions.js
+++ b/app/actions/TrezorActions.js
@@ -1,6 +1,12 @@
 import { wallet, fs } from "wallet-preload-shim";
 import * as selectors from "selectors";
-import { hexToBytes, str2utf8hex, rawToHex } from "helpers";
+import {
+  hexToBytes,
+  str2utf8hex,
+  rawToHex,
+  rawHashToHex,
+  putUint16
+} from "helpers";
 import {
   walletTxToBtcjsTx,
   walletTxToRefTx,
@@ -24,6 +30,7 @@ import {
   SIGNMESSAGE_SUCCESS
 } from "./ControlActions";
 import { getAmountFromTxInputs, getTxFromInputs } from "./TransactionActions";
+import { blake256 } from "walletCrypto";
 
 const session = require("trezor-connect").default;
 const {
@@ -40,6 +47,15 @@ const NOBACKUP = "no-backup";
 const TRANSPORT_ERROR = "transport-error";
 const TRANSPORT_START = "transport-start";
 const BOOTLOADER_MODE = "bootloader";
+const testVotingKey = "PtWTXsGfk2YeqcmrRty77EsynNBtxWLLbsVEeTS8bKAGFoYF3qTNq";
+const testVotingAddr = "TsmfmUitQApgnNxQypdGd2x36djCCpDpERU";
+const SERTYPE_NOWITNESS = 1;
+const OP_SSGEN_STR = "bb";
+const OP_SSRTX_STR = "bc";
+const OP_TGEN_STR = "c3";
+const STAKE_REVOCATION = "SSRTX";
+const STAKE_GENERATION = "SSGen";
+const TREASURY_GENERATION = "TGen";
 
 let setListeners = false;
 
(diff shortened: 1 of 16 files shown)


app/actions/TrezorActions.js
```diff
@@ -49,6 +49,11 @@ const OP_SSGEN_STR = "bb";
 const OP_SSRTX_STR = "bc";
 const STAKE_REVOCATION = "SSRTX";
 const STAKE_GENERATION = "SSGen";
+// VOTING_XPRIV_KEY and VOTING_XPRIV_VALUE must never change in order to
+// have deterministic voting keys.
+const VOTING_XPRIV_KEY = "Create decred voting xpriv seed bytes?";
```
JoeGruffins, line 54: We may be able to get around this by using an IV.

app/actions/TrezorActions.js
```diff
@@ -993,6 +998,46 @@ export const getWalletCreationMasterPubKey = () => async (
...
+        encrypt: true,
+        askOnEncrypt: true,
+        askOnDecrypt: true
+      });
+      return res.payload;
+    });
+    const seed = Buffer.from(ckv.value, "hex");
+    if (seed.length != 32) throw "unusable voting HD pivate key seed bytes";
+    dispatch({ type: TRZ_VOTINGXPRIVSEED_SUCCESS });
+    return seed;
```
JoeGruffins, line 1035: I think the best thing to do with the seed is to hand it over to dcrwallet in a new grpc request and have it mathed into the xpriv there, using tested logic.

JoeGruffins (commented):
The logic behind the `cipherKeyValue` endpoint is [here](https://github.com/satoshilabs/slips/blob/master/slip-0011.md). Very interested in @matheusd opinion, of course.

app/actions/TrezorActions.js
matheusd: An IV has nothing to do with the immutability of the key, it's only used as part of the input to the encryption algorithm.

app/actions/TrezorActions.js
matheusd: Correct.

app/actions/TrezorActions.js
matheusd: An IV has nothing to do with the immutability of the key, it's only used as part of the input to the encryption algorithm.

app/actions/TrezorActions.js
```diff
@@ -993,6 +998,46 @@ export const getWalletCreationMasterPubKey = () => async (
...
+    });
+    return;
+  }
+
+  const chainParams = selectors.chainParams(getState());
+
+  try {
+    const ckv = await deviceRun(dispatch, getState, async () => {
+      const res = await session.cipherKeyValue({
+        path: accountPath(WALLET_ACCOUNT, chainParams.HDCoinType),
```
matheusd, line 1023: Probably best to use `addressPath` instead of simply `accountPath`

matheusd (commented):
It's important to note in the PR/commit that this method was chosen specifically because of instructions by the Trezor team in https://github.com/trezor/trezor-firmware/pull/1249#issuecomment-775208665.

I also think it's important to create a go CLI tool that can recreate this xpriv in software only, so we can be sure that in worst case scenarios (e.g. lost trezor without access to a replacement) a user can recreate its voting xpriv.

My [BIP0039 to DCR Seed](https://github.com/matheusd/bip39-to-dcr-seed) can go from a trezor (+ optional passphrase) to raw dcr seed, so the tool to recreate the xpriv could go from the raw seed to xpriv (or whatever). The important bit is that this must be verified to be working to be offered as a last resort to users.

app/actions/TrezorActions.js
JoeGruffins: Fine with me, but why is that?

app/actions/TrezorActions.js
JoeGruffins: Will do.

app/actions/TrezorActions.js
JoeGruffins: I see. The key must not change then.

JoeGruffins:
> I also think it's important to create a go CLI tool that can recreate this xpriv in software only, so we can be sure that in worst case scenarios (e.g. lost trezor without access to a replacement) a user can recreate its voting xpriv.

So, after retrieving, should ask the user to save it, similar to the wallet's seed when creating a wallet?

app/actions/TrezorActions.js
matheusd: We can extract private keys of individual addresses using standard CLI tools if needed.

matheusd:
> > I also think it's important to create a go CLI tool that can recreate this xpriv in software only, so we can be sure that in worst case scenarios (e.g. lost trezor without access to a replacement) a user can recreate its voting xpriv.
> 
> So, after retrieving, should ask the user to save it, similar to the wallet's seed when creating a wallet?

I'm not sure I understood the question, but the idea is that you can pass in the raw dcr hex seed and get back the voting xpriv, so it's a specific process  only used in "disaster recovery" scenarios. Something like:

```
$ bip39-to-dcr-seed
[type seed]
af291933...

$ dcr-voting-xpriv
af231933....
Pm29a00...
```

After that, you can manually import the xpriv in dcrwallet.

app/actions/TrezorActions.js
JoeGruffins: By cli tools do you mean [trezor-cli](https://wiki.trezor.io/Using_trezorctl_commands_with_Trezor)? We cannot extract any private keys afaik.

JoeGruffins:
< I'm not sure I understood the question

I suppose I was thinking we couldn't get the seed again, but it's not like the wallet where you only have the one chance. We can get the seed any time. So nevermind there.

app/actions/TrezorActions.js
matheusd: Standard decred tools if you have the seed imported in a dcrwallet instance

JoeGruffins: oh, that's right, the trezor seed to dcrwallet, I understand now.

alexlyp:
Any status on this?  It's been in draft since June 2021.

JoeGruffins:
We decided to get dcrwallet changes done first, so waiting for review over there. https://github.com/decred/dcrwallet/pull/2120 and then https://github.com/decred/dcrwallet/pull/1946 and there will need to tie up up some loose ends. So... I can take these down and put them back up later if they are an eyesore.

alexlyp:
Gotcha, thanks for the explanation.  I'll chat with @jrick tomorrow about what he thinks of the dcrwallet portions and then we can go from there.  No need to close these since we're waiting elsewhere for movement.  Thanks!

alexlyp:
Same as from my comment on the Trezor purchase ticket PR.

PR #3509, opened 2021-06-14
[multi] Reduce dependencies in main and preload layers

matheusd:
This reduces the dependencies to third party packages in the main electron and preload script layers of the application. 

This improves the codebase by relying less on third party code, slightly reduces compilation time and improves security by reducing the attack surface of a possible supply chain attack.

The main changes introduced by this PR are:
  - Add a script that can generate graphs for the dependency tree of each layer of the application
    - This was added as the first commit so that dependencies before and after the refactor can be checked by doing a `node ./scripts/generateDepGraphs.js`
  - Remove the now unneeded (due to updated electron version) intl-relativetimeformat polyfill
  - Reduce (but not yet eliminate) dependency on lodash by performing specific imports and functions
  - Remove the need to import the `wallet` by moving the single outstanding dependency directly into the `main_dev` package
  - Remove the dependency to the `websocket' (`ws`) library by using the `jsonrpc` api on dcrd in post mode (instead of websocket mode)
- Switch from using the `fs-extra` package to the node standard `fs` (we don't use any of the additional features offered by `fs-extra`)
- Run `yarn-deduplicate`

The script added to generate graphs of the app's dependencies generates a link to [npmgraphs.js.org](https://npmgraphs.js.org) which can be used to compare transitive dependencies as specified in the npm registry:

### Before this PR

- [Main electron layer](https://npmgraph.js.org/?q=%40formatjs%2C%40grpc%2Cbs58%40%5E4.0.1%2Cbs58check%40%5E2.1.1%2Cdom-helpers%40%5E3.4.0%2Celectron-store%40%5E7.0.2%2Cfs-extra%40%5E9.1.0%2Cgoogle-protobuf%40%5E3.15.6%2Cini%40%5E2.0.0%2Cint64-buffer%40%5E1.0.0%2Cis-running%40%5E2.1.0%2Clodash%40%5E4.17.14%2Clogform%2Cminimist%40%5E1.2.3%2Cqr-image%40%5E3.2.0%2Creact-dom%40%5E16.13.0%2Creact%40%5E16.13.0%2Creselect%40%5E4.0.0%2Cstring-argv%40%5E0.3.1%2Cwinston%40%5E3.3.3%2Cws%40%5E7.4.4) - 104 Modules
[...]

PR #3515, opened 2021-06-18
[Multi] Confirm access to VSPs via BrowserView dialog

matheusd:
~**Requires #3509**~

This adds a confirmation dialog entirely controlled by the preload script and the main electron process through the use of a `BrowserView` instance.

The main benefit for this approach is having a confirmation screen that lives outside the main UI and cannot be modified by code running in the UI layer.

The `BrowserView` class is an electron feature that allows overlapping a new rendering frame on top of the existing window, but without adding it to the corresponding document's DOM. All the code loaded in a BrowserView is effectively in a separate window and thus inaccessible to anything except the main electron process.

After splitting the code across the various layers of the app, turning on electron's security features (`nodeIntegration` and `contextIsolation`) and reducing the dependencies of the main and preload layers to a minimum, we can finally introduce this method for confirming external accesses that reduces the ablility of code injected into the main UI layer (for example, via third party dependencies) to exfiltrate data from the wallet.

The confirmation dialog introduced by this PR is intended to be shown whenever a new VSP (using an arbitrary domain) needs to be accessed via the network, thus allowing users to be selective about which services they want to potentially leak metadata to (e.g. times when the user is online and with their wallet open) and works as a deterrence against malicious code attempting to contact a remote server (since accesses will now need explicit permission).

The `BrowserView` based confirmation dialog works as follows:

- The UI code determines that it needs to access a VSP and calls the `allowVSPHost()` function exported by the preload layer. 
- The preload layer builds the html of the confirmation dialog.
- The preload layer sends a message to the main electron layer, which opens the `BrowserView` with the specified contents
- The `BrowserView` window is closed when the user either accepts or rejects the given request
- The reply is sent back to the originating `allowVSPHost()` call

Note that the `confirmation-dialog.html` is using very simple markup and inline scripts and _no external depencies_, including react. This is intentional, given we want this window to not have a chance to include code from third party packages.

Main changes introduced by this PR:

- Fix some VSP-related calls that were in a batched fashion but not correctly waited for
- Add a config entry to track VSP hosts allowed to be contacted and prevent UI code from directly modifying this entry
- Add the new confirmation dialog
- Switch the `allowVSPHost()` call to ask for confirmation via the new confirmation modal

Note to testers: for the moment, to ease testing this PR, confirmation for all VSPs returned by the official VSP listing public API is ask for on every wallet load. This allows viewing the confirmation every time you run the PR. This will be reverted once this PR is closer to being merged.

### DIFF
--- a/app/actions/VSPActions.js
+++ b/app/actions/VSPActions.js
@@ -662,7 +662,7 @@ export const getVSPsPubkeys = () => async (dispatch) => {
       } catch (error) {
         // Skip to the next vsp.
       }
-    };
+    }
     dispatch({ type: GETVSPSPUBKEYS_SUCCESS, availableVSPsPubkeys });
   } catch (error) {
     dispatch({ type: GETVSPSPUBKEYS_FAILED, error });
@@ -816,7 +816,7 @@ export const setVSPDVoteChoices = (passphrase) => async (
       // TODO: this should be tracked so that users know this VSP isn't voting
       // according to their wishes.
     }
-  };
+  }
   try {
     dispatch({ type: SETVSPDVOTECHOICE_ATTEMPT });
     const walletService = sel.walletService(getState());
--- a/app/actions/VSPActions.js
+++ b/app/actions/VSPActions.js
@@ -22,30 +22,31 @@ export const GETVSP_ATTEMPT = "GETVSP_ATTEMPT";
 export const GETVSP_FAILED = "GETVSP_FAILED";
 export const GETVSP_SUCCESS = "GETVSP_SUCCESS";
 
-export const getVSPInfo = (host) => (dispatch) =>
-  new Promise((resolve, reject) => {
-    dispatch({ type: GETVSP_ATTEMPT });
-    // 5 seconds for timeout.
-    const TIMEOUT_TIME = 5000;
-    // add https into vsp host, so we can fetch its information.
+export const getVSPInfo = (host) => async (dispatch) => {
+  // 5 seconds for timeout.
+  const TIMEOUT_TIME = 5000;
+
+  dispatch({ type: GETVSP_ATTEMPT });
+
+  try {
+    // Check if user allows access to this VSP. This might trigger a confirmation
+    // dialog.
     host = `https://${host}`;
-    wallet.allowVSPHost(host);
+    await wallet.allowVSPHost(host);
+
     const timeout = new Promise((resolve, reject) =>
       setTimeout(() => reject({ isTimeout: true, vspHost: host }), TIMEOUT_TIME)
     );
     const getInfo = wallet.getVSPInfo(host);
 
-    Promise.race([timeout, getInfo])
-      .then((info) => {
-        clearTimeout(timeout);
-        dispatch({ type: GETVSP_SUCCESS, info });
-        resolve(info.data);
-      })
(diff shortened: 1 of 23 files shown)


app/actions/VSPActions.js
```diff
@@ -802,14 +807,16 @@ export const setVSPDVoteChoices = (passphrase) => async (
...
-    if (pubkey) {
+  for (const vsp of availableVSPs) {
+    try {
+      const { pubkey } = await dispatch(getVSPInfo(vsp.host));
       vsp.pubkey = pubkey;
-      return vsp;
+    } catch (error) {
+      // Ignore fetching info (won't set the vote choice for this vspd).
+      // TODO: this should be tracked so that users know this VSP isn't voting
+      // according to their wishes.
```
matheusd, line 817: Surfacing TODO. this should become an issue to be dealt with.

app/wallet/config.js
```diff
@@ -10,11 +11,24 @@ export const getWalletCfg = (...args) => {
   };
 };
 
+// This is a map that stores which keys cannot be directly set by UI code and
+// must be set only by preload script functions.
+//
+// TODO: switch to an allowlist mode instead, after identifying the relevant
+// entries and move setting code to the preload layer.
```
matheusd, line 18: Surfacing TODO. This should be an issue to be dealt with.

app/wallet/vsp.js
```diff
@@ -50,9 +53,44 @@ export const getVSPTicketStatus = promisifyReqLogNoData(
...
+
+    // Allow access to all VSPs returned by the official VSP listing endpoint.
+    // This is less then ideal because this endpoint might eventually return
+    // domains that switched owners or that are otherwise available for hijack,
+    // but is needed due to Decrediton (currently) having to iterate over all
+    // existing VSPs to discover their pubkey and sync tickets.
+    //
+    // Eventually this can be further locked down once that iteration isn't
+    // performed any more and VSPs are only accessed when attempting to purchase
+    // a ticket.
```
matheusd: Surfacing issue. Ideally the public VSP listing API should be providing all the basic information we need such that we _don't_ need to be reaching out to every VSP on every wallet start to figure out this info.

Need to double check whether the VSP public key is the only thing missing or if there's anything else.

app/wallet/vsp.js
```diff
@@ -93,8 +94,16 @@ export const getAllVSPs = withLogNoData(
   "getAllVspsInfo"
 );
 
-// allowVSPHost enables the external request to a specif VSP host.
+// allowVSPHost enables the external request to a specific VSP host.
 export const allowVSPHost = log(
-  (host) => Promise.resolve(ipcRenderer.sendSync("allow-vsp-host", host)),
-  "Allow StakePool Host"
+  async (host) => {
+    // TODO: only ask for confirmation if VSP host is not yet allowed.
```
matheusd: This needs to be fixed before merging (this is only being left as-is to ease testing).

matheusd:
Marking as ready to review, though the last TODO needs to be fixed before this is merged.

bgptr (approved):
Nice work! Tested on dev and production mode on linux and looks great.
Left only a minor suggestion

alexlyp (approved):

PR #3519, opened 2021-06-25
[wallet] Use BrowserView ConfirmDialog on SignTx

matheusd:
~**Requires #3515**~

This switches the wallet's `SignTx` function to confirm with the user the transaction to be signed via the recently introduced confirmation dialog that runs in a BrowserView.

This ensures that any attempts by UI code to modify the transaction are visible to the user before the tx is sent to the underlying wallet to be signed.

The dialog is formatted such that only outputs and amounts that are not returning to the wallet (i.e. addresses in outputs that do _not_ correspond to one of the wallet's BIP0044 account) are shown.


bgptr (approved):
LGTM
Tested in dev and production mode on linux.

alexlyp (approved):
tACK

PR #3523, opened 2021-06-29
[main] Add menu toggle devtools in conf window

matheusd:
This makes it so that the devtools for the confirmation window doesn't show up automatically, but still allows opening it from the main menu for debugging.


bgptr (approved):
Tested and works nicely

alexlyp (approved):

PR #3532, opened 2021-07-12
[settings] Add DiscoverUsage in place of existing gap limit in settings

alexlyp:
Requires https://github.com/decred/dcrwallet/pull/2071

This removes the existing gap limit in settings and replaces it with the discover address usage.

This action in wallet properly discovers all addresses in wallet based on the gap limit provided.  Once the discover is complete a rescan is also issued to make sure the balance is properly reflected upon completion.


bgptr:
Tested and works fine.
I noticed only one issue: the gap limit was not saved to the wallet config, but maybe it's intentional.

app/actions/ControlActions.js
```diff
@@ -485,6 +485,28 @@ export const revokeTicketsAttempt = (passphrase) => async (
...
+  const walletService = sel.walletService(getState());
+  try {
+    const discoverUsageResponse = await wallet.discoverUsage(
+      walletService,
+      gapLimit
+    );
+    dispatch({ discoverUsageResponse, type: DISCOVERUSAGE_SUCCESS });
+    dispatch(rescanAttempt(0));
+  } catch (error) {
+    dispatch({ error, type: DISCOVERUSAGE_FAILED });
```
matheusd, line 506: `DISCOVERUSAGE_FAILED` needs to be added to the snackbar, otherwise users don't get any feedback on errors.

alexlyp:
That's correct @bgptr.  Currently the only way to change your gap limit is to change it on wallet creation (and then it will be used every load after that) or manually changing it in the config file (not recommended obvi). 

This new address discover allows users to scan for address usage that could be outside of their current gap limit but will remain gap limit they have set will remain.

matheusd (commented):
The lack of any feedback makes this hard for users to track: there's no indication the process is going on after the modal is closed and before rescan starts.

You should set a redux state flag that remains active between the `DISCOVERUSAGE_ATTEMPT`..`DISCOVERUSAGE_SUCCESS` actions, to block successive executions of the process.

JoeGruffins (approved):

app/actions/ControlActions.js
```diff
@@ -485,6 +485,28 @@ export const revokeTicketsAttempt = (passphrase) => async (
...
+  getState
+) => {
+  dispatch({ type: DISCOVERUSAGE_ATTEMPT });
+  const walletService = sel.walletService(getState());
+  try {
+    const discoverUsageResponse = await wallet.discoverUsage(
+      walletService,
+      gapLimit
+    );
+    dispatch({ discoverUsageResponse, type: DISCOVERUSAGE_SUCCESS });
```
matheusd, line 503: After this line, add the following (before rescanning) to force reloading the address filters in SPV mode:

`await wallet.loadActiveDataFilters(walletService);`

This should fix the last issue with this PR.

alexlyp:
OK @matheusd ok this is all rebased and should be ready for final review.

PR #3535, opened 2021-07-20
[governance] feat: add pagination for proposals inventory

### LINKED ISSUE #3517: Politeia: Add pagination for inventory call.

victorgcramos:
This diff closes #3517, by adding pagination for votes inventory on governance page. Also, this diff removes `Active Voting` tab and moves it into the `Under Review` tab, in order to standardize the UX (see: https://github.com/decred/politeiagui/pull/2431).

Since we cannot rely on the full inventory to check for proposals updates, I decided to remove the `compareInventory` action, because inventories cannot be compared anymore, because once an inventory is full, it will move some tokens to the next page, which may cause lots of edge-cases. In addition, the `Refresh Inventory` action now fetches the whole proposals list again, in order to get the most recent updates for each proposals (previously we only updated the proposals with tokens that differed from the current inventory). This now provides a more consistent data among all proposals.


app/middleware/politeiaapi.js
```diff
@@ -3,12 +3,12 @@ import { postJSON } from "helpers/fetch";
 // Uncomment this and comment the following definition to test locally.
 // Also uncomment a code part at externalRequests.js, as politeia is a ssl-enabled
 // service and decrediton will fail to fetch locally with a self signed cert.
-// export const POLITEIA_URL_TESTNET = "https://localhost:4443";
+export const POLITEIA_URL_TESTNET = "https://localhost:4443";
```
matheusd, line 6: This needs to be commented back.

app/main_dev/externalRequests.js
```diff
@@ -49,17 +49,20 @@ export const installSessionHandlers = (mainLogger) => {
...
+      (event, webContents, url, error, certificate, callback) => {
+        if (url.match(/^https:\/\/localhost:4443\/.*$/)) {
+          event.preventDefault();
+          callback(true);
+        } else {
+          callback(false);
+        }
+      }
+    );
+  }
```
matheusd, line 65: These changes need to be reverted.

app/middleware/politeiaapi.js
```diff
@@ -3,12 +3,12 @@ import { postJSON } from "helpers/fetch";
...
 
 // Politeia doc source:
 // https://github.com/decred/politeia/blob/master/politeiawww/api/www/v1/api.md
 
-export const POLITEIA_URL_TESTNET = "https://test-proposals.decred.org/api";
+// export const POLITEIA_URL_TESTNET = "https://test-proposals.decred.org/api";
```
matheusd, line 11: This needs to be uncommented.

matheusd (commented):
This currently has an annoying behavior where, when loading additional items, the entire listing is set to a loading indicator and the scroll rolls back to the start of the listing (so it makes it hard to track what the new items are).

Is it possible to do something similar to the transactions history listing, where the indicator is displayed at the bottom of the listing and only new items are added?

matheusd:
Also, please drop all changes in files related to translaction (.po, .pot, .json). Those are only done as part of a release build.

alexlyp:
Still need the issues arose by @matheusd completed.  Closing, but would happily receive an updated version to review.

Issue #3296, opened 2021-03-07
LN UX improvements concepts – discussion / validation


linnutee:
We’ve worked out concept design drafts to validate. These are based on a small study done on a few more established BTC LN wallets, available articles on LN UX and brief task-based testing w/ 3-4 with beginner-level DCR users. See drafts here: https://www.figma.com/file/EanUbqJmPs8EMqY26j9zwo/decred-piui-master?node-id=556%3A0

The main areas we’ve touched point on are:
1.  Providing a better understanding and feedback of how balances and capacities function 
2. Simplified understanding and connecting to nodes / creating channels
2. Building consistency w/ patterns to normal on-chain transacting in Decrediton. 

Navigation:
- Overview
- Channels
- Send
- Receive
- Advanced/Misc     Ideally 

Ideally, Overview and Channels views could eventually be combined if the channel management gets mostly automated. However in the current state, without leaving much out, combining the views led to a too crowded view in comparison to rest of Decrediton individual views.  

**Overview**: has two sections, a statistics component and an activity log.
- The statistics card should provide a quick overview of both the user's wallet/account and the network in general. 
- Activity log should list all transactions (including on-chain ones that affect the LN account), changes regarding invoices, all changes regarding channels. Each of the status log row can be distinguished with relevant icons, and only contain key information and a way to follow-up if needed.  

**Channels**: some of the bigger problems for users is understanding how nodes and channels work, and what they need to do with them if not running on autopilot. Few takeaways on the UI to improve this through manual interactions: 
- Communicating and visualizing the mechanics, ie. nodes being on the top level and doing most of the processing. 
- Providing better accessibility to discover nodes, as well decide why to choose one. (Search feature and rooting use of custom node names in opposed to only ID's) (Recent / Suggested Nodes?) (Ranking statistics based on capacity, age, availability, location etc? – If this would have security issues, could perhaps be opt-in/out as for Pi for example)
- Simplifing connecting to nodes and creating channels.
- Channels capacity visualization. 
- Keeping users in control of what's going on, whether channels management is automated or what degree of manual management is wanted.
- Regarding the search and browsing would need the most dev input, what's possible and feasible to do.

**Send and Receive**: overall aiming to have as much similarity to onchain transacting as possible. Key takeaways: 
[...]

PR #3537, opened 2021-07-21
[redesign][ln] Receive tab

bgptr:
Toward issue #3296

Some notes:
- show settle date on invoice modal instead of expiration time when the payment already has been received
- cancel invoice feature added
- add snackbar messages for indicating add and cancel invoice events
- implemented filter, sort, and search function for invoice list
- tests added

Updated screenshots:
DeepinScreenshot_select-area_20210809215032
DeepinScreenshot_select-area_20210805140422
DeepinScreenshot_select-area_20210809214956


matheusd (commented):
This is going in the right track, but I see some UX issues:

- You can't create a 0 amount invoice anymore. This is needed for "donations" (a 0 amount invoice indicates the sender can send any amount)
- Both canceled and timed out invoices are appearing as "canceled" (not sure that's the intention)
- Maybe we should disable copying the payment request if the payment was canceled/timed out? It will fail if someone attempts to pay for it, so this could make it less prone to cause this error
- Amount in the individual invoice box (in the invoices listing) is showing a "+" sign
- I'd like to get an additional expandable view (either in the details modal or directly in the listing) with the extended info for the invoice (hash, description, htlcs, etc).

bgptr:
@matheusd Thanks for your review. I've pushed the suggested changes and updated the screenshots.

bgptr:
Update: implemented the polished details part of the invoice modal thanks to @MariaPleshkova. (screenshots are updated)
Also, converted the details table to a shared component since more of the ln pages will use it.

app/components/modals/LNInvoiceModal/helpers.js
```diff
@@ -0,0 +1,75 @@
+import { FormattedMessage as T } from "react-intl";
+export const getInvoiceDetails = (invoice, tsDate) => {
+  const details = [
+    {
+      label: <T id="ln.invoicesModal.hash" m="Hash" />,
+      value: invoice?.rHash
```
matheusd, line 6: This is being displayed as a base64 encoding. Usually this is shown as an hex.

matheusd (commented):
Also, it's probably useful to have the payment hash be copyable, since that's used in the command line to verify things.

MariaPleshkova:
@bgptr 
I added the copy button next to the hash as @matheusd suggested. We already use the same button for the Remote Pubkey in the Channel Properties.
image

bgptr:
@matheusd @MariaPleshkova Thanks, guys. I pushed the changes and updated the screenshots.

matheusd (approved):
Working nicely now!

PR #3539, opened 2021-07-30
[redesign][ln] ln wallet creation warnings view

bgptr:
Toward issue #3296

This diff implements the redesigned warning view before ln wallet creation.
~~I set this PR draft because the graphics haven't yet been added. 
@linnutee, @MariaPleshkova once they'll arrive, it will be easy to add them.~~

Screenshot:
Screenshot 2021-12-24 at 14 19 31
Screenshot 2021-12-24 at 14 19 43
Screenshot 2021-12-24 at 14 19 51
Screenshot 2021-12-24 at 14 19 57
Screenshot 2021-12-24 at 14 20 06
Screenshot 2021-12-24 at 14 20 12


alexlyp:
@matheusd when you have some extra cycles can you review this?  Looks really slick and would be great to have all set.

matheusd (approved):
Looks great as is! The only small nit I have is that by default this will require scrolling the screen due to the sizes of the images, but i guess we can iterate that after the design team has had another look.

alexlyp:
Agreed, looks like we'll need to have another iteration for the art.  I believe this is an issue on mostof the updated views.

PR #3543, opened 2021-08-27
[redesign][ln] Channels tab

bgptr:
Toward issue #3296

I used the StatusBar pi-ui component and needed a new type of layout, which has been implemented in https://github.com/decred/pi-ui/pull/354
Until it will be merged, I edited the package.json.

Notes:
- as @matheusd suggested, the "suggested nodes" are left out because figuring out good nodes is a hard problem and points beyond this PR.
- until the first channel has been created the search button is hidden because there is no way to discover the network. 

Updated screenshots:

DeepinScreenshot_select-area_20210906200639
DeepinScreenshot_select-area_20210906200838
DeepinScreenshot_select-area_20210902162507

DeepinScreenshot_select-area_20210906200415
DeepinScreenshot_select-area_20210906200428

DeepinScreenshot_select-area_20210906201122
DeepinScreenshot_select-area_20210906201609


app/components/views/LNPage/ChannelsTab/hooks.js
```diff
@@ -1,31 +1,137 @@
...
+      nodePubKey = split[0];
+      const addressSplit = split[1].split(":");
+      if (addressSplit.length > 2) {
+        setNodeErrorMsg(intl.formatMessage(messages.invalidNodeAddressFormat));
+        return false;
+      }
+    } else {
+      nodePubKey = node;
+    }
+    if (!secp256k1.publicKeyVerify(Buffer.from(nodePubKey, "hex"))) {
```
matheusd, line 126: Weirdly, seems this is missing to validate some obviously wrong ids. Not sure wether the secp256k1 lib here is not processing correctly or the `Buffer.from` call is silencing errors.

For example (`038fde001dbe4d6286ab168cfd1e9711ad0cbb8fc4e3c2312f2b42063b72af8e71` is a valid and existing ID on testnet):

```
038fde001dbe4d6286ab168cfd1e9711ad0cbb8fc4e3c2312f2b42063b72af8e7x  (last digit changed to invalid hex char `x`)
038fde001dbe4d6286ab168cfd1e9711ad0cbb8fc4e3c2312f2b42063b72af8e71xxx (added invalid hex chars)
03ce0b14fb842b1ba549fdd675c98075f12e9c510f8ef52bd021a9a1f4809d3b4c (invalid pubkey as tested [here](https://github.com/decred/dcrd/blob/7474441bbe947c718db98522e37701a9b27023e3/dcrec/secp256k1/pubkey_test.go#L106-L109))

**Review by matheusd — COMMENTED** [review]
Only found a small validation issue. Otherwise, UI is looking very nice

File: app/components/views/LNPage/ChannelsTab/hooks.js (hunk 1 above)
Comment (bgptr) [reply app/components/views/LNPage/ChannelsTab/hooks.js]: @matheusd It seems the `publicKeyVerify` function doesn't examine (properly) whether a point is on the curve. I've added an additional check using directly the elliptic library. Also, add hex validation.
Thank you for your review.

File: app/components/views/LNPage/ChannelsTab/hooks.js [hunk 2]
```diff
@@ -123,7 +128,22 @@ export function useChannelsTab() {
...
       nodePubKey = node;
     }
-    if (!secp256k1.publicKeyVerify(Buffer.from(nodePubKey, "hex"))) {
+
+    let isPubKeyValid;
+    try {
+      if (!isHexValid(nodePubKey)) {
+        throw new Error("pubkey is not a valid hex");
+      }
+      const pubKeyBuffer = new Uint8Array(Buffer.from(nodePubKey, "hex"));
```
matheusd, line 137: Seems this needs to drop the `new Uint8Array` bit, otherwise the UI throws an exception when copy pasting the address for my test node https://testnet-dcrln-01.matheusd.com/ (address is at the bottom of the page, in the footer)

matheusd (approved):
Working as intended now

app/components/views/LNPage/ChannelsTab/hooks.js
bgptr: The tests did not fail, and I left it by mistake :( 
Thanks for catching it.

PR #3546, opened 2021-08-30
settings: Update account passphrases on private passphrase change

alexlyp:
Previously we were just updating the private passphrase, but after the forced upgrade for account locking we need to update all accounts to the new private passphrase as well.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -209,12 +209,36 @@ export const changePassphraseAttempt = (oldPass, newPass, priv) => (
   dispatch({ type: CHANGEPASSPHRASE_ATTEMPT });
   return wallet
     .changePassphrase(sel.walletService(getState()), oldPass, newPass, priv)
-    .then((res) =>
+    .then(async (res) => {
+      if (priv) {
+        const oldAccounts = sel.balances(getState());
+        await Promise.all(
+          oldAccounts.map(async (acct) => {
+            // just skip if imported account.
+            if (acct.accountNumber === Math.pow(2, 31) - 1) {
+              return acct;
+            }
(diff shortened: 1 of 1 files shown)


app/actions/ControlActions.js
```diff
@@ -209,12 +209,36 @@ export const changePassphraseAttempt = (oldPass, newPass, priv) => (
...
   return wallet
     .changePassphrase(sel.walletService(getState()), oldPass, newPass, priv)
-    .then((res) =>
+    .then(async (res) => {
+      if (priv) {
+        const oldAccounts = sel.balances(getState());
+        await Promise.all(
+          oldAccounts.map(async (acct) => {
+            // just skip if imported account.
+            if (acct.accountNumber === Math.pow(2, 31) - 1) {
```
matheusd, line 218: Needs to be `>=` instead of `===`. While we don't have a way to import multiple accounts into a wallet right now, this is possible in dcrwallet so we could end up with multiple imported accounts at some point in the future.

alexlyp: Oh good point.  Though when an account is imported it starts backwards from the default Imported account right?  Not sure the proper catch in that case.

matheusd: No, imported accounts are still in ascending order, starting from the index that marks hardened account derivation (i.e. default imported account): https://github.com/decred/dcrwallet/blob/ae5292e5edcf21be17a966a03be402b3204c7689/wallet/udb/addressmanager.go#L1346-L1354

alexlyp: whoops I was mistaken.

matheusd (approved):
Tested by changing the passphrase on a cspp wallet.

PR #3548, opened 2021-09-01
[transactions] Limit number of non wallet outputs

alexlyp:
We were able to determine various crashes in 1.6.3 were caused by the processing of transactions that contained very large amounts of outputs.  So avoid this we will limit the number of outputs to check to 10.  Also this is just for display purposes in transaction details so it shouldn't effect any functionality.

### DIFF
--- a/app/actions/TransactionActions.js
+++ b/app/actions/TransactionActions.js
@@ -507,18 +507,29 @@ const getNonWalletOutputs = (walletService, chainParams, tx) =>
         Buffer.from(tx.rawTx, "hex"),
         chainParams
       );
+      // Some transactions with a lot of outputs for mining payments
+      // etc cause crashes in decrediton due to this repeated request for
+      // validate address.  Since this is merely for display purposes we will
+      //  we limit the amount of outputs to 10.
+      const nonWalletLimit = 10;
+      let count = 0;
       const updatedOutputs = decodedTx.outputs.map(async (o) => {
         const address = o.decodedScript.address;
-        // Validate address so we can check if it is our own.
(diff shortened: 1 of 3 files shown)


bgptr (approved):
LGTM

app/actions/TransactionActions.js
```diff
@@ -507,18 +507,29 @@ const getNonWalletOutputs = (walletService, chainParams, tx) =>
...
       const updatedOutputs = decodedTx.outputs.map(async (o) => {
         const address = o.decodedScript.address;
-        // Validate address so we can check if it is our own.
-        // If that is the case it is a change output.
-        const addrValidResp = await wallet.validateAddress(
-          walletService,
-          address
-        );
+        let isMine = false;
+        if (count < nonWalletLimit) {
```
matheusd, line 519: This isn't working as you intended, due to this condition (and the update below) happning inside an async function (starting at the `decodedTx.outputs.map(async (o) => {` above). 

Due to the async, you can get a race condition where more then 10 addresses are checked. I didn't have a tx with 100s of outputs to verify, but it could reliably still request `validateAddress` for all outputs when the tx had 30 outputs.

Since for the txs where this is a problem (i.e. the ones with 100's of addresses), it's probably best to just skip this function altogether if the tx has > 10 outputs and display a notice in the transaction detail page about output ownership not having been validated.

Also, you should probably limit the "To Address:" pane in the transaction details page as well.

alexlyp: Ah good catch.  I'll try that

matheusd (approved):

PR #3550, opened 2021-09-01
Bump CI node version to 16 from 15

alexlyp:


.github/workflows/nodejs.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 15.x]
+        node-version: [14.x, 16.x]
```
chappjc, line 9: While here, I'd give `actions/setup-node@v2` a try.  Supposedly it's faster.  Could even test it with `cache: 'yarn'` under `with:` to see if that speeds anything up.

matheusd:
Do note that the latest version of electron (v14.0.0) is _still_ using node v14.17.0, so for all intents and purposes we're limited to that node version.

PR #3551, opened 2021-09-08
[redesign][ln] Overview tab

bgptr:
Toward issue #3296
Requires #3543

Also, in this diff, as @MariaPleshkova suggested, the Wallet, Network, and Watchtowers tabs are grouped under a newly introduced Advanced tab.

Updated screenshot:


matheusd (approved):
Nice job on getting the transactions list to the overview page!

I feel the notation is slightly awkward, since (hopefully) no funds are lost on a channel creation so I'm not sure using a negative amount is correct there, but we can iterate on this stuff.

PR #3553, opened 2021-09-10
[transactions] Fix change address for mixed sends

### LINKED ISSUE #3541: Change amount from mixed account to DEX
When you go through CoinShuffle and you send funds from 'mixed' to 'dex', the change amount gets sent to the 'dex' account too. Wouldn't it make more sense to send the change back to 'unmixed'?

More generally, shouldn't all doxic change be sent back to the unmixed wallet?

alexlyp:
Fixes #3541 

Previously change addresses were the same as the sent addresses for mixed sends. 
This adds a new action for getting new change addresses.  This allows us to get new change addresses without messing with the new address requests which also change various GUI elements (receive account select).

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -704,7 +704,6 @@ export const constructTransactionAttempt = (
 
   dispatch({ type: CONSTRUCTTX_ATTEMPT, constructTxRequestAttempt: true });
   try {
-    console.log("there?", change);
     const constructFunc = all
       ? wallet.constructSendAllTransaction
       : wallet.constructTransaction;
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -36,6 +36,30 @@ export const getNextAddressAttempt = (accountNumber) => (dispatch, getState) =>
       });
   });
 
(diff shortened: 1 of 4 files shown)


bgptr (approved):
Tested and works as expected.

app/actions/ControlActions.js
```diff
@@ -679,6 +704,7 @@ export const constructTransactionAttempt = (
 
   dispatch({ type: CONSTRUCTTX_ATTEMPT, constructTxRequestAttempt: true });
   try {
+    console.log("there?", change);
```
bgptr, line 707: console.log

scripts/dexsite.js
```diff
@@ -3,7 +3,8 @@ var exec = require('child_process').exec;
 var os = require('os');
 
 if (os.type() === 'Linux' || os.type() === 'Darwin' ) 
-   exec("rm -rf bin/site && cp -R ./node_modules/dcrdex-assets/dexc/site bin/"); 
+   //exec("rm -rf bin/site && cp -R ./node_modules/dcrdex-assets/dexc/site bin/");
```
amass01, line 6: comment

matheusd (approved):

PR #3554, opened 2021-09-12
Fix error flows in create wallet

### LINKED ISSUE #3524: Fix error flows in create wallet
#3521 surfaced this

When an error occurs during the final stages of wallet creation / seed restore, the wallet is partially created and started but the app can't proceed with any operations.

There needs to be some rollback operation to return the app to a usable state.

bgptr:
Closes #3524

Found and fixed these errors in create wallet flow:
1. rejecting the final confirmation dialog (or receiving an error from the wallet create request) at the final stage of creation/restore wallet jumps back to the create/restore view (`PreCreateWallet`) and displays an error: `loader.OpenExistingWallet: item does not exist:: wallet.OpenDB: missing database file`. The wallet is partially created and started but until the restart of the app, it can not be stopped or start any other wallet. 
2. when an app refresh breaks the creation/restore process, a similar thing happens. The partially created/restored wallet is running in the background and it can not be stopped or start any other wallet. 
3. the flashing wallet selection screen after creating a wallet is also fixed. It is mentioned here: https://github.com/decred/decrediton/blob/c46d2f8e0fe40cebf16cecab1adb3b6056775bca/test/unit/components/views/GetStaredPage/PreCreateWallet.spec.js#L363/L365

### DIFF
--- a/app/actions/DaemonActions.js
+++ b/app/actions/DaemonActions.js
@@ -319,7 +319,15 @@ export const createWallet = (selectedWallet) => (dispatch, getState) =>
           selectedWallet.value.wallet,
           network == TESTNET
         );
-        await dispatch(startWallet(selectedWallet));
+        try {
+          await dispatch(startWallet(selectedWallet));
+        } catch (error) {
+          // This error message happens if we start creating a wallet. As its
+          // wallet.db file still not created.
+          if (!error.message.includes("missing database file")) {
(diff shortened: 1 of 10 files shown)


matheusd (approved):
Thank you for addressing this! This was a really annoying behavior.

Tested the listed situations.

PR #3555, opened 2021-09-13
[trezor] keep the trezor word modal alive during the restore device process

### LINKED ISSUE #3427: [trezor] dont unblur between words when restoring wallet
When restoring a model one trezor wallet (where seed words are input in Decrediton's UI), the UI should not unblur between individual words.

bgptr:
This keeps the trezor word modal alive during the restore device process and display loading state between the individual word request.

closes #3427


matheusd (approved):
Working as intendended.

PR #3558, opened 2021-09-22
Fix missing navigation links

bgptr:
@JoeGruffins reported an issue on the matrix: "If I go to restore from seed, and reject there, Settings are missing from the top right until I go to create/restore and go back"

This diff fixes it.


JoeGruffins (approved):
I see Settings now after rejecting a seed confirmation.

bgptr:
Sorry for keeping the typo. I've fixed it now.

alexlyp (approved):

PR #3559, opened 2021-09-23
[governance] Choice getId error fix 

alexlyp:
Previously, when the treasury agenda is clicked it would throw a choice.getId() is not a function.


JoeGruffins (approved):
No longer errors on mainnet.

On testnet, with dcrwallet on master, because something was updated there, I can change voting preferences for only `reverttreasurypolicy`. There are two others that say `... In Progress`, should I also be able to change those preferences, considering they are in progress?

<details closed>
<summary>evidence</summary>

image
image
image
</details>

matheusd (approved):

PR #3572, opened 2021-10-14
[dex] Add custom bitcoin directory select and refine conf handling

alexlyp:
Requires #3552

This PR adds the ability for users to select custom installation directories they may have for their bitcoin wallet.  

We have also removed any attempts at updating existing confs.  If these found confs do not have the required fields, a message is shown to the user what fields they must add and to come to the support channel if they encounter any further difficulty.

### DIFF
--- a/app/components/views/DexPage/CreateWalletsPage/CreateWalletsPage.jsx
+++ b/app/components/views/DexPage/CreateWalletsPage/CreateWalletsPage.jsx
@@ -13,6 +13,7 @@ const CreateWalletsPage = () => {
     onCreateWalletDex,
     createWalletDexAttempt,
     onBTCCreateWalletDex,
+    btcCreateWalletDexAttempt,
     dexDCRWalletRunning,
     dexBTCWalletRunning,
     dexAccount,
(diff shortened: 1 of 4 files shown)


chappjc:
When it becomes an option to use dexc's upcoming built-in btcwallet, or which there's no .conf, will it be straightforward to skip or dismiss these checks, possibly based on a dialog regarding wallet type choice?
This may happen sooner than later.

alexlyp:
@chappjc Ya I don't foresee it to be an issue at all really.  I'm envisioning something to the effect of adding an option of the user selecting whether or not they have a previous install of bitcoind/qt.  If not then the built-in will be used.

bgptr (commented):
Tested and it works. Left only a few nits.

matheusd:
The issue with closing and opening a different wallet seems fixed, but I'm still getting stuck with the "create BTC config" button: after clicking on it, I can't proceed in attempting the connection or doing anything.

Also noticed something while testing this PR, but might not have been caused by it: attempting to create a new DEX account when the wallet has privacy enabled fails after the account is created.

alexlyp:
@matheusd sorry totally missed your comment about the create BTC config button! Will take a look now.  As well as the other issue you mention.

matheusd (approved):
Working as intended now.

PR #3569, opened 2021-10-13
i18n: Add custom translation file and doc

matheusd:
This allows loading custom translation files for the main UI (.json) and adds a doc explaining the translation process for communitity members.

A development build with the ability to import the custom translation files is available below. [Preview the doc](https://github.com/matheusd/decrediton/blob/community-i18n/app/i18n/community_translators.md) for community translators.

**Warning**: Ideally only use those binaries on a VM to ensure it doesn't mess around on your existing install.

https://github.com/matheusd/decred-weekly-builds/releases/tag/v20211013150209

PR #3588, opened 2021-11-02
i18n: Translator upgrades

### LINKED ISSUE #3583: i18n: Translator upgrades
Below are ideas how to further improve [translator.html](https://github.com/decred/decrediton/blob/dc00928008d37072146ad4141eb9d81436cc7771/app/i18n/translator.html) collected while fixing issues listed in #3575.

Since it is a small single-file tool I'm putting it all into one issue to not spam the issue tracker with small stuff.

**User-facing changes**

- [ ] Always show all input fields
  - previous string is hidden if it is empty or equal to original string
  - this causes frequent changes to form's structure, confusing the eye and making the buttons jump up and down
  - instead, always show all fields but gray out the previous string if it is equal to new string

- [ ] Add label showing the current string key
  - sometimes I found myself asking "what key I'm looking at?" and the only way to know it is to find which item is selected in the left-hand list
  - put the label above Previous Original String input

- [ ] Add label for the translated string input
  - previous and original inputs have labels but not the translated input
  - an extra label will help to get familiar a bit faster

- [ ] Fix scroll position getting carried from one list to another
  - if I scroll down on one list and switch to another, it gets scrolled by the same %
  - ideally, make the lists scrollable independently

- [ ] Move "copy original" button (`V`) before "save & next" (`>`)
  - to match the order in which they are used

- [ ] Rename `>` and `V` to `save & next` and `copy original`
   - helps to learn the UI faster
   - increases the click area
   - alternatively, add tooltips to explain what these buttons do

- [ ] Jump to first item after `save & next` on last item
   - looping over the list sounds more fun than current behavior

- [ ] Add button to jump to previous string
  - sometimes I missed it after doing a mistake and hitting `save & next`

- [ ] Place Deleted Entries as another radio button + list
  - currently it looks out of place at the bottom
  - moving it along the other lists streamlines the layout, to expose what string was deleted
  - maybe also populate previous string input (read-only) on click
  - after this change it will be easier to make the left-hand list area span all visible vertical space (currently it is hardcoded to `34em`)

- [ ] Navigate to item selected by `save & next` action
  - switch to the list where the "next" item was selected (All/Untranslated/Changed), even if the user changed to a different list before hitting `save & next`
  - scroll the list to make the newly selected item visible

- [ ] Show app versions of previous and original strings
  - change string labels from "Previous Original String" and "Original String" to e.g. "String in v1.6.3" and "String in v1.7.0-pre"
  - this requires getting these versions somewhere, e.g. add a meta `_version` key to `previous_original.json` and `original.json`
  - the idea is to increase translator's awareness of what is being compared

- [ ] Investigate if diff highlighting is easy to do
  - highlight the diff between the "previous original" and "original" strings
  - for instance, it is not easy to detect changes when there is a new whitespace/eol that has been added/removed somewhere or when long Strings are compared
  - originally [suggested](https://github.com/decred/decrediton/issues/3575#issuecomment-949547398) by @JaseurPattu
  - I guess it will add a lot of complexity and take time to implement, but there could be a quick hack to help with the described use case

**Internal**

- [ ] Do not re-prepare file until download is requested

- [ ] Do not `saveTranslation()` on a mere select (click in the list), only save on `save & next` action

- [ ] Use strict comparison (`===` and `!==`) everywhere, verify nothing breaks

- [ ] Optimize finding next key
  - find a way to not re-scan all keys from 0 in `nextKey()`
  - e.g. maybe store an integer index in `state` so that figuring out next key becomes a simple `current_list[state.index + 1]` or something like that (mind edge cases ofc)

- [ ] Wrap code to whatever line length is adopted in this repo

bgptr:
Closes #3583

Notes and screenshots:
- for diff highlighting, I've imported http://github.com/google/diff-match-patch. This feature would be difficult to implement from scratch, but this lib does the heavy lifting quite nicely. I've added a toggle button for viewing the original string for convenience.
showdiff
hidediff
t1
t2


xaur:
Per [this comment](https://github.com/decred/decrediton/pull/3569#issuecomment-949579988), `previous_original` comes from some older v1.5.x. I would use `v1.5-approx` in `meta.version` for now. When all strings are ready, this file will be [updated](https://github.com/decred/decrediton/pull/3569#issuecomment-951896367) to `v1.6.3`

xaur:
Wow this is incredible and way beyond my expectations. Thank you for working on this!

I will be testing this in the coming days, but one quick thing is, is it possible to remove requests to the Eye of Cloudflare? Like commit the lib in this repo (if maintainers are ok with it) or at least use a different hosting (decred.org perhaps)?

To clarify, I will not be a heavy user of this tool and real translators might be ok with it, but asking just in case there is an easy solution.

app/i18n/translator.html
```diff
@@ -126,103 +165,278 @@
...
 
         const downBtn = elID("down_translation_btn");
         downBtn.href = url;
         downBtn.download = "custom-translation.json";
       }
 
+      function isEntryDeleted(key) {
+        return (
+          state.original[state.selectedKey] === undefined &&
+          state.translated[state.selectedKey] !== undefined
```
xaur, line 187: Haven't tested yet but just logically I think an entry is considered "deleted" if it existed in the previous English file and is missing in the new English file.

bgptr: I've simplified the condition. It checks now if the key is in the `deletedEntries` which is created in the `startTransaction` function and I did not alter the logic there.

bgptr:
I'm ok with committing the lib in the repo. 
@alexlyp, what are your thoughts about it?

xaur:
Let's add for context that we're talking about `diff_match_patch.js` (21 KiB [compressed](https://github.com/google/diff-match-patch/blob/master/javascript/diff_match_patch.js), 77 KiB [uncompressed](https://github.com/google/diff-match-patch/blob/master/javascript/diff_match_patch_uncompressed.js)).

xaur:
Allright, sorry for the wait. First round of testing.

New features look great. This tool is much more powerful now.

Bugs:

- on `Deleted` list, `save & next` button should change to `next` and become enabled
  - match behavior of the `previous` button
- `previous` now means "previous in the navigation stack" and not "previous in the list of the selected entry"
  - saving >1 pairs `{filter, key}` in a navigation stack produces confusing behavior
  - e.g. do `save & next` on `Unstranslated`, change to `Changed`, mouse-click an entry, hit `save & previous` -> view changes to `Untranslated` instead of staying in `Changed` and going to the previous entry
  - "previous" shold be symmetrical to "next": next goes to next entry in active list, previous goes to previous entry in active list (and not in its own nav stack)
  - I suggest to remove the stack and only store 1 pair of "selected entry" + "filter name for selected entry"
  - `jumpToPreviousKey()` will need to be updated since it relies on `previousKeys.length > 1`
  - update selected entry and its filter on any method of selection (mouse click, `save & next/previous`)

Tweaks (optional, can do in future PRs):

- increase font size of gray labels a bit
  - in both left-hand list and right-hand fields
  - value `80%` felt good (current `60%`)
  - bigger font looked better with 1-2px extra top padding

- paint key labels in `Deleted` list in same gray color as key labels on other lists

- make `Hide diff` button all lowercase for consistency

- rename `Download Translation` to `save translation` or `export translation` or `export .json`
  - it's all local so no "download" is taking place
  - lowercase for consistency
  - opinions on what is more intuitive, `translation`/`.json`?

- make left-hand list width fixed
  - is is jumping when switching between `Changed` and `Deleted` lists

Performance (optional, can do in future PRs):

- operations on the `All` list are slow (changing to it, clicking on entries, `save & next`)

app/i18n/translator.html
```diff
@@ -128,111 +165,280 @@
...
+            filterVal === "changed" &&
+            state.changedEntries.indexOf(k) === -1
+          ) {
+            continue;
+          }
 
           // When on Changed view, skip to next key considered "changed".
           if (filterVal === "changed" && state.changedEntries.indexOf(k) == -1) {
             continue;
           }
```
xaur, line 395: Looks like a duplicate of the previous code block. Better remove this one since it has one remaining `==`.

app/i18n/translator.html
```diff
@@ -128,111 +165,280 @@
...
       function nextKey() {
         let gotPrev = false;
         let keys = Object.keys(state.original);
-        const filterVal = document.querySelector('input[name="filter"]:checked').value;
-        for (let i = 0; i < keys.length; i++) {
+        let filterVal = getFilterVal();
+        const { filterVal: previousFilterVal } =
+          state.previousKeys.length > 0 &&
+          state.previousKeys[state.previousKeys.length - 1];
+        if (previousFilterVal !== undefined && filterVal != previousFilterVal) {
```
xaur, line 371: One remaining `!=`.

app/i18n/translator.html
```diff
@@ -128,111 +165,280 @@
...
+        }
         let v = state.translated[newkey] ? state.translated[newkey] : "";
         elID("translated_str").value = v;
-        updateState({ selectedKey: newkey });
+        const filterVal = getFilterVal();
+        let delta = {
+          selectedKey: newkey,
+          selectedFilterVal: filterVal
+        };
+        if (newkey != state.selectedKey) {
```
xaur, line 214: One remaining `!=`.

app/i18n/translator.html
```diff
@@ -128,111 +165,280 @@
...
+              return;
+            }
+            if (
+              filterVal === "changed" &&
+              state.changedEntries.indexOf(k) === -1
+            ) {
+              return;
+            }
+            if (
+              filterVal === "deleted" &&
```
xaur, line 298: This `if` block may never run. It is in the `else` branch of `if (filterVal === "deleted")` on line 273.

xaur:
Diffing looks very helpful for translator's awareness and productivity, thanks for investigating!

Noticed the page uses `20121119/diff_match_patch.js` from the Cloudflare CDN, but Google's [Git repo](https://github.com/google/diff-match-patch/tree/master/javascript) has a newer version (+7 years but not much changes :O). If we import it, let's use the newer one.

@alexlyp or maybe @matheusd any objections to [committing](https://github.com/decred/decrediton/pull/3588#issuecomment-961907165) the lib to this repo?

app/i18n/translator.html
```diff
@@ -108,12 +141,16 @@
         original: {},
         translated: {},
         selectedKey: "",
+        previousKeys: [],
         seenKeys: [],
```
xaur, line 145: `seenKeys` looks unused.

matheusd:
> @alexlyp or maybe @matheusd any objections to committing the lib to this repo?

Seems reasonable to me, and indeed better then to hit CF every time this is opened.

bgptr:
@xaur, thanks for the review. I've pushed the fixes. I didn't improve anything on performance now, but I will after it functionally works as expected.

xaur:
Tested:

- two bugs found earlier do not reproduce
- All list is way faster now: initial selection of it takes a bit of time, but selection of items in it, previous/next - are all fast
- bigger label font is nice
- left list width no longer jumps

One more idea that crossed my mind is to load the translated string even on the Deleted list keys, to allow translators to recover and reuse any strings that will go away. This is optional, we can do this later.

Otherwise, I see no functional bugs. Great work!

app/i18n/translator.html
```diff
@@ -128,111 +169,269 @@
...
-        const filterVal = document.querySelector('input[name="filter"]:checked').value;
-        for (let i = 0; i < keys.length; i++) {
-          const k = keys[i];
-          // When on Untranslated view, skip to next key considered
-          // "untranslated", i.e. undefined (new key) or "" (existing key with
-          // empty translation).
-          if (filterVal === "untranslated" && state.translated[k]) {
-            continue;
-          }
+      function jumpBackToSelectedList() {
```
xaur, line 382: Nit: `Back` may be removed from the name I think to avoid confusion.

xaur:
Code-wise, good idea to add the uncompressed lib in case we'll ever need to tweak it.

I see a big diff with [upstream](https://github.com/google/diff-match-patch/blob/master/javascript/diff_match_patch_uncompressed.js) master (looks formatting only), which may complicate porting upstream changes.

app/i18n/translator.html
```diff
@@ -128,111 +169,269 @@
...
+            state.translated[state.selectedKey] = newValue;
+
+            // update untranslated list if necessary
+            const el = elID(`li-untranslated-${state.selectedKey}`);
+            if (newValue !== "" && el) {
+              el.remove();
+            } else if (newValue === "" && !el) {
+              const stringList = elID("string_list_untranslated");
+              stringList.innerHTML = "";
+            }
```
xaur, line 207: I have not tested this code path and wonder what it does?

Reads like "if new translated value is empty AND the element for selected key not found in the untranslated list, then... clear the entire untranslated list"?

bgptr: You read the code well.
If the user, for some reason, clears a previously translated translation, the item becomes untranslated. It should be added back to the untranslated list and be rerendered by clearing the related DOM. Another solution would be just adding back only the certain item. Since it could be a rare event, I think it's not too expensive to rerender the entire untranslated list with the proper order of the original list.

xaur: It makes sense now, thanks.

xaur:
@bgptr the uncompressed version looked more future-proof in terms of maintenance, did you found a reason how the compressed one is better?

bgptr:
@xaur I considered only the file size, but you're right. The uncompressed version could be more future-proof. I've updated it.

xaur:
Ok great. There's still a diff with [upstream](https://github.com/google/diff-match-patch/blob/master/javascript/diff_match_patch_uncompressed.js) lib but it's up to you guys. The thing is working pretty good.

@matheusd you may want to pick the review from here since you kickstarted this tool and are familiar with code.

xaur:
@Insaf01 helped me verify that we have a common issue with bidirectional text, where a mix of Arabic and English is ordered incorrectly. For example, this is for key `agenda.card.finishedTooltip` in the `ar.json` file:



and this is how it shold look like:



I hacked the `<textarea>` element of the translation box to add the `dir="rtl"` attribute and I think it fixes it:



While this PR is not merged, we can use it to fix this issue too.

If my hack is good enough, we'll need to conditionally set the `dir` attribute based on something. How about, add the `meta` object to the translated files too, and if it's present and has the `dir` attribute, set it in the text area?

bgptr:
@xaur I've set the `dir` attribute to `auto`. It solved the issue for me.

xaur:
I also now see Arabic text flowing right-to-left and `auto` is a much simpler solution, thanks @bgptr. I'll ask Insaf to confirm that it looks proper now.

matheusd (approved):
Working as intended. I also verified the imported diff_match_path.js matches the one from the original github repo.

PR #3594, opened 2021-11-16
[redesign] Confirm Seed view

### LINKED ISSUE #3593: Seed verification view updates
Seed Verification step updates based on [#2345 (Enhance Decrediton seed confirmation) ](https://github.com/decred/decrediton/issues/2345#issue-528505906)  and [#232 (Make seed verification less painful)](https://github.com/planetdecred/dcrandroid/issues/232#issue-381933869).

You can find the updated designs here https://www.figma.com/file/EanUbqJmPs8EMqY26j9zwo/decred---piui---decrediton?node-id=363%3A0 (Decrediton - Latest page)

In the current desktop version users have to type in the first few letters of every missing word in the seed and select the right one, which is pretty long and painful process from a ux perspective. @jzbz suggested to use similar concept that is already implemented for the mobile clients and that's a good idea as this can simplify user interaction with this step a lot. 

So, the new process of confirming the seed will be like this (there should be all 33 rows as on mobile, but used only some of them on the layout just to show the concept. Though any smaller amount would be more convenient, but I guess using all 33 is more safe) :
image

Here users just have to choose and click on the right word within each of the 33 rows.

As the whole page doesn't fit the screen height, there is a bar stuck to the bottom that shows "Create Wallet" and "Back" buttons. 
image

When scrolled till the end of the page the bar disappears and transforms into static buttons under the Private Passphrase fields. When starting to scroll up, the bar on the bottom appears again
image

bgptr:
Closes #3593

Notable changes:
- implement redesigned Confirm Seed view
- found and fixed a bug on the existing seed view. The error message did not show and was not indicated on the controls.
- change the Private Passphrase section on the existing seed view also for the sake of consistency (screenshot attached)
- update automated tests

Screenshots:
image
image
image
image


app/components/views/GetStartedPage/messages/messages.jsx
```diff
@@ -58,6 +59,10 @@ export const messages = defineMessages({
   passphrasePlaceholder: {
     id: "getStarted.discoverAddresses.passphrasePlaceholder",
     defaultMessage: "Private Passphrase"
+  },
+  confirmSeedMissingWordError: {
+    id: "getStarted.confirmSeed.missingWordError",
+    defaultMessage: "*Please confirm the missing words"
```
matheusd, line 65: I know this is the original message, but re-testing, this error reads weird when you have verified all seed words, but one of them is wrong. There are no "missing" words, but there _is_ a "wrong word". If possible, I suggest differentiating between not having entered all words yet and an actually wrong word.

Another thing I've noticed is that there's now no feedback on _which_ word is the wrong one. Is this intentional?

bgptr: @matheusd, thanks for the review.
The error message differentiating is a good idea. I've added it.

I've left out the wrong word feedback intentionally, but I'm not sure about it. Since the goal of this confirmation view is to ensure the user wrote down the correct seed words, I feel the error messages give too much room for being lazy and easy to click through this view by trial and error.

I've added the error feedback though. It is easier to see this during operation. 
It is not in design, and I choose the colors. @MariaPleshkova, please take a look at the screenshot of it.

matheusd: > I feel the error messages give too much room for being lazy and easy to click through this view by trial and error.

Indeed, displaying the error directly upon clicking seems too easy to cause the user to be lazy.

Another idea would be to only do the seed validation after the user fills all words, and _then_ display whether a particular word is wrong or not.

bgptr: Good idea 👍️

MariaPleshkova: Agree with @matheusd regarding the seed validation after the user fills all words. And the red colored wrong words are also look good. I'd also suggest to show maybe an error message after validation as well? Like "you have chosen the wrong words" or "some words from the seed are incorrect" or something like that. And a user can be automatically moved up to the top of the page after validation and see this message. 
An example:
image

bgptr: @MariaPleshkova There was an error message on the bottom but the screenshot did not contain it.  I've updated it following your suggestion and uploaded a fresh screenshot to the description. Thank you

JaseurPattu:
I may have seen a bug related to these changes while testing the latest master.
I am able to avoid the seed phrase verification and create the wallet on the seed phrase confirmation screen with the following scenario :
- don't select any word in the list
- type the private passphrase
- type the private passphrase confirmation
- click on create wallet button

bgptr:
@JaseurPattu, thanks for the report. 
Which network are you on? On testnet, it's intentional and used for testing.

JaseurPattu:
Yes, I'm on testnet, I was not aware that behaviour was intentional. 
Many thanks for the explanation and sorry for the wrong report.

PR #3595, opened 2021-11-17
Fix Transactions History Range Filter

### LINKED ISSUE #2874: [1.6.0-rc2] Transactions History Range Filter from 0-100
When opening the transactions history the range filter for the dcr amount is set to 0-100 per default.

Once the min amount is moved to a minimum of 1dcr to filter out all dust mix transactions the range filter is not able to be put back to 0 to see all transactions again.
image

bgptr:
Closes #2874


matheusd (approved):
Nice catch! Tested fine.

PR #3608, opened 2021-11-29
update for dex 0.4

chappjc:
This updates to the current dcrdex v0.4 release branch, which is presently in a beta status as dcrd and dcrwallet modules are yet to be fully-updated for the Decred 1.7 release.

The changes include new wallet "types" in the `CreateWallet` and `ReconfigureWallet` methods, one of which is `"SPV"` for BTC.  The `"SPV"` type for BTC indicates a native/built-in btcwallet (in SPV mode).
The `ReconfigureWallet` method also now takes its arguments in a struct.

Having chatted with @alexlyp on matrix, I believe the plan for the BTC wallet creation/updates is to just remove all actions from within Decrediton to instead have the user setup their BTC wallet (or not) from within the DEX window.  I believe this is best for two reasons.  The first is that the user may choose between a "Native" or "External" BTC wallet:

image

Also, v0.3/v0.4 bring the ability to pay the registration fee in BTC as well as DCR:

image

Given this change to registration, should Decrediton also stay out of the account creation / registration process?  I believe a primary motivation for Decrediton helping with registration was to save the user typing in the host name, but another 0.4 change is that it is a predefined option:

image

And they only need to click on that host to get the registration fee asset selection dialog:

image


app/selectors.js
```diff
@@ -1932,6 +1932,15 @@ export const dexDCRWalletRunning = compose(
   dexUser
 );
 
+export const dexBTCWalletConfigured = compose(
+  (user) =>
+    user &&
+    user.assets &&
+    user.assets["0"] &&
+    user.assets["0"].wallet, // nil *WalletState if not configured
```
chappjc, line 1940: The `dexBTCWalletRunning` function below this goes on to check `running`, but in CreateWalletsPage.jsx it may be important to know if it is actually configured yet, rather than both configure and connected to the external bitcoind.  If it is not yet configured, they the user has the opportunity to pick which wallet type they want to use (Bitcoin Core external or SPV native/built-in).

app/components/views/DexPage/CreateWalletsPage/CreateWalletsPage.jsx
```diff
@@ -47,6 +48,7 @@ const CreateWalletsPage = () => {
   return (
     <div className="flex-column align-start">
       {!dexBTCWalletRunning ? (
+        /* !dexBTCWalletConfigured ? ( offer choice between native bitcoind RPC and native SPV ) */
```
chappjc, line 51: Depending on their choice in this hypothetical dialog, the Decrediton would call `btcCreateWalletDex` with a `walletType` arg of either `"bitcoindRPC"` or `"SPV"`.

alexlyp (approved):

PR #3620, opened 2021-12-24
use the tagged dcrdex v0.4.0-rc1

chappjc:
Since the last revision referenced in the go.mod, this includes a check for the Bitcoin Core wallet type to ensure it is not a "descriptor" wallet, which does not support a number of standard RPCs (!).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/decred/decrediton
 go 1.16
 
 require (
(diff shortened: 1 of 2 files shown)


chappjc:
The gosum db is slow to pick this up at https://sum.golang.org/lookup/decred.org/dcrdex@v0.4.0-rc1
I had to `GOSUMDB=off go mod tidy`, but hopefully that will update.

alexlyp (approved):

PR #3621, opened 2021-12-27
trezor: Use upstream trezor connect.

JoeGruffins:
I'm not sure when it happened, but it looks like they merged my connect work so we can use their `trezor-connect@8.2.4-extended`

@matheusd please have a good look at this if you could, test it out.

I believe it still shouldn't work with stake outputs yet, so needs to be a "clean" trezor.


package.json
```diff
@@ -318,12 +317,14 @@
     "redux": "^4.0.5",
     "redux-thunk": "^2.3.0",
     "reselect": "^4.0.0",
+    "secp256k1": "^4.0.2",
```
JoeGruffins, line 320: It wanted this for LN stuff, otherwise I am unable to do "yarn build"

JoeGruffins:
I'm confused about the test failure, it couldn't pull a dependency?

JoeGruffins:
Still failing at an odd place `test/unit/components/views/LNPage/ChannelsTab.spec.js`

JoeGruffins:
Ticket purchasing still seems fine in #2869

matheusd (approved):
Tested on the emulators for trezor model T and model one.

Issue #3404, opened 2021-04-22
Common spacing rules


MariaPleshkova:
Some spacing rules has been changed during the latest design updates. So here is a short description of some main ones:

Generally now we use spacing values that are multiples of 5. More popular ones are 5px, 10px, 15px, and 30px.
Also for the smallest distances we use 1px (for example, between the rows in tables):
![image](https://user-images.githubusercontent.com/8841018/115724954-16e73e80-a38a-11eb-8a91-31fc54c255af.png)

Below you can find some most common patterns and spacings there.

- The spacing to the left and right of the main content area is 60px:
![image](https://user-images.githubusercontent.com/8841018/115725026-25cdf100-a38a-11eb-91b9-713f93721889.png)

- The spacing from the top of the window to a title is 40px:
![image](https://user-images.githubusercontent.com/8841018/115731675-ff12b900-a38f-11eb-8412-2837161bce3a.png)

- Basically it’s one common pattern for all such top sections (let’s call them headers):
![image](https://user-images.githubusercontent.com/8841018/115731719-08038a80-a390-11eb-8fc0-3212d2aab1ee.png)

- And when there is no tabs we still use 30px under description:
![image](https://user-images.githubusercontent.com/8841018/115731766-12258900-a390-11eb-9ca7-83c08281276b.png)

- 30px between the header and a title below:
![image](https://user-images.githubusercontent.com/8841018/115731831-2073a500-a390-11eb-805c-92aa8b7ec41d.png)

- And 15px between a title and its section:
![image](https://user-images.githubusercontent.com/8841018/115725496-91b05980-a38a-11eb-99b7-188ddb47a2b5.png)

- 30 and 40 pixels are the values inside content section:
![image](https://user-images.githubusercontent.com/8841018/115731878-29fd0d00-a390-11eb-9ccd-9fda745536a3.png)

- And 20px between the section and a button below it:
![image](https://user-images.githubusercontent.com/8841018/115725574-a55bc000-a38a-11eb-9649-412ae5178d80.png)

- The spacings within dropdowns are 15px from all sides:
![image](https://user-images.githubusercontent.com/8841018/115732755-f1116800-a390-11eb-85c0-e3108d7fb6ec.png)
![image](https://user-images.githubusercontent.com/8841018/115732861-08505580-a391-11eb-801c-115e3e64c7e0.png)

[...]

PR #3626, opened 2021-12-30
Fix header spacing on Governance tab

jholdstock:
Ensure the spacing between Title and Description is 5px, as specified in #3404. More detail in commit messages.

### DIFF
--- a/app/components/layout/TabbedPage/TabbedPage.module.css
+++ b/app/components/layout/TabbedPage/TabbedPage.module.css
@@ -21,14 +21,14 @@
 
 .tabs {
   background-color: var(--background-back-color);
-  padding: 40px 60px 0 96px;
+  padding: 25px 60px 0 96px;
   position: relative;
   display: flex;
 }
 
 .tabContent {
   background-color: var(--background-container);
-  top: 71px;
+  top: 55px;
   bottom: 0;
   left: 0;
   right: 0;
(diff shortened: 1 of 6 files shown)


bgptr (changes requested):
@jholdstock, thanks for this PR. Looks nice, only the description part looks a bit more crowded now. There should be `10px` margin between the description lines in the latest design: https://www.figma.com/file/EanUbqJmPs8EMqY26j9zwo/decred---piui---decrediton?node-id=92%3A2457

design vs current state:
image

jholdstock:
Right, I see the problem here. I was only seeing one line of description because I had the external HTTP request for treasury balance disabled. Fixing

jholdstock:
Pushed some changes,  I think this now also fixes #3627

app/components/layout/TabbedPage/TabbedPage.module.css
```diff
@@ -21,14 +21,14 @@
...
-  padding: 40px 60px 0 96px;
+  padding: 25px 60px 0 96px;
   position: relative;
   display: flex;
 }
 
 .tabContent {
   background-color: var(--background-container);
-  top: 71px;
+  top: 55px;
```
jholdstock, line 31: :nauseated_face:

bgptr (approved):
Tested, and looks nice.

alexlyp (approved):

PR #3611, opened 2021-12-08
Split `SettingsPage/SettingsTab/Settings.module.css`

bgptr:
This diff splits `SettingsPage/SettingsTab/Settings.module.css` into smaller css modules and cleans up the structure.

PR #3631, opened 2022-01-02
[redesign] Settings view

### LINKED ISSUE #3613: Settings views updates
In the current release (1.6.3) all sections of Settings are on one page and, thus, a bit out of our components library (had to use subtitles within sections there, bit it's not the way we show them). So we reorganized the view a bit.
We split a view into 3 main parts (Connectivity, General, Privacy and Security) and reworked them into a tab view.
Here is an example from the Settings in Launcher:
image
Basically there are just 3 separate pages now, so all information is broken down into 3 categories(tabs). This way it's more clear where needed info is located.

For Settings in the Launcher we use two-columns grid for showing subitems on the page, because there's enough space for that:
image

What concerns Settings within an existing wallet, we mostly have to use one-column grid there (big left part of the screen is taken by the menu). But if there are short sections such as Timezone and Misc, then we can place both of them in one row.
image
image

For wider screens we can still always place 2 sections in a row:
image

And as there are already some tabs (Settings, Sources, Tutorials and Logs), we just split Settings tab into 3 new ones (Connectivity, General, Privacy and Security). So now there are 6 tab items in total.

Also we added the title ( Address Usage) above the Discover Address Usage button. This way it's more clear that this button is the separate option within Misc.

What concerns Privacy and Security tab, Private Passphrase updating is now out of the section, and the section with checkboxes is renamed to Outbound Connections.
image

Please, notice that we also added a new section with Launcher option ticker to the General tab. Describing what is this for in #3614 (point 4, when the wallet is launched)
image

_P.S.There is a question to think about though. It is a save feature. Now the Save button is always out of the visible area (I need to scroll to see it). So initially I as a user have no idea that I have to click on it to save the changes.
So I see 2 options here:_
_1. Add an autosave feature after every change (that would be the best one in my opinion)
2. Add Save button below each section_

Launcher Settings can be found within Concepting-Launcher page in Figma (see also 2 more pages to the right): https://www.figma.com/file/EanUbqJmPs8EMqY26j9zwo/decred---piui---decrediton?node-id=2978%3A11507

In-wallet settings design is within the Decrediton-Latest page:  
https://www.figma.com/file/EanUbqJmPs8EMqY26j9zwo/decred---piui---decrediton?node-id=3103%3A12223

bgptr:
Builds upon #3611
Requires https://github.com/decred/pi-ui/pull/413 (merged)
Closes #3613

Notable changes:
- implements the redesigned Settings view.
- applies pi-ui `Select` component & ditch `react-select` lib
- removed the save button and implemented the autosave feature. Also added a `Settings saved successfully.` snack bar.
- update automated tests

Screenshots:
image
image
image
image


alexlyp:
Ready to give this a PR a look.  Please rebase.  Thanks!

matheusd (approved):
Tested fine.

Nice job getting rid of react-select!

alexlyp:
Overall looking great.  

One thing that stood out is the usage of the tooltips on the Connectivity tab.  Is the idea to say that it will require a restart after changing these?  Currently doesn't make a ton of sense

bgptr:
@alexlyp Actually, I think the tooltips should indicate if a certain parameter was set from cli. The `pi-ui` `Tooltip` component does not support the `disabled` prop which I didn't notice. Nice catch, I've fixed it.

alexlyp (approved):

PR #3633, opened 2022-01-04
Fetch `vspinfo` async

### LINKED ISSUE #3622: Hit VSP endpoints in parallel
Upon opening a wallet, Decrediton hits the `/vspinfo` endpoint of every known VSP. It does each of them one by one which takes a significant amount of time. It should be possible to run these requests concurrently which would smooth the startup experience quite noticably, especially if the user has a poor network connection.

bgptr:
Closes #3622

### DIFF
--- a/app/actions/VSPActions.js
+++ b/app/actions/VSPActions.js
@@ -18,6 +18,7 @@ import { USED_VSPS } from "constants/config";
 import * as cfgConstants from "constants/config";
 import { shuffle } from "helpers";
 import { mapArray } from "fp";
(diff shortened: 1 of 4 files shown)


matheusd (approved):
This tested fine, though TBH I didn't time a significant difference in testnet (might be different when running on mainnet due to the nb of VSPs)

jholdstock:
Yep thats right, 15 mainnet VSPs vs 3 on testnet.

alexlyp (approved):

PR #3643, opened 2022-01-05
[LN] Show startup progress and log line

### LINKED ISSUE #3636: Properly track LN sync progress
I'm trying to open a testnet LN wallet which hasnt been opened since ~300,000 blocks ago. It's catching up to tip, and Decrediton shows `Syncing LN wallet to network`, but after 60 seconds decrediton kills dcrlnd and shows message `dcrlnd wallet not synced after 60 seconds`.

I guess decrediton assumes the sync should be completed in <60 seconds, but this assumption is not correct.

A proper solution would not timeout after a fixed period, but query sync progress and only timeout if progress halts. A progress bar would be nice as well, similar to how dcrd sync is displayed.

matheusd:
This adds a linear progress component during the router pruning stage of the
LN wallet startup. It also adds the last log line of the dcrlnd process to the
startup UI and removes the 60 second limit of the startup sync time.

These help mitigate issues when opening an LN wallet that has been offline for
a long time.

~Depends on https://github.com/decred/dcrlnd/pull/153~ (merged)
Closes #3636

### DIFF
--- a/app/actions/LNActions.js
+++ b/app/actions/LNActions.js
@@ -378,21 +378,35 @@ const connectToLNWallet = (
   return { client: lnClient, wtClient };
 };
 
-const waitForDcrlndSynced = (lnClient) => async () => {
+export const LNWALLET_WAITSYNC_PROGRESS = "LNWALLET_WAITSYNC_PROGRESS";
+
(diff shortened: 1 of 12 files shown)


bgptr (approved):
Tested and works as expected. 👌

alexlyp (approved):

PR #3645, opened 2022-01-05
[dex] Add BTC SPV question and remove Register

alexlyp:
This adds the ability to chose whether or not you'd like to use the new native (spv/neutrino) btc wallet that is included within dexc.

We have also decided to bypass the Register page and actions since the funnel within dexc is much more flushed out and allows for payment from a variety of assets instead of just DCR.  I have left the RegisterPage and Actions in for now and when we have finalized the functionality we can remove all that is unused.


matheusd (commented):
First pass test. Having the ability to use dex from within decrediton without having to setup a bitcoincore wallet is a game changer in usability!

Trying to open the dex window after reloading the UI fails (there's a flash but the dex window isn't launched)

matheusd (approved):
This looks good as is, modulo the issue I've noted (but I guess that's not super critical)

alexlyp:
Hrm I'll open an issue for that.  I was able to repro it

PR #3652, opened 2022-01-07
Fix Peer Count display

alexlyp:
Currently peer count is broken and always shows 0.  I'm not sure the reason for checking resp.wrappers_ in the past, but doesn't seem necessary anymore.


bgptr (approved):
LGTM

matheusd (approved):

PR #3653, opened 2022-01-07
[dex] Add ExportSeed and ConfirmSeed page

### LINKED ISSUE #3646: [dex] Show and confirm dex seed during init
The dex seed is provided in the response from the init call to dexc when setting the app password.  This seed should be shown to the user and they should be alerted to it's importance and the need to back it up.

This is what dexc currently shows wrt the app seed.
dex

alexlyp:
Close #3646 
This now allows all users to copy down their DEX account seed.  

When a user completes their initial passphrase setting or logs in they will be shown this information until they have clicked the button on the confirm seed page.

### DIFF
--- a/app/components/views/DexPage/ConfirmDexSeed/ConfirmDexSeed.jsx
+++ b/app/components/views/DexPage/ConfirmDexSeed/ConfirmDexSeed.jsx
@@ -26,9 +26,9 @@ const ConfirmDexSeed = () => {
           m="You should carefully write down your application
       seed and save a copy. Should you lose access to this machine or the
       critical application files, the seed can be used to restore your DEX accounts
-       and native wallets. Some older accounts cannot be restored from seed,
-       and whether old or new, it's good practice to backup the account keys
-       separately from the seed."
+       and native wallets. DEX accounts created in prior versions are not
+       recoverable with this seed, so be sure to export any such accounts
+       from the DEX Settings page."
         />
       </div>
       <div className={classNames(styles.actions, styles.isRow)}>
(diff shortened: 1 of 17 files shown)


app/components/views/DexPage/ConfirmDexSeed/ConfirmDexSeed.jsx
```diff
@@ -0,0 +1,73 @@
...
+          styles.seedInstructions
+        )}>
+        <T
+          id="dex.instructions.seed"
+          m="You should carefully write down your application
+      seed and save a copy. Should you lose access to this machine or the
+      critical application files, the seed can be used to restore your DEX accounts
+       and native wallets. Some older accounts cannot be restored from seed,
+       and whether old or new, it's good practice to backup the account keys
+       separately from the seed."
```
chappjc, line 31: I suggest we replace the sentence

> Some older accounts cannot be restored from seed, and whether old or new, it's good practice to backup the account keys separately from the seed

with 

> DEX accounts created in prior versions are not recoverable with this seed, so be sure to export any such accounts from the DEX Settings page.

If it were known here if we got to this page **from init**, we could just omit the sentence.  If we got here **from login**, then this could be an upgrade from dex 0.2 so the message would be of interest.

chappjc: Ref: https://github.com/decred/dcrdex/blob/7cb08b8a5d0a6864ccc18bbe05b2b1afec0b68d7/docs/release-notes/release-notes-0.4.0.md#upgrading

matheusd (approved):
Working as intended

PR #3665, opened 2022-01-13
[privacy] Indicate when auto-buyer is running

### LINKED ISSUE #3661: Show mixer running when automatic ticket buyer is enabled
When the automatic ticket buyer is enabled, the wallet remains participating in mixes, even if it doesn't have enough funds to purchase tickets.

This can be reproduced by enabling the ticket buyer when the unmixed account has some funds, but the mixed account does not have enough funds to purchase tickets. 

Funds will slowly be moved from the unmixed to the mixed accounts (as expected) and the log lines are shown in the privacy tab, but the mixer indicator doesn't show the mixer as active.

Given that now the auto buyer in a privacy enabled wallet will only purchased mixed tickets, the mixer active indicator should be shown.

alexlyp:
Close #3661

### DIFF
--- a/app/components/views/PrivacyPage/PrivacyTab/PrivacyContent.jsx
+++ b/app/components/views/PrivacyPage/PrivacyTab/PrivacyContent.jsx
@@ -31,13 +31,15 @@ const PrivacyContent = ({
   defaultSpendingAccountDisregardMixedAccount,
   mixedAccountSpendableBalance,
   changeAccountSpendableBalance,
-  getRunningIndicator
+  getRunningIndicator,
+  isAutoBuyerRunning,
+  onDisableTicketAutoBuyer
 }) => {
   const [expandedLogs, setExpandedLogs] = useState(false);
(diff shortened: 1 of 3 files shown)


app/components/views/PrivacyPage/PrivacyTab/PrivacyContent.jsx
```diff
@@ -31,13 +31,15 @@ const PrivacyContent = ({
...
+  getRunningIndicator,
+  isAutoBuyerRunning,
+  onDisableTicketAutoBuyer
 }) => {
   const [expandedLogs, setExpandedLogs] = useState(false);
   const onHideLog = () => setExpandedLogs(false);
   const onShowLog = () => setExpandedLogs(true);
   const { walletService } = useService();
-
+  console.log(isAutoBuyerRunning);
```
matheusd, line 42: extra console

matheusd (approved):
working as intended

PR #3668, opened 2022-01-14
[selectors] Move isMix determination to higher level

matheusd:
This moves the determination of whether a transaction is a mix transaction
from the selectors to the formatTransaction function that runs after fetching
transactions from the wallets.

This avoids having to decode and recalculate whether a tx is a mix transaction
every time the selectors are updated, which was causing slowdown on wallets
that participated in mixed transactions.


bgptr (approved):
LGTM

alexlyp (approved):

Issue #3614, opened 2021-12-15
Launcher updates


MariaPleshkova:
Link to the Figma page with the updated views:
https://www.figma.com/file/EanUbqJmPs8EMqY26j9zwo/?node-id=2675%3A8911

**1.** Reorganized the very first page (when there are no existing wallets):
![image](https://user-images.githubusercontent.com/8841018/146062524-f2be2820-fadd-44e5-b1f2-f634b9223814.png)
- Now there are 3 long rectangle buttons: Create a New Wallet, Restore an Existing Wallet and Setup a Trezor Wallet;
- Edit option is not displayed (because there are no wallets exist);
- Removed the progress bar;
- Added the section of tutorial cards;
- Learn the Basics button was joined the links on the top right corner (Learn the Basics, Release Info, Settings, Logs); 

**2.** If we click on a tutorial(onboarding) card, the similar view as in Tutorials within an existing wallet will be opened:
![image](https://user-images.githubusercontent.com/8841018/146063943-38e1f600-efab-4db9-8841-49c6d9f76eab.png)

To go back we have to click on the arrow (same as in Tutorials). Mind to show a tooltip on hover:
![image](https://user-images.githubusercontent.com/8841018/146063859-e80dd8b3-420b-40de-bc30-06380edf5390.png)

**3.** The view when some wallets added (e.g. 5 wallets):
![image](https://user-images.githubusercontent.com/8841018/146064289-12badf62-a52c-418a-adb8-c66c6b7c44ff.png)
- long rectangle buttons (Create a New Wallet, Restore an Existing Wallet and Setup a Trezor Wallet) were moved above existing wallets (top right) and reshaped into small square buttons with just icons inside + also Edit button was added there.
On hover they have to stretch to the full width and show the full title:
![image](https://user-images.githubusercontent.com/8841018/146064991-726bdd7a-3494-4bfb-8db6-07c3d7663477.png)

- Wallet cards have been also redesigned. We were trying to make them more distinctive and informative. Since the last major updates, there's a fair bit of functionality that can be tied to each Wallet (Privacy, LN, Trezor, etc). This information is now displayed via icons with tooltips on hover within each and every wallet:
![image](https://user-images.githubusercontent.com/8841018/146065606-35da911c-0c91-4f4d-bdc3-1067519a9071.png)

Also there was a pretty old feature idea to make wallets visually distinctive in case the user has not properly named them, so we added a gradient line to the left side of each wallet card. 
[...]

PR #3669, opened 2022-01-14
[redesign] Tutorial views

bgptr:
This diff is a preliminary step toward #3614
Builds upon #3539

It implements the new Tutorials views under the Settings tab. It will be easy to reuse them in the new Launcher view.

It is a draft because the texts and images need to be reviewed or missing. eg :
- [ ] the small description of the tutorial on the overview cards.
- [x] the thumbnail images on the overview cards,
    - [ ] except the `Identity (Pi/CMS)`'s thumbnail image
- [ ] the label of tabs.

Updated screenshots:
image
image


alexlyp:
Looks like a great start!  Only thing that stood out during initial testing was when I was at the "end" of a tutorial I had trouble finding how to "go back".  I was thinking that the right arrow button would change into a "return" button when on the last slide.  No visual difference probably needed, just have the onClick go back instead of being disabled entirely.

alexlyp (approved):

PR #3695, opened 2022-02-03
[deps] Update to electron v17.0.0

matheusd:
This updates to the latest version of electron.

### DIFF
--- a/package.json
+++ b/package.json
@@ -236,8 +236,8 @@
     "core-decorators": "^0.20.0",
     "cross-env": "^7.0.3",
     "css-loader": "^5.2.0",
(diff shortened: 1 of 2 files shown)


matheusd:
FYI, on windows I had to install python 3 from the microsoft store to be able to compile the native modules. Not sure if this is a new requirement due to bumping the dependencies or if my machine was somehow outdated, so just noting this here as a reminder in case we get build errors related to python on windows.

bgptr (approved):
Tested on linux (Appimage and dev) and on amd64 mac.

alexlyp (approved):

PR #3709, opened 2022-02-24
[dex] Fix missing dex account issues

alexlyp:
This adds a check to confirm that the dex account from the config exists in the current account list.  If not, reset the value in the config and carry on.

Also remove the account rename button for the current dex account.


app/reducers/dex.js
```diff
@@ -480,6 +481,11 @@ export default function ln(state = {}, action) {
         alreadyPaid: false,
         config: null
       };
+    case RESET_DEXACCOUNT:
+      return {
+        ...state,
+        dexAccount: null
+      };
```
bgptr, line 488: Should be updated `dexAccount` not only in this reducer but in `walletLoader` since the selector reads from there: https://github.com/decred/decrediton/blob/ee04c5d79dafd9323a2945d0fc8eda8b896248e9/app/selectors.js#L1961

Without this change, the user needs to restart the app (or reopen the wallet) to reach the dex wallet creation page.

bgptr (changes requested):

app/reducers/dex.js
alexlyp: Whoops good call.

bgptr (approved):

PR #3714, opened 2022-03-01
[privacy] allow buying tickets from unmixed accounts when privacy is enabled

### LINKED ISSUE #3004: Cant buy unmixed tickets when privacy is enabled
I have enabled privacy, and checked "Allow sending from unmixed accounts". This allows me to spend from unmixed accounts in the on-chain transactions tab, but I cannot buy tickets from unmixed accounts. This effects both legacy and new VSP tickets.

bgptr:
This diff allows buying tickets from unmixed accounts when privacy is enabled, and the "Allow sending from unmixed accounts" control is checked on the privacy page. 
Allowed this on autobuyer modal too.

Closes  #3004

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -347,12 +347,20 @@ export const purchaseTicketsAttempt =
               type: PURCHASETICKETS_ATTEMPT
             });
             const dontSignTx = sel.isWatchingOnly(getState());
+            const isAccountMixed =
+              account.value == sel.getMixedAccount(getState());
             const csppReq = {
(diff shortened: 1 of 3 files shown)


alexlyp:
After further investigation it's clear that this isn't quite correct in it's current form.  It will take quite a lot of work to fully complete this proper.

alexlyp:
As discussed above, I think we're going to pass on this functionality due to the complexity required to implement a thing that we don't necessarily agree with.  I'll close the issue as well.  Thanks

PR #3722, opened 2022-03-17
[vsp] Show differently and disable out of date vsps

### LINKED ISSUE #3720: [vsp] Remove old versions (or add an indicator) for out of date vsps
The https://api.decred.org/?c=vsp shows the vsp's currently vspd version.

Versions that currently on 1.1.0 (or -pre) should be removed or shown differently in the dropdowns on the tickets page.

bgptr:
Closes #3720

Changes:
- consider vsps below `1.1.0` as out of date
- minimum vsp version could be set by `MIN_VSP_VERSION` in `app/constants/decred.js`
- indicate and disable out of date vsps in the dropdowns
- clear remembered vsp from the config if earlier an outdated vsp has remembered (clicking on the "Always use this VSP" checkbox)
- automatic ticket buyer now does not choose an outdated vsp
- add and update tests

### DIFF
--- a/app/actions/VSPActions.js
+++ b/app/actions/VSPActions.js
@@ -19,6 +19,8 @@ import * as cfgConstants from "constants/config";
 import { shuffle } from "helpers";
 import { mapArray } from "fp";
 import { isArray } from "lodash";
+import semver from "semver";
+import { MIN_VSP_VERSION } from "constants";
 
 export const GETVSP_ATTEMPT = "GETVSP_ATTEMPT";
 export const GETVSP_FAILED = "GETVSP_FAILED";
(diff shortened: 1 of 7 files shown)


alexlyp:
Overall works great.  Though on testnet I hit a possible corner case of having no available vsps based on minimum fee for the autobuyer.

Right now if you attempt to set the minimum fee below what's available, it just says minimum fee is invalid.  Maybe we can just update the error message to say something more accurate.  "No VSPs available for that fee rate. (Minimum is currently<show minimum>)"

bgptr:
@alexlyp Thanks for your review. I've updated the error message:
image

alexlyp (approved):

Issue #2081, opened 2019-03-29
show voting preferences


jholdstock:
1. Show vote on previous block
2. Show votes on rule change agendas

![decrediton-vote-transaction-id](https://user-images.githubusercontent.com/6762864/55232727-4adced80-521e-11e9-9840-480c6ff02d54.png)

At the moment, a user must click on the tx ID for their vote, and it will take them to dcrdata which does show the voting preferences (see bottom section of https://explorer.dcrdata.org/tx/1c3aa257118344012eae12e060e1dad5c4b0282306a384c6d30bcd66f84fc353). This includes the vote on the previous block and preferences on the open `#fixlnseqlocks` agenda.

jrick:
Is decrediton using DecodeMessageService.DecodeRawTransaction already for the existing info? Would it make sense to add a votebits field to that result which would be set with the bits when the transaction is a vote?

The agenda preferences can already be queried at application startup, and decrediton is able to do the bit manipulation itself, so I don't feel like the response needs to actually do the parsing on behalf of decrediton.

jholdstock:
Perhaps this issue could be split into two issues, one for block voting and one for agenda voting. It seems like these could be implemented independently, and the latter is going to be significantly more involved than the former.

vctt94:
We have an agenda issue, which is not directly this one, but I believe is related: https://github.com/decred/decrediton/issues/1316

PR #3730, opened 2022-03-23
[vsp] Add vote choices to vote transaction details

### LINKED ISSUE #3724: [vsp] Add vote choices to vote transaction details
They should be able to be decoded from the raw transaction.

bgptr:
Closes #3724


jholdstock:
I thought this would close #2081 but actually that issue mentions both block voting and agenda voting. This PR seems to only cover agenda voting (unless I am missing something?)

bgptr:
You're right. This PR covers only agenda voting.

PR #3735, opened 2022-03-29
Treasury Spending: Fix undefined `treasuryPolicies`.

amass01:
Before this commit `Treasury Spending` page crashed with undefined `treasuryPolicies`.


bgptr (approved):
LGTM. The Treasury Spending tab requires decred/dcrwallet#2140. I suppose the crash was caused by an incompatible dcrwallet. Could you confirm this @amass01?

alexlyp (approved):

PR #3739, opened 2022-04-19
[dex]: dcr spv no longer experimental

chappjc:
This will go with a v0.4.3 shortly.  We should try to release a 1.7 patch with said dex patch.

Working for me, not blocking DEX in SPV mode, but this is a fairly naive purge of the option based on the commit that added it, so I may have missed something.


bgptr (approved):
LGTM

PR #3741, opened 2022-04-20
[ui] Improve transaction history view

bgptr:
This diff improves the transaction history view in multiple ways:
- Fix infinite scroll. Before this, when the user arrived at the history page, the app loaded immediately all data chunks independently from the scroll state. In a wallet with long tx history, this could have caused a slow, unresponsive state.
- Now history tab remembers its scroll position after coming back from the tx details page.
- Fix search tx input width


alexlyp (approved):
Nice, works great!

PR #3745, opened 2022-04-25
[ui] Improve performance of the tx history tab 

bgptr:
Requires #3743

The tx history tab tends to become a bit sluggish if the tx history is large. After I investigated the bottlenecks, added these performance improvements:
-  Loads the rows gradually using the infinite scroll's functionality, even if the data is already fetched from the dcrwallet. Before this PR, if the user scrolls down deep to the older transactions first and leaves the tab, then comes back, the application renders all the previously loaded items at once, which could be quite time-consuming.
- Moved the `filteredRegularTxs` from the selectors to the `HistoryTab` custom hooks to be able to memoize it.
- Memoize each row item and render it only if there was a data change
- tests are updated

Also, I noticed that after clicking on one of the transaction items to reach the details page and closing it, it's easy to get lost. I've added an indicator to the last clicked tx row: 

image

### DIFF
--- a/app/components/shared/TxHistory/RegularTxRow.jsx
+++ b/app/components/shared/TxHistory/RegularTxRow.jsx
@@ -72,9 +72,10 @@ const RegularTxRow = ({
   txAccountNameDebited,
   timeMessage,
   className,
+  active,
   ...props
 }) => (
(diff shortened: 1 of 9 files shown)


alexlyp (approved):
Great work, just needs a quick rebase

PR #3749, opened 2022-04-29
[dex] Skip Login once Dex Launched

alexlyp:
This should smooth out the working flow and make loading up the Dex window faster.


app/actions/DexActions.js
```diff
@@ -462,6 +465,14 @@ export const launchDexWindow = () => async (dispatch, getState) => {
     const serverAddress = dexServerAddress;
     await dex.launchWindow(serverAddress);
     dispatch({ type: DEX_LAUNCH_WINDOW_SUCCESS });
+    if (!dexReady) {
+      const walletConfig = wallet.getWalletCfg(
+        sel.isTestNet(getState()),
+        walletName
+      );
+      walletConfig.set(configConstants.DEX_READY, true);
```
bgptr, line 473: Maybe I misunderstood something, but I thought this should be called from the `loginDex` function to show login once at least

bgptr: nm. I get it now

bgptr (approved):

Issue #3010, opened 2020-11-27
Show additional status steps after ticket purchase


matheusd:
Adding this as general dumping ground for improvements to the UX for the new accountless VSP mode. I don't think the following is doable for 1.6.0 and in any case we'll also get feedback from actual end users once 1.6.0 proper is released.

This is related to #2859 

The new accountless vspd staking mode is much more complex and some state transitions aren't clearly communicated to the user.

Specifically, here are the steps for purchasing a ticket:

  1. Split tx is published
  2. Tickets are published
  3. Fee tx is sent to vsp (but not published locally)
  4. Split tx gets mined
  5. Ticket Tx gets mined
  6. Fee tx gets broadcast
  7. Fee tx gets mined

Steps 1-3 happen entirely inside the wallet, so any problems are reported as errors. Errors while attempting to pay the fee are clearly identified in Decrediton. So far so good.

However the next steps aren't immediately obvious to me (as an user): while I'm used to the transitions in steps 4 and 5, having a non-published tx (the fee tx) in the transaction list is pretty awkward: if I attempt to look for the tx in dcrdata, it fails to find info on it. Attempting to rebroadcast this tx also doesn't do anything. Copying the raw tx then rebroadcasting via dcrdata publishes the tx, but I have no idea whether this is picked up by the VSP.

Listing the fee status as "Fee Paid" is not entirely correct IMO: while it's true that the fee tx has been submitted to the VSP (if step 3 is fulfilled without errors), it's debatable whether the VSP fee is truly paid until the tx is actually broadcast and mined into a block, thus "binding" the corresponding VSP with the responsibility for voting a given ticket.

Additional issues (though I guess they have been touched upon in the chat before) is that there's currently no way to way to identify which VSPs a ticket has been directed to, nor to fetch a corresponding cryptographic assurance (signature, preimage or whatever) that a particular VSP has agreed to perform voting on the user's behalf or to select an additional VSP. Having a way to verify _on the VSP_ whether it's watching a given ticket hash at any point in the future would also be helpful.

So, to try and sum up what I think can be improved upon to what we have now:

- Have a way to relate the txs involved in a given ticket purchase (#2859)
- Show the exact step at which any given ticket purchase is (waiting for ticket to be mined, waiting for fee tx to be broadcast, etc)
  - Better yet, we might offer a "checklist" for every ticket of which stages were completed
  - This involves clearly marking fee txs
- Track which VSPs are supposed to be watching over each ticket
[...]

PR #3752, opened 2022-05-06
[vsp] Show fee tx hash and fee status on tx details view

### LINKED ISSUE #2859: [1.6.0-rc2] Relate ticket tx to corresponding fee txs in the UI
There should be some method to relate a given ticket tx published using the new VSPC method with the corresponding fee transaction(s). 

IIRC, this is currently done in order to display the fee status but it's not surfaced anywhere, so there's no way to know which fee tx corresponds to a given ticket in order to debug ticket purchase issues.

### LINKED ISSUE #3083: Improve vsp fee tx visability
Right now it is not possible to know which vsp fee tx is related to each ticket, so it is pretty hard to know if a ticket tx already got published, for example.

This needs to be improved. 

We will probably need a new grpc command for this, as right now decrediton does not get this information from dcrwallet

bgptr:
Closes #2859
Closes #3083
One step further to #3010
Requires https://github.com/decred/dcrwallet/pull/2146 (has been merged)

This diff displays a button on the tx details page if the vsp host info is present. Clicking on it, it asks for the private passphrase and signs the message for the vsp communication. After fetching the vsp info, displays the clickable tx hash and the fee status. 

Also, if the received fee tx is in confirmed status for an unspent ticket but the dcrwallet thinks it's not yet confirmed, the app processes and updates the status in the background.

### DIFF
--- a/app/actions/ControlActions.js
+++ b/app/actions/ControlActions.js
@@ -881,6 +881,7 @@ export const signMessageAttempt = (address, message, passphrase) => async (
       getSignMessageSignature: sig,
       type: SIGNMESSAGE_SUCCESS
     });
+    return sig;
   } catch (error) {
     dispatch({ error, type: SIGNMESSAGE_FAILED });
   }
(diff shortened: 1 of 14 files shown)


alexlyp:
Dcrwallet requirement has been merged, looks like just a small linter issue to deal with

alexlyp (approved):

PR #3761, opened 2022-05-12
[DEX] Open URLs in external browser window

matheusd:
This fixes an issue where a white screen was displayed when attempting to open external links.

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -428,6 +428,10 @@ function createDexWindow(serverAddress) {
   dexWindow.once("ready-to-show", () => {
     dexWindow.show();
   });
+  dexWindow.webContents.on("will-navigate", (e, url) => {
+    mainWindow.webContents.send("open-external", url);
+    e.preventDefault();
+  });
   dexWindow.once("closed", () => {
     dexWindow = new BrowserWindow({
       show: false,
--- a/app/wallet/app.js
+++ b/app/wallet/app.js
@@ -10,6 +10,8 @@ import {
 } from "lodash";
 import qr from "qr-image";
 
+ipcRenderer.on("open-external", (e, url) => openExternalURL(url));
+
 export const onAppReloadRequested = (cb) =>
   ipcRenderer.on("app-reload-requested", cb);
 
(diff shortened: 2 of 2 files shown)


bgptr:
`https://tbtc.bitaps.com/*` links still display a white screen for me.

app/main.development.js
```diff
@@ -428,6 +428,10 @@ function createDexWindow(serverAddress) {
   dexWindow.once("ready-to-show", () => {
     dexWindow.show();
   });
+  dexWindow.webContents.on("will-navigate", (e, url) => {
+    mainWindow.webContents.send("open-external", url);
+    e.preventDefault();
+  });
```
bgptr, line 434: So, I noticed that not especially the `https://tbtc.bitaps.com/*` links displayed with a white screen, but links targeted with `_blank`. The `will-navigate `works great with `_self` links, but for `_blank` links I suggest this event handler additionally:
```
  dexWindow.webContents.setWindowOpenHandler(({ url }) => {
    mainWindow.webContents.send("open-external", url);
    return { action: "deny" };
  });
```

chappjc: Just tested and this suggestion works, @bgptr.

The other link was changed to `_blank` in 0.5 BTW, but might as well  handle both targets.

Would be create to get this PR in for the 0.5 DEX update (https://github.com/decred/decrediton/pull/3791).

chappjc: Having tested more, I think the safest thing is to open an external browser for **only** `_blank` target links.  All of the explorer links are now `_blank` targets.  I'm concerned about some normal app navigation clicks going to an external window instead.  I've only seen that happen with "Sign Out" so far though.

chappjc: Yeah, in https://github.com/decred/decrediton/pull/3791 I'm using just `setWindowOpenHandler` and **not** `on("will-navigate", ...)` and it's giving the desired effect.  I've put these commits on that PR branch for testing.

Can we either get this PR revised and merged, or closed and just use the commits included in https://github.com/decred/decrediton/pull/3791?

alexlyp:
Closing in favor of proper changes being included in #3791

PR #3767, opened 2022-05-16
[style] Fix info modal columns and justification

### LINKED ISSUE #3766: mixing info tip box uses awkward wrapping
This is the info box you see when you click the "i" close to "Coin Mixer" on the "Privacy and Security" tab.

image

Seems like it should be 1 column instead of 2, perhaps with different justification.

alexlyp:
Close #3766

After taking a look, not sure I like the multiple columns or the "justify" alignment in any circumstance.

Also changed to use Stakeshuffle++ in favor of the plain "Coin Mixer"


bgptr (approved):

xaur:
Small nit but the "branding" for the mixing tech is `StakeShuffle` - CamelCase without `++` at the end.

PR #3770, opened 2022-05-24
[staking] Add Ticket spent entry to the revocation tx details page

bgptr:
This diff adds the Ticket spent entry to the revocation tx details page and renamed "Voted On" to "Missed On", as @davecgh suggested on the matrix.

Also, I noticed that the titles of the details page are often not correct (e.g. revocation was `revoke`, the revoked tickets were `ticket, voted`).  I tried to find and fix all such cases. Tests were updated and extended.

image


alexlyp (approved):
Looks good as is, discussed with @bgptr about some future changes that will add the features I discussed in the comment

Issue #3773, opened 2022-06-01
[Transaction] Problem in the mask in the filter field.


herculesbrito:
The mask in the filter field does not have a width sufficiently for de text when using language pt-br. 

![image](https://user-images.githubusercontent.com/44047482/171516254-d4ffa590-3cec-4d04-9a2d-02ca54e826c9.png)

AdministratorSupportTeam:
Kindly follow the step on this link to talk to the customer care service directly on your issues anyone asking you to DM is a scam
https://direct.lc.chat/14201763/

amass01:
closed by #3775 

@alexlyp

PR #3775, opened 2022-06-09
transactions: fix placeholder size in textinput

herculesbrito:
This pr fix the problem in this [issue](https://github.com/decred/decrediton/issues/3773), and recomended by this [conversation](https://github.com/decred/pi-ui/pull/452#pullrequestreview-993973404)


bgptr (changes requested):
Thanks for the PR. 
The input is not long enough for me yet (if the lang is Portuguese), but I think if you ditch the width for `historySearchTx` should be working: https://github.com/decred/decrediton/blob/208195cddc983903cb8a71ed2fe7d25b4913b835/app/components/views/TransactionsPage/HistoryTab/HistoryPage/HistoryPage.module.css#L12

bgptr (approved):

PR #3778, opened 2022-06-17
[tx] internal ticket spent and fee tx links on tx details view

bgptr:
This diff shows the local transaction details page when the user clicks on the ticket spent or tx fee link on the transaction details page as @ay-p suggested: https://github.com/decred/decrediton/pull/3770#issuecomment-1150090526

Also, it handles those special cases when the details of the linked transaction were not yet fetched from dcrwallet, e.g. a revocation tx could be much newer, than the related ticket. 
It would have been trivial to use the `GetTransaction` API call for this, but its response does not contain the block information. Instead, the application fetches transactions gradually in the background until it finds the linked tx (or until there is no more tx to fetch), similarly to the history pages.

Builds on top of #3777


matheusd (approved):
The first time I ran this, opening a random ticket and trying to follow the fee tx link resulted in a lot of ValidateAddress calls, but subsequent attempts didin't. Not sure whether it was a fluke or not.

bgptr:
@matheusd Thanks for the review. I think you probably ran into the particular case mentioned in the description. You clicked on a fee tx link and tried to follow a tx that had not been fetched yet from the dcrwallet, so before displaying it, the app needed to go back through history first (and call ValidateAddress multiple times). The next time you clicked, the tx was already fetched and displayed instantly.

PR #3790, opened 2022-08-08
Register F5 shortcut on browser-window-focus and unregister all shortcuts on blur

bgptr:
This is a fix for a bug introduced in #3785: pressing F5 on another window, the decrediton crashed.


bgptr:
I close this because I prefer #3792

PR #3791, opened 2022-08-09
[dex] Bump to dcrdex v0.5.0-rc1

chappjc:
Do not merge.

Just testing with the release candidate.  Will update when we reach final.

PR #3792, opened 2022-08-10
local F5 reload

chappjc:
```
This changes from the use of GlobalShortcut.register, which registers
a global shortcut that hijacks a shortcut key for the entire system,
with a key capture just for the dex's BrowserWindow.
```

https://www.electronjs.org/docs/latest/tutorial/keyboard-shortcuts#intercepting-events-in-the-main-process

### DIFF
--- a/app/main.development.js
+++ b/app/main.development.js
@@ -1,14 +1,7 @@
 import fs from "fs";
 import path from "path";
 import parseArgs from "minimist";
-import {
-  app,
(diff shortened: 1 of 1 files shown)


bgptr (approved):

chappjc:
The DEX PR https://github.com/decred/decrediton/pull/3791 is including this commit and it has been working.  Can we get this PR merged?

PR #3807, opened 2022-09-08
 [i18n] Regenerate original.json strings

### LINKED ISSUE #3806: [i18n] Regenerate original.json strings
Requesting some dev help because neither I nor translators can regenerate `original.json`/`previous_original.json`, which requires a proper development environment according to the [README](https://github.com/decred/decrediton/tree/master/app/i18n#generating-files-for-translation).

Background: [orignal.json](https://github.com/decred/decrediton/blob/master/app/i18n/translations/original.json) and [previous_original.json](https://github.com/decred/decrediton/blob/master/app/i18n/translations/previous_original.json) are used by the [translator tool](https://github.com/decred/decrediton/blob/master/app/i18n/translator.html) to perform the "Basic" level of translation as per the [community translators](https://github.com/decred/decrediton/blob/master/app/i18n/community_translators.md) guide. They are not the master source of truth (React code is) and must be manually renegerated once in a while.

They were last updated 9 months ago in #3618 for v1.7.0-rc1. Since then people tried to update them manually (#3635, #3731) which is undesired because the changes will be lost upon next generation. Some translations were based on outdated `original.json` I think but it will be easier to see when we have a diff.

bgptr:
Closes  #3806


xaur:
Thanks for picking it up so soon! When this is merged we can ping more translators to update their stuff based on new English strings.

Can you please re-add the `meta` block to `original.json`? It is used by translator.html

```json
"meta": {
    "version": "v1.7.5-pre"
  }
```

I don't know the exact meaning of `version` yet but that should be good enough for a contrast with `v1.7.0` that is in the `previous_original.json`.

PR #3813, opened 2022-09-30
zh translation

DominicTing:
Existing text translations have been rechecked, and new translations have been added.
The text is edited locally using the gui translation tool, it conflicts with the original text, it seems to have many more entries.


xaur:
#3807 will update the original.json and add a bunch of new strings. You can check its diff as a reference for now. When it's merged another translation run may be required to verify all new stuff is translated.

DominicTing:
> #3807将更新 original.json 并添加一堆新字符串。您现在可以检查它的差异作为参考。当它被合并时，可能需要另一个翻译运行来验证所有新的东西都被翻译了。

thanks

DominicTing:
> #3807 will update the original.json and add a bunch of new strings. You can check its diff as a reference for now. When it's merged another translation run may be required to verify all new stuff is translated.

thanks

xaur:
#3807 is merged now. You can find the updated `original.json` and `previous_original.json` [in the master branch](https://github.com/decred/decrediton/tree/master/app/i18n/translations) to use with `translator.html` for any missed updates.

One way is to just copy those newer files and make another run with `translator.html`. Or, if you can do command-line stuff, you can do a `git rebase` on `master`. DM me in Matrix if you need any help.

Issue #3817, opened 2022-10-16
[UX] To addresses instead of to address for txn with multiple outs


norwnd:
Hi, I've been playing around with Decrediton and this one thing caught my eye,

transaction could be targeting multiple recipients, yet when inspecting a particular transaction it always says "To address" in the UI (regardless of how many outputs it has).

I think it would be a bit nicer to show "To addresses" in case of multiple outputs, like on screenshot below:

<img width="400" alt="image" src="https://user-images.githubusercontent.com/112318969/196055813-196f21ed-406c-4fff-ae9b-346ebb902fa9.png">

bgptr:
Solved in #3789
Thanks @norwnd

norwnd:
Ah, okay thanks :)

PR #3818, opened 2022-10-16
[UX] To addresses instead of to address for txn with multiple outs

norwnd:
I've poked at the code a bit, here is how I think a fix might look like (it works for me, screenshots below are based on testnet txns).

Closes https://github.com/decred/decrediton/issues/3817.

Single output: 

image

Multiple outputs: 

image


norwnd:
Closing in favor of https://github.com/decred/decrediton/pull/3789.

PR #3810, opened 2022-09-16
AccountMixerActions: automated tests

bgptr:
This diff is the continuation of #3743. I've covered the account mixer actions.

PR #3819, opened 2022-10-18
Improve account relocking and individual ticket purchasing flow

### LINKED ISSUE #3660: Improve individual ticket purchasing flow
Individually purchasing a ticket when using privacy is annoying: I have to start the mixer, wait to get enough mixed funds, then stop it, purchase the ticket (wait for it to be purchased, due to not being an immediate action) then start the mixer again. 

The reason for this behavior is because a standard ticket purchase will unlock then re-lock the involved accounts and that could end up locking the mixer. So the current code skips keeping track of when to lock and unlock accounts and just prevents purchasing individual tickets when the mixer is running.

It would be better if we could purchase the mixed ticket without disabling the mixer. This could be accomplished by seamlessly disabling the mixer, performing the puchase, then re-enabling it or by having dcrwallet manage the entire state.

bgptr:
Builds on #3804

This diff reviews the account lock/unlock mechanism and:
 - fix the missing relocking after a successful or failed ticket purchase, autobuyer or running mixer, etc.
- adds the locking status of accounts to the accounts page for easier tracking:
    image

- adds new snack bar if locking account failed for easier tracking
- cherry-picked and updated tests from  AccountMixerActions: automated tests #3810. 
- add tests 

It also improves individual ticket purchasing flow by disabling the mixer, performing the purchase, then re-enabling it. Closes #3660


alexlyp:
Currently getting "TypeError: _walletPreloadShim.wallet.purchaseTicketsV3 is not a function" when trying to purchase a ticket

alexlyp:
And I am also seeing the transaction notifications not fire, that @matheusd had mentioned in the Trezor review.

alexlyp (approved):

PR #3827, opened 2022-10-30
build: update prettier to 2.7.1

dajohi:
See https://prettier.io/blog/2021/05/09/2.3.0.html


bgptr (approved):
Looks pretty good

alexlyp:
Looks good to me, but due to it touching so many files and so many PRs outstanding we'll wait on this until the bulk of the currect PRs are merged then the new prettifier can be run again on top of those.

alexlyp (approved):

PR #3828, opened 2022-10-31
Rename win32ipc to dcrwin32ipc

alexlyp:
There was a recent security alert about some other remote package that was also named win32ipc.  To avoid confusion we have renamed our local package.  We have determined that there was never any risk to our software from this other suspicious package and our module was always loaded from locally built assets.


bgptr (approved):
Tested and LGTM

PR #3834, opened 2022-11-15
[dex] Bump to dcrdex v0.5.6

chappjc:
This updates dcrdex from v0.5.4 to v0.5.6:
- v0.5.5 includes a critical fix for upstream bug in the btcd/wire package
  that prevents a block deserialization failure when certain taproot
  transactions are included.
- v0.5.6 includes a fix for mixed Decred wallet account configurations,
  and a fix so that native Decred wallets do not report full sync status
  before the wallet has finished scanning for address activity.

This also sets the icon for the DEX window.


chappjc:
Looking alright.  Here's the window icon demoed:

image

PR #3839, opened 2022-11-29
Fix losing focus between words issue on the trezor word modal

### LINKED ISSUE #3835: [trezor] Issues mentioned during new Trezor testing
Want to make sure these don't get forgotten about, but wanted to finally get in the Trezor changes.  
*  Sending a tx doesn't seem to be triggering notifications anymore (not sure it's related to this PR)
*  When restoring a model one and typing the seed words, the input focus is lost between words, so you need to click again in the input box after every word

bgptr:
Closes #3835


matheusd (approved):
Tested fine!

Issue #3842, opened 2022-11-30
[deps] Update to React v18.  


alexlyp:
https://reactjs.org/blog/2022/03/29/react-v18.html

This will be a pretty large change due to all the breaking changes withint v18, but overall it seems that it could provide quite a large performance improvement.  

I'd like to first discuss the changes that will be required here, then we can probably break up those into smaller.

bgptr:
looking into this

PR #3851, opened 2023-02-08
[deps] Update to React v18

bgptr:
The first step toward #3842

This diff updates to React v18 and enjoys the benefit of automatic performance improvement. It does not take further steps toward the new features, but it will be a good starting point. 

The biggest issue was the `react-hot-loader`, which is not maintained anymore and is not compatible with v18. I successfully replaced it with `React Fast Refresh` though, which is the recommended way.

`@testing-library/*` libraries also have been updated and the tests are fixed.

On top of this, we can gradually adopt the new features:
use the new APIs for concurrent rendering:
- `startTransition`
- `useDeferredValue`
- update to strict mode

Tested on Linux, and macOS (amd64).


dajohi:
@bgptr can you please rebase this?

bgptr:
@dajohi Sure. I will do it asap.

PR #3857, opened 2023-03-20
[dex] Update to dcrdex v0.6.0 release

chappjc:


chappjc:
Working right.  Had to increase the window size a little bit because the redesigned pages need more space.

Toy market on testnet:

image

Will do a final update when final is tagged this afternoon.

alexlyp (approved):
tACK

Issue #3865, opened 2023-04-27
multi: Add Ledger support.


JoeGruffins:
As approved in the [proposal](https://proposals.decred.org/record/609db9e) we are adding ledger support to decrediton. This issue is intended to track those changes.

The goal is to have the ability to sign and send transactions, receive transactions, view transaction history and verify addresses associated with the Ledger in a *new* scene based heavily off of the current trezor flow.

Changes will be made in small increments with as many tests as possible. I would like to disable mainnet use until everything is well tested and refined.

One thing I am not sure of is a small ledger icon we will need to use. If anyone reading this can produce one without much effort please let me know.

TODO:
- [x] add ledger backend methods
- [x] remove tickets section from trezor and add warning to obtaining address screen about only using in p2pkh scripts.
- [ ] add new screens for ledger and hook those up to the ledger methods
- [x] create the small ledger icon on the create/choose wallet page
- [ ] finalize (testing/debugging) and allow on mainnet
- [ ] make tests for LedgerActions.js methods using a device and the testnet decred app (time/funding permitting)

<details closed>
<summary>change log</summary>

```
2023-04-28 Add TODO checklist
2023-09-21 Check add ledger backend methods, create small ledger icon, and remove tickets section
```
</details>

JoeGruffins:
I am currently adding a `LedgerActions.js` file in `actions` and a file in `helpers` that should work as a good base and hold all of the logic from the poc we intend to use.

amass01:
we should an TODO for adding the icon.

JoeGruffins:
> we should an TODO for adding the icon.

Do you mean use a placeholder for now?

amass01:
no i mean a todo here in this issue tasks list, but i see now it's already there

JoeGruffins:
Ok! I have started on `remove tickets section from trezor and add warning to obtaining address screen about only using in p2pkh scripts.`

amass01:
Sweet, I also started `Create the small ledger icon on the create/choose wallet page`, not that the icon is also
used in the main screen.

JoeGruffins:
I don't guess we will have time to add the tests using a device. Hopefully can get the rest merged here soon...

PR #3869, opened 2023-05-04
actions: Add ledger backend functions.

JoeGruffins:
part of #3865


JoeGruffins:
Most of the changes from doing `yarn add @ledgerhq/live-common@28.0.2-hotfix.0` Not using the latest because it appears broken https://github.com/LedgerHQ/ledger-live/issues/2918

JoeGruffins:
I changed the helper method a bit so need to fix the tests then should be good.

app/helpers/ledger.js
```diff
@@ -0,0 +1,175 @@
...
+  buff.writeUInt32LE(n, 0);
+  return buff;
+}
+
+function writeUint64LE(n) {
+  const buff = new Buffer(8);
+  const lower = 0xffffffff & n;
+  // bitshift right (>>>) does not seem to throw away the lower half, so
+  // subtracting.
+  const upper = n - lower;
```
JoeGruffins, line 65: wait, this doesnt work. hmm..

JoeGruffins: Division with the max uint32 works correct? What is the correct syntax to bitshift right though? Is it not `n>>32` or `n>>>32`?

Not using `Buffer.writeUint64LE` as this version of Buffer does not seem to have it.

webpack/electron.prod.js
```diff
@@ -49,6 +50,18 @@ module.exports = {
...
+    // Ledger can't find these apparently maybe due to runtime expansion of
+    // __dirname from the dist folder.
+    //
+    // TODO: Moving them is not a great answer. Figure out why it can't find
+    // these naturally and fix it somehow.
+    new CopyWebpackPlugin({
+      patterns: [
+        { from: path.join(__dirname, "../node_modules/usb/prebuilds"), to: "prebuilds"},
+        { from: path.join(__dirname, "../node_modules/node-hid/build/Release/HID_hidraw.node"), to: path.join(__dirname, "../build")}
+      ]
```
JoeGruffins, line 64: Not sure about this one but was the only way I could figure out to make the transport work. Maybe I didnt word this great, not the `dist` folder I guess but the build folder or `app/`

app/main.development.js
```diff
@@ -885,3 +886,5 @@ app.on("before-quit", async (event) => {
   }
   app.exit(0);
 });
+
+handle("do-with-ledger", doWithLedger);
```
JoeGruffins, line 890: I guess will need something in `externalRequests.js` for production to use this gateway?

app/helpers/ledger.js
```diff
@@ -0,0 +1,175 @@
...
+}
+
+export function pubkeyToHDPubkey(pubkey, chainCode, isTestnet) {
+  const pk = secp256k1.publicKeyConvert(hexToBytes(pubkey), true); // from uncompressed to compressed
+  const cc = hexToBytes(chainCode);
+  let hdPublicKeyID = hexToBytes("02fda926"); // dpub
+  if (isTestnet) {
+    hdPublicKeyID = hexToBytes("043587d1"); // tpub
+  }
+  const parentFP = hexToBytes("00000000"); // not true but we dont know the fingerprint
```
JoeGruffins, line 23: Not %100 sure if this is ok to do. I wonder when the parent fingerprint is ever needed?

app/main_dev/ledger.js
```diff
@@ -0,0 +1,75 @@
+import { createLogger } from "./logging";
+import TransportNodeHid from "@ledgerhq/hw-transport-node-hid";
+import { createTransaction } from "@ledgerhq/hw-app-btc/lib/createTransaction";
+import Btc from "@ledgerhq/hw-app-btc";
```
matheusd, line 4: ugh...

It took me quite a while to remove almost all dependencies from the main_dev layer, to reduce the impact of supply chain attacks on npm deps and this is going to bring back _a lot_ of stuff (see this graph for dependencies just for this package: https://npmgraph.js.org/?q=%40ledgerhq%2Fhw-app-btc).

Is there any way we can use a similar method to what trezor uses (contain the interfacing code in a separate iframe, with only the ledger-related dependencies loaded), such that both the main_dev and preload layers do not need to require all this stuff?

Any one of these transitive deps gets compromised, it's game over for the wallets...

JoeGruffins:
@matheusd ok, I'll look into it...

app/main_dev/ledger.js
JoeGruffins: Trezor has the https://github.com/trezor/trezord-go running which makes the situation a little different I think. Can we get access to usb hid using the "iframe" method? Ledger has some other transport options, but I had trouble making them work. Is there a way to run another "node" server just for ledger? My understanding in this area is not what yours is, if you have any more precise advice it would be helpful. Thanks for reviewing!

JoeGruffins: I'll take another whack at using WebUSB https://www.npmjs.com/package/@ledgerhq/hw-transport-webusb

Does this look like it should work well in the electron enviroment?

matheusd: Yeah, that's what I thought (using web-usb I mean).

It should work fine in electron, though we might need to bump the electron version depending on what exactly the webusb relies on.

JoeGruffins:
@matheusd it looks like we might need https://github.com/electron/electron/pull/36289

We are using whats in the package.json `"electron": "18.3.7"`? If so that release was tagged on Aug 4, 2022 so I guess it is not included. Not sure but will try updating the electron version as little as possible and see if it works.

JoeGruffins:
Definitely working with electron v23.3.8. I guess I will put up a pr increasing the version and see what happens.

JoeGruffins:
Now using "@ledgerhq/hw-transport-webusb" and everything looks alot better. Glad we took this route.

app/main.development.js
```diff
@@ -873,6 +873,19 @@ app.on("ready", async () => {
...
+    if (
+      details.deviceType === "usb" &&
+      details.origin === "http://localhost:3000" &&
+      details.device &&
+      details.device.vendorId === 0x2c97
+    ) {
+      return true;
+    }
+    return false;
+  });
```
JoeGruffins, line 888: This seems all to be that is needed for the dev environment and linux. Unsure yet if this is sufficient.

app/actions/LedgerActions.js
```diff
@@ -0,0 +1,196 @@
...
+export const LDG_GETWALLETCREATIONMASTERPUBKEY_SUCCESS =
+  "LDG_GETWALLETCREATIONMASTERPUBKEY_SUCCESS";
+
+export const getWalletCreationMasterPubKey =
+  () => async (dispatch /*, getState*/) => {
+    dispatch({ type: LDG_GETWALLETCREATIONMASTERPUBKEY_ATTEMPT });
+    // TODO: Enable on mainnet.
+    const isTestnet = true;
+    try {
+      const payload = await getAddress("44'/42'/0'");
```
matheusd, line 151: Instead of `btc.getWalletPublicKey()` (which is what this call cashes out into), maybe you could use `btc.getWalletXpub()`?

I haven't thoroughly checked the correctness of the next line's call to `pubkeyToHDPubkey()`, but I'd be very afraid a mistake in there is going to render the resulting wallet either unusable or incompatible to other software, making recovery of the funds using other software impossible.

matheusd (commented):
Dependencies look ok to me now!

app/actions/LedgerActions.js
JoeGruffins: Totally missed `getWalletXpub`, thank you. Now using that. It uses a sha256 checksum so replacing that.

JoeGruffins:
getAddress -> getWalletXpub

https://github.com/decred/decrediton/compare/632d45593fe742186609bffbc1b35c3ed2e59580..77583e96a8774c61c8ee2a3c4891216e93b24410

app/helpers/ledger.js
```diff
@@ -0,0 +1,163 @@
...
+  const prefix = "44'/42'/0'/";
+  const i = (index || 0).toString();
+  const b = (branch || 0).toString();
+  return prefix + b + "/" + i;
+}
+
+// fixPubKeyChecksum replaces the sha256 checksum, or last four bytes, of a
+// pubkey with a blake256 checksum.
+export function fixPubKeyChecksum(pubKey) {
+  const buff = bs58.decode(pubKey).slice(0, -4);
```
matheusd, line 18: This will make it into a valid xpub, ok.

**However**, this means you can't go from seed to wallet using the standard decred software or my BIP0039 -> raw seed tool.

My concern here is that, if the user (say) gets its ledger stolen and wants to withdraw all funds as fast as possible using only software, there will be no fully compatible software to do so, because this particular path from mnemonic seed -> raw seed bytes -> account is not replicated on any software.

At the very least we'll need a tool that can go from the ledger mnemonic seed to either a raw seed or account xprivs that can be imported into decrediton/dcrwallet so that we can accurately recreate the wallet in software only if needed.

JoeGruffins: > However, this means you can't go from seed to wallet using the standard decred software or my BIP0039 -> raw seed tool.

Where was the tool again? In the decred repo? I'm unsure why it would not work but I trust you are correct.

JoeGruffins: @matheusd I was able to use the tool at https://github.com/matheusd/bip39-to-dcr-seed to restore a decred wallet from seed.

I made a mainnet wallet with decrediton and ledger. Send it a small amount of funds. After restoring using your tool, the seed words, and dcrwallet I was able to access the funds. It works!

JoeGruffins: I should mention that the tool will not work on testnet because the ledger testnet app forces cointype 42' although dcrwallet wants 1'. This is fine for mainnet though.

matheusd: Does that testnet cointype restriction apply even if you change the following line:

https://github.com/decred/decrediton/compare/632d45593fe742186609bffbc1b35c3ed2e59580..77583e96a8774c61c8ee2a3c4891216e93b24410#diff-bebc2af593a6d40794e960263162e817fe54e22c9e72586a4faf7e766ad3d71cR194

?

JoeGruffins: Change "44'/42'/0'" to "44'/1'/0'"? Ledger will give me the error `"Ledger device: Internal error, please report (0x6f15)"`.

app/main.development.js
JoeGruffins: After building, this is not enough. This does not work when packaged.

JoeGruffins:
Just removed the origin here and it is working for the package https://github.com/decred/decrediton/compare/77583e96a8774c61c8ee2a3c4891216e93b24410..f6153862fef4af2b2622e264916cce7abf636059

matheusd (approved):
I've been testing this for bit in both Linux and Windows, and while there are a few testing nits, this is good to go as a first pass implementation.

This is currently disabled in mainnet, so we'll keep iterating on it.

app/helpers/ledger.js
```diff
@@ -0,0 +1,163 @@
...
+    locktime: writeUint32LE(tx.lockTime),
+    nExpiryHeight: writeUint32LE(tx.expiry)
+  };
+}
+
+function createPrefix(tx) {
+  const numOuts = tx.outputs.length;
+  if (numOuts > 2) {
+    throw "more than two outputs is not expected";
+  }
```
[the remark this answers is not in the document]
JoeGruffins: Added a TODO https://github.com/decred/decrediton/compare/f6153862fef4af2b2622e264916cce7abf636059..74284d2e8b32f60dba421b96971a7a2b2c4b0655

app/helpers/ledger.js
```diff
@@ -0,0 +1,163 @@
...
+  let i = 0;
+  buff[i] = numOuts;
+  i += 1;
+  for (const out of tx.outputs) {
+    buff.set(writeUint64LE(out.value), i);
+    i += 8;
+    buff.set(writeUint16LE(out.version), i);
+    i += 2;
+    // TODO: Clean this up for production? Should use smarter logic to get varInts?
+    buff[i] = out.script.length; // varInt for 23/25 bytes
```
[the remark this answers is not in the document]
JoeGruffins: Yeah. Leaving for now...

PR #3882, opened 2023-06-19
[dex] update to v0.6.2 (Decred 1.8 support) and fix response buffer size

chappjc:
This update DEX to [v0.6.2](https://github.com/decred/dcrdex/milestone/32?closed=1), which is required for the defined Decred 1.8 hard forks.

This also increases the "response buffer" used for dex module calls.  Several users have reported an error due to the buffer being too small, so we are increasing it.

Tested with the **native** SPV wallet.


chappjc:
Release notes for dex 0.6.2 are up: https://github.com/decred/dcrdex/releases/tag/v0.6.2

Needed for Decred 1.8 hard forks.

alexlyp (approved):

PR #3884, opened 2023-06-25
multi: Update to electron v23.3.8

JoeGruffins:
This is needed for `webusb` support because it is the earliest release to include these changes https://github.com/electron/electron/pull/36289

`webusb` is needed to use Ledger over usb without having to add anything to the "main_dev" layer.


app/main.development.js
```diff
@@ -772,7 +772,7 @@ app.on("ready", async () => {
     minHeight: 299,
     height: 790,
     webPreferences: {
-      nodeIntegration: false,
+      nodeIntegration: true,
```
JoeGruffins, line 775: Getting errors like:
```
VM4 sandbox_bundle:2 Unable to load preload script: /home/joe/git/decrediton/app/dist/wallet-preload.js
(anonymous)	@	VM4 sandbox_bundle:2
VM4 sandbox_bundle:2 Error: module not found: fs
    at preloadRequire (VM4 sandbox_bundle:2:82255)
    at Object.fs (<anonymous>:182057:18)
    at __webpack_require__ (<anonymous>:182411:42)
    at ./app/wallet/fs.js (<anonymous>:139438:34)
    at __webpack_require__ (<anonymous>:182411:42)
    at ./app/wallet-preload.js (<anonymous>:137577:34)
    at __webpack_require__ (<anonymous>:182411:42)
    at <anonymous>:182438:37
    at <anonymous>:182441:12
    at runPreloadScript (VM4 sandbox_bundle:2:83095)
bootstrap:28 Uncaught TypeError: Cannot read properties of undefined (reading 'blake256')
    at ./app/helpers/walletCryptoModule.js (walletCryptoModule.js:7:38)
    at __webpack_require__ (bootstrap:25:4)
    at fn (hot module replacement:63:1)
    at ./app/helpers/addresses.js (addresses.js:10:1)
    at __webpack_require__ (bootstrap:25:4)
    at fn (hot module replacement:63:1)
    at ./app/helpers/index.js (index.js:5:1)
    at __webpack_require__ (bootstrap:25:4)
    at fn (hot module replacement:63:1)
    at ./app/selectors.js (selectors.js:17:1)
```

And I guess this is not the change we want. Will try to find the correct solution.

JoeGruffins: Ok, it seems the with v20 the `sandbox` value's default was changed from false to true. So, setting false, and I believe this is not a change from the previous behavior. https://www.electronjs.org/docs/latest/tutorial/sandbox

https://github.com/decred/decrediton/compare/e15e45658a20572520d2e9a8b2a0533d3297d4af..76047017969fc70e826335538c497a963dcdfd22

matheusd: The `Preload Scripts` section of the doc you linked to hints as to why you're getting the error: 

> [...] preload scripts attached to sandboxed renderers will still have a polyfilled subset of Node.js APIs available.

And notably, `fs` is **not** one on the list of available modules. So in order to support sandbox mode, we'd have to drop move all references to `fs` (and any other unsupported node/electron modules) from the preload layer to the main_dev layer. It would have to be done carefully though, to avoid re-opening generic holes from which stuff in the renderer layer could use to escape to the main_dev layer.

It would be a worthwhile project to do, but I agree that disabling sandbox with the current impl as we're upgrading the electron version is the way to go now, and working to get `sandbox: true` would be a separate issue/pr.

matheusd (approved):
Tested on both linux and windows.

FYI @alexlyp, I had to update my windows build setup (msys2 version, node version, python version, VC build tools version) in order to get the native dcrwin32ipc module to build, and do the whole clean-yarn-rebuild natives dance to get it to compile, but I do have it syncing now.

@JoeGruffins any specific reason not to update to the _latest_ (v25) electron version? Regardless, this is working as intended.

JoeGruffins:
> @JoeGruffins any specific reason not to update to the latest (v25) electron version? Regardless, this is working as intended.

The reason was that this is the lowest version that works, that is all. Would you like to update to v25?

alexlyp:
> Tested on both linux and windows.
> 
> FYI @alexlyp, I had to update my windows build setup (msys2 version, node version, python version, VC build tools version) in order to get the native dcrwin32ipc module to build, and do the whole clean-yarn-rebuild natives dance to get it to compile, but I do have it syncing now.
> 
> @JoeGruffins any specific reason not to update to the _latest_ (v25) electron version? Regardless, this is working as intended.

Thanks for the headsup @matheusd

PR #3911, opened 2023-09-28
trezor: Use connect v9.

JoeGruffins:
Temporarily copy separately built iframe files.

I am unable to use trezor on master currently. Unsure if it is my set up but I tried on linux and windows and the iframe seems to be giving a 404 page not found error. This diff updates connect to use the newest version and this is working for me. Unfortunately I cannot find a way to build the iframe.js like we were before yet. For now a temporary, prebuilt package is added.


JoeGruffins:
Removed all firmware https://github.com/decred/decrediton/compare/e9d4f2e842ed459c98e36cbb9cc634e4213c6666..d9d973dee478716b3391eeaeb234ba0e8c485957

Issue #3901, opened 2023-08-28
Trezor fixes (bundle bridge + udev rules)


jzbz:
As of this moment Decrediton relies on the user manually installing the [Trezor Bridge](https://data.trezor.io/bridge/latest/index.html) software before they can use their device.

- Decrediton should make use of whatever mechanism Trezor Suite uses to access the device, which likely means bundling all the necessary tools for Trezor to just work without the user needing to hunt down additional tools.
- It should also prompt the user and offer to automatically install the [Trezor udev rules](https://data.trezor.io/udev/51-trezor.rules).
- We may want to take inspiration from how [Trezor solved this](https://github.com/trezor/trezor-suite/pull/4133) for Trezor Suite.

Relevant Matrix discussions:

https://matrix.to/#/!MYLcxlwzxwViTaFPGo:decred.org/$dqHvqW7IXOhrxvJzViTFboXncwvgpd7qEKli_Q6ZE3M?via=decred.org&via=matrix.org

https://matrix.to/#/!xUNvyzkFgiMjhvPbIi:decred.org/$IIiIaoprWFJa_wm1vrMzClLWTJMTbT8Ab7kAdYsuTPc?via=decred.org&via=matrix.org&via=t2bot.io

JoeGruffins:
@jzbz 

Automatic application of the udev rules in problematic because it needs admin permissions to copy which is not great and does not work out of the box on some distros, like mine. Would instructions in case we suspect the udev rules to be missing be enough? More comments on #3950

jzbz:
> @jzbz
> 
> Automatic application of the udev rules is problematic because it needs admin permissions to copy which is not great and does not work out of the box on some distros, like mine. Would instructions in case we suspect the udev rules to be missing be enough? More comments on #3950

Yes I think that's a fair compromise, prompt the user with the instructions on how to do it themselves if they seem missing.

PR #3913, opened 2023-09-29
trezor: Add trezord-go to bin.

JoeGruffins:
If a user opens a trezor wallet, also start the trezor bridge that is
needed for communications.

part of #3901

### DIFF
--- a/app/actions/TrezorActions.js
+++ b/app/actions/TrezorActions.js
@@ -1,4 +1,5 @@
 import { wallet, fs } from "wallet-preload-shim";
+import { trezord } from "wallet-preload-shim";
 import * as selectors from "selectors";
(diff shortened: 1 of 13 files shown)


jzbz:
Bumping this and adding a question; can we add a prompt that asks the user for permission to add the relevant udev rules so everything "just works"?

https://github.com/trezor/trezor-common/blob/master/udev/51-trezor.rules

JoeGruffins:
> can we add a prompt that asks the user for permission to add the relevant udev rules so everything "just works"?

I was going to do this in another pr I think but I'll try to add the comment here. Ill also update this and make sure its still ok now.

JoeGruffins:
Oh theres been no new commits in five months and its still based on master, lol

PR #3927, opened 2023-11-07
electron: Update to 27.0.3.

### LINKED ISSUE #3921: Ledger: Broken on master.
master is currently 391d8a54e73b543eb8570705c0ff56d62aeca89c

Ledger needs WebUSB in order to work which is not supported by the current electron version. Please see the discussions in #3884

JoeGruffins:
closes #3921

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,8 @@ To start decrediton from command-line (assuming it is installed in `/Application
 
 ### Windows
 
+Requires at least Windows 10.
+
 Windows is tricky, due to some things working better on MSYS2, while some things
 only working on cmd.exe, and native module building being very tough to get
 right.
--- a/package.json
+++ b/package.json
@@ -239,7 +239,7 @@
     "core-decorators": "^0.20.0",
     "cross-env": "^7.0.3",
     "css-loader": "5.2.7",
-    "electron": "21.2.3",
+    "electron": "27.0.3",
     "electron-builder": "22.14.13",
     "enzyme": "^3.11.0",
     "eslint": "7.32.0",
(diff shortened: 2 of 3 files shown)


alexlyp:
@matheusd That was what I encountered as well.  I was able to narrow it down to something in https://github.com/electron/electron/releases/tag/v21.3.0 that was causing it to not work on Linux anymore.  Hence why I rolled back to 21.2.3.  Afaict that's the last working version of electron.  And the only change in there that could possibly have a difference is a Chromium update.

matheusd:
While debugging I set the dex window to openDevTools and I get an error there. That may or may not be the issue. I'm retesting on master and then will test on 21.3.0 to see if that's the issue

matheusd:
This is the error I get starting on electron 21.3.0 when opening the dex window (this is the console for the dex window, **not** the main decrediton window):

image

matheusd:
I expect something in chromium broke a lib that the dex interface uses on linux.

matheusd:
I believe this is the issue:

`navigator.languages`
  - on 21.2.3:  `['en-US']`
  - on 21.3.0: `['en-US', 'c']`
  
We'll probably need to check with the DEX UI ppl how to fix this

JoeGruffins:
> On Windows the `rebuild-natives` call failed due to node-abi not supporting electron 27.0.3. I had to downgrade to 26.4.3 to rebuild native modules. DEX worked on that version.

@matheusd Shall I make this upgrade to 26.4.3? That also solves the main ledger issue.

matheusd:
> @matheusd Shall I make this upgrade to 26.4.3? That also solves the main ledger issue.

Yes please. This should fix the windows build, though there's still the issue of the DEX UI failing on linux.

alexlyp:
Closing this one in favor of another that includes the dcrdex fix as well.

PR #3928, opened 2023-11-08
electron: Update to 26.4.3

alexlyp:
This allows for ledger to work as expected and fixes an issue that was causing DEX window not to show anything past 21.3.0  electron.


matheusd (approved):
This one worked.

PR #3950, opened 2024-05-21
trezor: Copy udev rules.

### LINKED ISSUE #3901: Trezor fixes (bundle bridge + udev rules)

JoeGruffins:
closes #3901

### DIFF
--- a/app/main_dev/ipc.js
+++ b/app/main_dev/ipc.js
@@ -18,6 +18,7 @@ import {
   launchDCRLnd,
   GetDcrlndPID,
   GetDcrlndCreds,
+  copyTrezorUdevRules,
   launchTrezord,
   GetTrezordPID,
   closeTrezord,
@@ -261,8 +262,14 @@ export const startTrezord = async () => {
   }
 
   try {
-    const started = await launchTrezord();
-    return started;
+    await copyTrezorUdevRules();
+  } catch (e) {
+    logger.log("error", `error copying trezor udev rules: ${e}`);
+    return e;
+  }
+
+  try {
+    launchTrezord();
   } catch (e) {
     logger.log("error", `error launching trezord-go: ${e}`);
     return e;
--- a/app/main_dev/launch.js
+++ b/app/main_dev/launch.js
@@ -6,7 +6,9 @@ import {
   getAppDataDirectory,
   getDcrdPath,
   getCertsPath,
-  getSitePath
+  userDataPath,
+  getSitePath,
+  getBuildPath
 } from "./paths";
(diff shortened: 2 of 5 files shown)


JoeGruffins:
It seems like `electron-sudo` this needs the package https://www.nongnu.org/gksu/ installed, which isn't even offered on my distro. It should be on ubuntu. Trezor is using `pkexec` which should be only command line only. I need to try out their package again with no rules to see what it looks like.

matheusd:
I'm pretty hesitant on this. 

First, it adds an additional dependency on the electron layer, which we should be trying to avoid at all costs (due to possible supply chain attacks). 

Second, adding sudo capabilities natively to the wallet seems unwise (imagine the cross section of a supply chain attack having access to the sudo abilities of the wallet, masking what is attempting to do with a standard wallet dialog).

I would be a bit less worried if this was isolated completely in a separate wallet process/layer (not sure if this is possible) and was only launched by an explicit user action to install the trezor udev rules (to avoid having this code possibly accessible to other parts of the app).

JoeGruffins:
Yeah, the trezor suite download does not work for me. Stuck here forever:

image

@jzbz @matheusd 

Would just a popup with an explanation and a link to maybe https://trezor.io/learn/a/udev-rules be fine? We can also not use the link and include the rules and how to copy them.

matheusd:
Showing a popup with instructions and a link for additional explanation seems fine to me (this is linux only after all).

JoeGruffins:
Going to show instructions instead of doing it this way in a new pr.

PR #3959, opened 2024-08-23
libdexc: upgrade to v1 dcrdex and initialize market maker

buck54321:
@alexlyp I think you were already looking at the upgrade part, but the market maker initialization wasn't obvious. I've also flipped [the new switch](https://github.com/decred/dcrdex/pull/2927) to use dcrdex branding in Decrediton.

This currently points to the tip of the release-v1.0 branch. This will build with Go 1.23 now. There will be a couple more commits on that branch before release. I can update later, or you can update if you were working separately on the upgrade.


buck54321:
Updated to tip of release-v1.0 and squashed. I don't expect any more substantive changes.

modules/dex/libdexc/adapter.go
```diff
@@ -151,13 +151,32 @@ func (c *CoreAdapter) startCore(raw json.RawMessage) error {
...
 	c.wg.Add(1)
 	go func() {
 		defer c.wg.Done()
 		ccore.Run(c.ctx)
 	}()
 	<-ccore.Ready()
 
+	mmCM := dex.NewConnectionMaster(c.mm)
+	if err := mmCM.ConnectOnce(c.ctx); err != nil {
+		return fmt.Errorf("Error connecting market maker")
```
JoeGruffins, line 172: Capitalized error return.

JoeGruffins (approved):

alexlyp (approved):

PR #4031, opened 2026-04-14
Fully sanitize wallet name of invalid chars.

jholdstock:
When a new wallet name has been provided by the user, every instance of invalid characters should be removed from the name, not just the first.


bgptr (approved):
lgtm

alexlyp (approved):

